PR #121, opened 2015-01-26
FairMQSocket: add versions of Send/Receive methods with int flags

rbx:
FairMQSocket: add versions of Send/Receive methods with int flags instead of string, which is more flexible and faster.
Int flags are mapped to their ZeroMQ/nanomsg versions behind the transport interface.
Methods with string flags are kept for backwards compatibility.


MohammadAlTurany:
merged

PR #132, opened 2015-02-12
Dbmq lvl4

denisbertini:
Decoupling of ParBase and FairDB Libraries. Now FairDB includes the FairRuntimeDB bindings.


denisbertini:
Be aware that these changes the inheritance layout of the parameter container classes ( FairParSet ).
which breaks the compatibility with the former parameter container written in ROOT files.

MohammadAlTurany:
Merged with dev, I test it with CBM reconstruction: Producing the parameter files with the old inheritance and reading them with the new one just produce a warning. but all work without problems.

PR #134, opened 2015-03-03
Add option -DWITH_DBASE to cmake

MohammadAlTurany:
Add the option -DWITH_DBASE to the cmake options, by default this option is on.


MohammadAlTurany:
Merged

PR #140, opened 2015-04-10
Beam gap

TobiasStockmanns:
Functionality to simulate time structure of beam with FairRoot added (e.g. 1600 ns beam and 400 ns gap).


MohammadAlTurany:
merged

PR #172, opened 2015-06-25
Test beam additions

TobiasStockmanns:
This adds a new run method RunOnTBData in FairRunAna which stops once it gets a FinishRun() signal from the FairRootManager


MohammadAlTurany:
Merged

PR #189, opened 2015-08-29
Use the "required" flag in FindFairRoot.cmake

Barthelemy:
If REQUIRED is not set we must not throw a fatal error.


MohammadAlTurany:
Hallo Barth,

Sorry that it took so long! Anyway could you please make the pull request to the dev branch and not to the master.
Thanks,

Mohammad

PR #190, opened 2015-09-01
Add FindDDS cmake module

rbx:


rbx:
Will remove `add_definitions`. This shouldn't be global.

PR #191, opened 2015-09-02
Changes regarding installation in mode 'reco_only'

matthiasrichter:
I did an attempt to install FairSoft packages without the simulation engines ('reco only'). I had to do two modifications in the FairRoot cmake setup to compile FairRoot:
1) Pythia6 is not installed but it was a required package. I made it optional, and the compilation works without problems
2) CLHEP is not installed. Module trackbase had the CLHEP include path in the setup, but this seems to be obsolete.

In addition I introduced the possibility to specify an external Geant3 package using flag -DGEANT3_PATH=<path>

Please consider merging those changes into FairRoot.

Cheers
Matthias


fuhlig1:
Implemented with commit 2660342 and the three previous commits.

PR #209, opened 2015-10-05
Reconstruction only option

MohammadAlTurany:
When Geant 3 is not found (optional, not required any more) switch off the modules depending on it!


MohammadAlTurany:
Merged

PR #216, opened 2015-10-13
Getting MCTracks sorted by weight without C++11

TobiasStockmanns:
Code is rewritten to use a function instead of a lambda expression.
Runs on my Mac and Linux machine but both have C++11 support.


MohammadAlTurany:
Merged

PR #220, opened 2015-10-15
Add methods for setting timeout on blocking Send/Receive

rbx:
- FairMQ: Add methods for setting timeout on blocking Send/Receive calls (and a test for it).
- FairMQDevice: Extend Doxygen documentation


rbx:
Rebased.

PR #225, opened 2015-10-30
Remove compress

TobiasStockmanns:
Removed the not used compress functionality of the FairRootManager.
Tested with PandaRoot. Please check before merge with FairRoot test macros.


MohammadAlTurany:
Merged

PR #233, opened 2015-11-03
Several FairMQ fixes

rbx:
- FairMQChannel: return if an unknown exception has been caught.
- FairMQChannel: add proper return values in set timeout.
- Examples: check return values of the Receive calls.
- nanomsg: fix return values of send/receive calls.


MohammadAlTurany:
Merged

PR #252, opened 2015-11-19
BoostSerializer and GenericDevices test

NicolasWinckler:
- add a serializer function that take a pointer to datatype in BoostSerializer
- change io directory for the tests of the GenericDevices example
- generate three different files  for the tests of the GenericDevices example


fuhlig1:
Merged with  16fcfe9b91ca08f5bef4d22b001efc520b87f590

PR #259, opened 2015-12-04
Dev

Barthelemy:
Enable more examples

Some tests do not depend on Geant yet they were not compiled if geant was absent. This is not the case any more.

I don't understand why there are two other commits sticking around my fork / branch, they already exist in FairRoot/dev.


MohammadAlTurany:
Hi Barth,

This came some how because these two comments where in master and not in dev (through the RC branch). The pull request is now merged with the dev branch.
Cheers,

Mohammad

PR #260, opened 2015-12-08
Create lib directory in build structure

Barthelemy:
In some cases (e.g. clion ide on linux) the lib directory in the build structure is not created before creating the rootmap.
To avoid problems, make sure to create a directory "lib" beforehand.


MohammadAlTurany:
Merged

PR #262, opened 2015-12-11
Fixes and modifications found during the Nov 15 CBM TOF beamtime at CERN 

PALoizeau:
All changes relate to LMD/MBS events sourcing, except the fix restoring the console text color after a FairLog FATAL.
Changes are relative to tag v-15.11.
Compilation tested with gcc 4.8.4 and FairSoft nov15p1 root5.


fuhlig1:
The changes have been implemented with two separate commits. One for the source related stuff ( 6b11ff1) and one for the FairLogger (17f3ebd).

PR #268, opened 2015-12-16
Fixing an efficiency gap in FairWriteoutBuffer

TobiasStockmanns:
This update fixes an efficiency gap in the FairWriteoutBuffer which causes data loss especially for very short dead time detectors.
This fix was programmed by Dominik Steinschaden (Dominik.Steinschaden@oeaw.ac.at).


fuhlig1:
The code was merged with commit d0eafd6.

PR #285, opened 2016-02-05
Boost fixes

ktf:
Sometimes boost libraries have a -mt (as in multithreading) appended to
their name. This is particularly the case on my mac, using homebrew.
With the attached I managed to get it compile fine.


ktf:
As discussed, I've changed the logic of my boost recipe so that it picks up boost from the system if and only if `libboost_thread` does not have a `-mt` suffix. We can then think how to deal with the problem later on.

MohammadAlTurany:
Hi Giulio,

Removing the explicit names of the libraries from the CMakefile.txt should solve this problem now! 
can you please confirm if it do?
best,
Mohammad

ktf:
Sure, looking into it.

PR #292, opened 2016-02-26
Correctly load and use the parameter file.

fuhlig1:


fuhlig1:
There is an updated pull request.

PR #303, opened 2016-03-10
Allocate dynamic size flatbuffers structure on the heap

rbx:
+ Allocate dynamic size flatbuffers structure on the heap.
+ Merger, Splitter: remove debug output when handling empty messages.


MohammadAlTurany:
Merged

PR #313, opened 2016-03-14
examples/advanced: cmake support for go-examples

sbinet:
This CL adds basic support for compiling the Go-based examples.
Go is quite opinionated wrt a project layout.
It needs an environment variable $GOPATH (not unlike $PYTHONPATH)
pointing at a directory, under which a 'src' directory will hold all
the sources:
```
 $> cd $GOPATH
 $> ls
 src bin pkg
 $> tree -L 4 src
 src
 └── github.com
     ├── go-mangos
     │   └── mangos
     │       ├── AUTHORS
     │       ├── compat
     │       ├── conn.go
     │       ├── core.go
     │       ├── ...
     │       ├── util.go
     │       └── waiter.go
     └── sbinet-alice
         └── go-ex-sink-mq
```
This allows to have:
```
 $> go get github.com/sbinet-alice/go-ex-sink-mq
```
- fetch the sources from github.com/sbinet-alice/go-ex-sink-mq
- retrieve the dependencies of that program
  (github.com/go-mangos/mangos/...)
- compile all libraries under $GOPATH/pkg
- compile and install programs under $GOPATH/bin

To integrate with CMake:
- put the sources of go-processor and go-sink under
  `advanced/GoTutorial/src`
- make GOPATH point at advanced/GoTutorial
- use $GOBIN to make `go get` install the program(s) under that
  directory instead of $GOPATH/bin (which would be against CMake
  out-of-tree-build policy)


sbinet:
@fuhlig1 PTAL.

sbinet:
ping?

MohammadAlTurany:
Merged with dev.

sbinet:
@MohammadAlTurany well, thanks, but the content of the `examples/advanced/GoTutorial/CMakeLists.txt` file was more of a "fill-in-the-blanks" than anything else :)

(at the time, I wasn't completely sure what was the `cmake` syntax to test for the presence of binaries (_e.g._ `go`), the syntax to run external programs, etc...)

MohammadAlTurany:
@sbinet, I tested only that it works and does not influence the rest of the stuff.  For the cmake we have to add a FindGo.cmake (Find package macro) to the cmake Modules. I will add one and modify the CMakeList.txt.
best,
Mohammad

PR #319, opened 2016-03-18
ROOT test script did not take into account externally defined LD_LIBR…

sawenzel:
…ARY_PATH

detail: test scripts generated from ROOT macros (such as examples/simulation/Tutorial1/macros)
        set an LD_LIBRARY_PATH which is not complete. In particular it does not pick up any external
        LD_LIBRARY_PATH to my compiler (which in this case resides somewhere on the afs) but just
        overwrites it.

        This patch provides a workaround by augmenting the tests LD_LIBRARY_PATH with the externally
        user-defined LD_LIBRARY_PATH. An alternative solution might be to fix the generated
        LD_LIBRARY_PATH which should take into account the compiler/libc paths
        (as it does in PATH by the way).

### DIFF
--- a/cmake/scripts/root_macro.sh.in
+++ b/cmake/scripts/root_macro.sh.in
@@ -8,8 +8,8 @@
 # to be passed to root.
 
 # Setup the needed environment
-export LD_LIBRARY_PATH=@MY_LD_LIBRARY_PATH@
-export DYLD_LIBRARY_PATH=@MY_LD_LIBRARY_PATH@
+export LD_LIBRARY_PATH=@MY_LD_LIBRARY_PATH@:$LD_LIBRARY_PATH
+export DYLD_LIBRARY_PATH=@MY_LD_LIBRARY_PATH@:$DYLD_LIBRARY_PATH
 export PATH=@SIMPATH@/bin:$PATH:@Geant4_DIR@
 export ROOTSYS=@ROOTSYS@
 export ROOTEXE=@ROOT_INSTALL_DIR@/bin/root.exe


fuhlig1:
Hi,

the proposed change is not needed. If you need the original LD_LIBRRAY_PATH added to the configuration you simply have to add a flag to your cmake command. If you do

cmake -DUSE_PATH_INFO=TRUE <your original flags and source_dir>

the content of the environment variables PATH and (DY)LD_LIBRARY_PATH are appended to the cmake variables.

Ciao

Florian

sawenzel:
Thanks. I can do this of course. 

What speaks against doing this by default? In my case, I actually did an automatically managed build and the path was not picked up correctly.

Cheers

Sandro

fuhlig1:
Hi Sandro,

The problem with this setting as default is that is not needed very often but in special cases. Some years ago we had this as default and had a lot of problems when
people messed up their configurations. Sometimes it was very hard to find why things went wrong. So we decide at that time to skip the PATH and LD_LIBRARY_PATH
environment variables and allow people to overwrite the default behavior when needed.

Do your use our Dart.sh script for your automatic build?

Ciao

Florian


Am 18.03.2016 um 14:22 schrieb sawenzel <notifications@github.com>:

sawenzel:
Hi Florian,

I used the ./configure ./alfaconfig scripts as well as a manual cmake build.

I am fine with using -DUSE_PATH_INFO. Maybe is should be documented/mentioned somewhere (which I did not check!! ).
Additionally, it might be a good idea to detect situations in which the compiler/runtime libs do not sit in standard locations and then either warn the user or add those relevant lib paths automatically.

PR #343, opened 2016-04-11
Fix fairrunonline

fuhlig1:


MohammadAlTurany:
Hi Florian,

FairSource.h a method is defined twice :

Line  39:  virtual Bool_t   ActivateObject(TObject**, const char*)  { return kFALSE; } 
Line: 43  virtual Bool_t   ActivateObject(TObject** obj, const char* ObjType)  { return kFALSE; }

Please correct it!

best,

Mohammad

fuhlig1:
Thanks. I fixed it. Let�s see if it compiles now.

Ciao

Florian

Am 11.04.2016 um 17:20 schrieb Mohammad Al-Turany <notifications@github.com>:

PR #355, opened 2016-04-25
Pixel9

karabowi:


fuhlig1:
Could you please handle in the Exec function of examples/MQ/9-PixelDetector/src/PixelDigitize.cxx also the case when there is no proper curNode returned.

karabowi:
Sorry Florian,

I missed that one. I have created another pull request.
yours
radek

On 25 Apr 2016, at 17:55 pm, fuhlig1 <notifications@github.com<mailto:notifications@github.com>> wrote:


Could you please handle in the Exec function of examples/MQ/9-PixelDetector/src/PixelDigitize.cxx also the case when there is no proper curNode returned.

MohammadAlTurany:
Merged

PR #359, opened 2016-04-28
Logger fix

rbx:
Previous fix did not work if the logger was not used before configuration, because it is initialized on first use. This makes sure this doesn't happen.

### DIFF
--- a/fairmq/options/FairMQProgOptions.cxx
+++ b/fairmq/options/FairMQProgOptions.cxx
@@ -35,6 +35,7 @@ FairMQProgOptions::~FairMQProgOptions()
 
 int FairMQProgOptions::ParseAll(const int argc, char** argv, bool allowUnregistered)
 {
+    LOG(NOLOG) << "";
     // init description
     InitOptionDescription();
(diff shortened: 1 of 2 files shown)


fairmq/options/FairMQProgOptions.cxx
```diff
@@ -35,6 +35,7 @@ FairMQProgOptions::~FairMQProgOptions()
 
 int FairMQProgOptions::ParseAll(const int argc, char** argv, bool allowUnregistered)
 {
+    LOG(NOLOG) << "";
```
AnarManafov: Alexey, obviously what you are trying to do is not a solution and not even a workaround.
Since you  based your logger on DDS's one, why don't you just implement an init method, which needs to be called in the main before using the logger?
Logger is a singleton and even multiple calls for init won't hurt.

rbx: I agree that this is not a good solution, but I want to wait for Nicolas before making any big changes to the logger. For now this will have to suffice to solve an occurring crash.

PR #361, opened 2016-05-04
Changes in FairRunOnline needed by CBM.

kresan:
Add possibility to skip writing an event to tree.
Move SetParContainers for unpackers to base FairSource class. Avoid dynamic cast in FairRunOnline.
Set the run ID using source.
Fix linking error with Xcode 7.3.
Add ReInit to the Unpacker class. Call ReInitUnpackers from FairRunOnline.

### DIFF
--- a/base/source/FairFileSource.h
+++ b/base/source/FairFileSource.h
@@ -45,6 +45,12 @@ public:
 
     virtual Source_Type GetSourceType() { return kFILE; }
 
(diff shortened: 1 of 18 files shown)


kresan:
Now fSource->Close() is called from FairRunOnline::Finish().
Remove -Wall and -Wextra for NIGHTLY build on OS X.

PR #363, opened 2016-05-11
Version check for ZMQ introduced, required version 4.1.3

matthiasrichter:
Version of ZMQ is extracted from zmq.h, required version set to 4.1.3
Searching now also in system directories (NO_DEFAULT_PATHS removed)


matthiasrichter:
one more correction needed

PR #365, opened 2016-05-11
Version check for ZMQ introduced, required version 4.1.3

matthiasrichter:
Version of ZMQ is extracted from zmq.h, required version set to 4.1.3
Searching now also in system directories (NO_DEFAULT_PATHS removed)

Bugfix to correctly propagate the QUIET and REQUIRED flags, variables
changed from ZMQ_FIND_* to ZeroMQ_FIND_*
Also the logic is fixed to let REQUIRED take precedence over QUIET.


MohammadAlTurany:
Merged

PR #367, opened 2016-05-12
Several fixes and improvements

rbx:
- runStateMachine: Remove redundant state change call
- FairMQDevice: revert the termination handler (needs to be done differently).
- FairMQSocket: Thread safe socket rate stats (to avoid very rare incorrect rate logging information).
- Cleanup examples (remove unused includes and unify code approach).
- MQ Parameter client example: cast returned parameter type to the expected type.
- Let config.ParseAll exit if unsuccessful:
```C++
// OLD:
if (config.ParseAll(argc, argv))
{
    return 0;
}
// NEW:
config.ParseAll(argc, argv);
```


MohammadAlTurany:
Merged

PR #374, opened 2016-05-27
nanomsg: reset maximum receive msg size

rbx:


rbx:
NN_RCVMAXSIZE is only defined since nanomsg 0.7. Need to add a guard/version check.

PR #384, opened 2016-06-01
cmake/modules: modular build support for nanomsg

sbinet:
cmake/modules: modular build support for nanomsg
    
use `NANOMSG_DIR` and/or `AlFa_DIR` to infer nanomsg location to support
modular builds.


MohammadAlTurany:
Merged

PR #390, opened 2016-06-08
Use Boost_ variables instead of hardcoded names

ktf:


ktf:
@fuhlig1 @MohammadAlTurany this is required to build using aliBuild (and in general on Mac).

PR #391, opened 2016-06-08
Corrected PROTOBUF conditional built of some tutorials

matthiasrichter:
Compilation fails in some tutorials if PROTOBUF is not available


rbx:
Hmm, I think tutorials 1,2 and 4 do not depend on protobuf? Only the third one does.

matthiasrichter:
correct, I did not properly interpret the error, will fix it

PR #392, opened 2016-06-08
Corrected PROTOBUF conditional built of Tutorial3

matthiasrichter:
now its correctly only affecting Tutorial3

### DIFF
--- a/examples/advanced/Tutorial3/CMakeLists.txt
+++ b/examples/advanced/Tutorial3/CMakeLists.txt
@@ -42,6 +42,7 @@ Set(SYSTEM_INCLUDE_DIRECTORIES
 )
 
 If(PROTOBUF_FOUND)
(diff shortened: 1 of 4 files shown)


fuhlig1:
Merged with branch RC_v-16.06 at commit 0f2f040.

matthiasrichter:
Hi, is there a reason why it should not go into the dev branch as well? 
Ciao Matthias

fuhlig1:
It will when the release is done. It would be good if you could use the RC_v-16.06 branch for the time being and test if everything is working as expected.

Ciao

Florian



Am 21.06.2016 um 09:27 schrieb Matthias Richter <notifications@github.com<mailto:notifications@github.com>>:


Hi, is there a reason why it should not go into the dev branch as well?
Ciao Matthias

matthiasrichter:
ok, will do

PR #394, opened 2016-06-29
Const Momentum() and matching Position()

janmayer:


fuhlig1:
Hi,

thanks for the fix which was applied with 57be7d2  to the RC_v-16.06 branch. The fix will be in the next release and then also in master and dev branches.

PR #399, opened 2016-07-18
Findnanomsg.cmake: Only search for the shared nanomsg library

rbx:
New nanomsg version (1.0.0) creates _either_ shared _or_ static library, so our existing check would fail since it is looking for both. Remove the search for the static library - we are only using shared nanomsg library.


MohammadAlTurany:
Merged!

PR #404, opened 2016-08-04
Test nanomsg transport only when nanomsg library is found.

rbx:


fuhlig1:
Merged with commit 16bdb8b

PR #405, opened 2016-08-04
Avoid hardcoded boost library names

ktf:
This should be the last of the longstanding issue due to boost having
different "-mt" extension policy on different systems.


ktf:
@pzhristov

ktf:
@MohammadAlTurany , @fuhlig1 can this be merged? This fixes a common issue on mac for people building with homebrew installed software.

PR #406, opened 2016-08-04
merge MQ parameters in the variable_map container of FairMQProgramOpt…

NicolasWinckler:
…ions class
* add UpdateValue function which update the variable_map. Use: config.UpdateValue<type>(key,val). The updated values are propagated to the channel map if the latter was initialized by UserParser() or by UpdateChannelMap() call.


fuhlig1:
merged wit commit 520d632

PR #409, opened 2016-08-11
Allow specifying where to find root-config explicitly

ktf:
This way we do not need to have ROOTSYS set when compiling FairRoot.


MohammadAlTurany:
Merged

PR #422, opened 2016-09-06
Refacture the tests for the FairLogger class.

fuhlig1:
Put all tests in different files/executables.
This was necessary because the FairLogger is a singleton
which was created only once and used for several tests.
Some of the tests failed after changes because
an earlier test changed the state of the instance.

Remove the old FairLogger tests.
Remove obsolete class to capture screen output.

Don't do coverage analysis for tests. We only want to know the
coverage for the FairRoot code.


MohammadAlTurany:
Merged

PR #426, opened 2016-09-16
MQ Config + logger

NicolasWinckler:
- Add a signal/slot mechanism in FairMQProgOptions with example. Alternative API with string is available but commented
- Simplify logger API, fix double printing issue on error messages


MohammadAlTurany:
Merged

PR #474, opened 2017-01-27
Pixel9a payload array

karabowi:
Introduction of example 9a.
The example shows how to migrate from analysing ROOT data in favour of binary data.
There is possibility to read and write binary file.
Transported and analysed data is only binary.
It is also possible to easily steer the message size by aggregating multiple events
in one multi-part message.


fuhlig1:
Merged with commit 3a7dbd04b51fa9173703d0520f8eae59ab28c345 .

PR #489, opened 2017-02-22
FindZeroMQ.cmake module: refactor and aliBuild/alienv integration 

dennisklein:
In addition to aliBuild/alienv integration I fixed the following smaller issues:

* added more docs
* fixed mismatch between module name `ZeroMQ` and prefix of generated variables `ZMQ` -> `ZeroMQ` is now the preferred prefix
* the aliBuild zeromq recipe does not install `zmq.hpp` header, so the module searches now for the `zmq.h` header (which is anyways the only one we use)
* some cosmetic changes


rbx:
At least AliceO2 and PANDA also use the ${ZMQ_INCLUDE_DIR} variable, and use our FindZeroMQ.cmake to find ZeroMQ.

dennisklein:
Ok, I will add the old variables and mark them deprecated. So nothing will break.

dennisklein:
done. squashed all commits again and pushed.

dennisklein:
If this gets merged, we may want to update minimum CMake version to
```cmake
CMAKE_MINIMUM_REQUIRED(VERSION 3.0.2 FATAL_ERROR)
```
from currently 2.8.2, because according to the documentation the deprecation message mode is only available from 3.0.2 on.

MohammadAlTurany:
@dennisklein Please update the minimum CMake version in this pull request.

dennisklein:
@MohammadAlTurany Done.

PR #499, opened 2017-03-07
Added SetEkinRange to FairBoxGenerator

janmayer:
For my/our (R3BRoot) use cases it is common to give the primary particle kinetic energy instead of the momentum.
Here is a possible solution using the particle mass to convert the given energy range to a momentum range during initialization. This way, ReadEvent can remain unchanged.

### DIFF
--- a/generators/FairBoxGenerator.cxx
+++ b/generators/FairBoxGenerator.cxx
@@ -1,8 +1,8 @@
 /********************************************************************************
  *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
  *                                                                              *
- *              This software is distributed under the terms of the             * 
- *         GNU Lesser General Public Licence version 3 (LGPL) version 3,        *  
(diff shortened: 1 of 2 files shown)


janmayer:
I have no idea why `run_rutherford_TGeant3` is failing, it does not fail on my machine (linux/gcc/root6). Is there a way to get the verbose output from travis (e.g. `CTEST_OUTPUT_ON_FAILURE=1`)?

fuhlig1:
Hi Jan,

the results of the Travis Tests are committed to our CDash Server at https://cdash.gsi.de/CDash/index.php?project=FairRoot.
here you can check the output of the failing tests.

In your specific case the reason is completely uncorrelated with your pull request. I have restarted the failing Travis Tests.

Ciao

Florian


Am 08.03.2017 um 09:08 schrieb Jan Mayer <notifications@github.com>:

fuhlig1:
Hi Jan,

the restarted Travis Test works now without any problems.

Ciao

Florian

PR #505, opened 2017-03-28
Migration of simulation code to multi-threading:

ihrivnac:
- Added new RootManager classes which allow to use ROOT IO with
  Geant4 multi-threading, according to the implementation in geant4_vmc/mtroot:
  FairRootGenericManager, FairRootManagerSim, FairRootManagerSimMT
  which use the FairRootManager methods implementation
- Updated the following classes for multi-threading:
  FairGenericStack, FairMCApplication

- Not yet migrated data:
  FairMCApplication::fRunInfo
  FairMCApplication::fFairTaskList, fRadGridMan - are these used in O2?
- Left problems - see "TO DO"

### DIFF
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -95,6 +95,9 @@ sim/FairMCApplication.cxx
 sim/FairModule.cxx
 sim/FairParticle.cxx
 sim/FairPrimaryGenerator.cxx
+sim/FairGenericRootManager.cxx
+sim/FairRootManagerSim.cxx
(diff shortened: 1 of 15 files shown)


MohammadAlTurany:
@ihrivnac  fFairTaskList is used when we do digitisation directly with the simulation.

ihrivnac:
The data, mentioned as not yet migrated (fFairTaskList), should not be affected by the changes in this pull request in sequential mode. They can be added in MT mode in the next iteration.

PR #506, opened 2017-03-28
[FairMQ/cmake] Fix libFairMQ/libfairmq_logger overlinking and minor cleanup/refactor

dennisklein:
https://fairroot-redmine.gsi.de/issues/144

Main decissions:
* Is it ok to drop root dictionary generation for libFairMQ?
* Is it ok to include 3rd party dependencies as done for PStreams? It is a small single file header-only library, no compilation needed. If `3rdparty` directory is not desired, I will file the appropriate patches against FairSoft/Alfadist.


dennisklein:
The failed travis builds on Linux ([1](https://cdash.gsi.de/viewConfigure.php?buildid=149289), [2](https://cdash.gsi.de/viewConfigure.php?buildid=149290)) are due to missing imported boost targets in old cmake versions. The needed cmake version is `3.5.0` (see [patch](https://github.com/Kitware/CMake/commit/3f9b081f6ee85e0691c36496d989edbe8382589d)). The current `master` branch of FairSoft already installs `3.7.2`.

Does the `dev` branch need to work with the FairSoft `may16p1` release still?

MohammadAlTurany:
It does not work with CMake 3.6.1, I had to install CMake 3.7.2 on my MAC to get it running, so we have to require CMake 3.7.2 in this PR before merging it.

> Does the dev branch need to work with the FairSoft may16p1 release still?

No, I would say the master should but the dev not

dennisklein:
> It does not work with CMake 3.6.1, I had to install CMake 3.7.2 on my MAC to get it running, so we have to require CMake 3.7.2 in this PR before merging it.

That's interesting: By using the imported boost targets, there are now two reasons for the increased cmake version. The FindBoost.cmake module inside cmake contains a map of boost dependencies in order to construct the imported targets. For recent Boost releases one then needs to keep up with recent cmake releases. If this becomes an issue, we could also copy the FindBoost.cmake module into our repo. This will relax the version constraints on cmake.

I have updated my pull request (bumped required cmake version and various other smaller things). Please test it again on mac. Thank you!

dennisklein:
Just rebased on current `dev` and fixed few issues with initializer lists on GCC `<= 4.9` and clang `<= 3.1`.

PR #510, opened 2017-04-04
ZeroMQ and Go fixes

dberzano:
* Changed ZeroMQ lookup strategy
  * First look for ZeroMQ in custom paths (they have precedence)
  * Then look for it in system's paths
* Fix Go builds (they fail if accidentally you have Go < 1.7 installed on your system)
  * Optionally disable Go with `DISABLE_GO` (not everybody needs it)
  * Fix Go build macro by checking the exitcode of `go get`
  * Add version check for Go (1.7+ required for current tutorial examples)

All changes are backwards-compatible.


dberzano:
@MohammadAlTurany can you have a look/assign it if it looks OK to you (and passes the tests?)

cmake/modules/FindZeroMQ.cmake
```diff
@@ -62,6 +62,9 @@ find_path(ZeroMQ_INCLUDE_DIR NAMES "zmq.h" "zmq_utils.h"
     DOC "ZeroMQ include directories"
     NO_DEFAULT_PATH
 )
+if(NOT ZeroMQ_INCLUDE_DIR)
```
dennisklein: You can achieve the same behaviour in a slightly more idiomatic way by dropping the `NO_DEFAULT_PATH` option of the previous [`find_path`](https://cmake.org/cmake/help/v3.0/command/find_path.html) and [`find_library`](https://cmake.org/cmake/help/v3.0/command/find_library.html) calls. See . If you could change that and add a comment at the top of the file at the paragraph near line 40, that the module will now also search the system paths - that would be awesome :)

dberzano: Hi, I am going to do the comment. You propose to:

* make a _single_ call to `find_*`
* by simply dropping the `NO_DEFAULT_PATH`

The final result would, however, be [as per the quite convoluted documentation](https://cmake.org/cmake/help/v3.4/command/find_library.html) (but I have tested it and can confirm), that **the system one will always have priority** .

The reason behind the double `find_*` calls (one with `NO_DEFAULT_PATH` and `PATHS`, the second only if nothing's found and only on system's paths) is that we want **our ZeroMQ version to always have priority, and fallback on the system one if only if a custom one is not found**.

Personally, I think that this should be CMake's behavior and shouldn't be worked around like this...

A new commit with the comment is coming :-)

MohammadAlTurany: Hi @dberzano 

> our ZeroMQ version to always have priority, and fallback on the system one if and only if a custom one is not found.

Yes, this is what we want.

dennisklein: > The final result would, however, be as per the quite convoluted documentation (but I have tested it and can confirm), that the system one will always have priority .

This [sentence](https://github.com/Kitware/CMake/blob/666bb0e3fa0d212c03f785bd9f6bc05d29925e1f/Help/command/FIND_XXX.txt#L93) is quite misleading to not read point 6. Happened to me at least. So, in the end the cmake docs are accurate :-) @dberzano Does dberzano/FairRoot#1 work for you?

dberzano:
@dennisklein using `HINTS` seems to do the trick indeed, and today I have learned something :-)

PR #514, opened 2017-04-05
Add ZeroMQ include path

ktf:


ktf:
The error seems unrelated.

PR #516, opened 2017-04-05
Fixes to get AliceO2 in somewhat working state again

ktf:


MohammadAlTurany:
@dennisklein It seems that this is needed for the Alibuild, it works with FairSoft, so I will merge it now and we can take it out later.

dennisklein:
PR #520 should have fixed it. I will investigate tomorrow and setup a proper AliceO2 environment on my system. This looks like the wrong place to fix. If really needed here (which I doubt), it should be `PRIVATE` at least.

PR #528, opened 2017-04-10
Fix travis

fuhlig1:


fuhlig1:
Pull request was against master. Close the request.

PR #533, opened 2017-04-11
Remove GenerateHtml()

kresan:
Remove functionality to generate HTML file for browsing a separate ROOT file with histogram objects - replaced by THttpServer support.


fuhlig1:
Committed with db1d431feb.

PR #540, opened 2017-04-25
Remove FairLink member + remove virtual functions from FairTimeStamp

sawenzel:


sawenzel:
@TobiasStockmanns : Can you please take a look?

TobiasStockmanns:
Runs now with PandaRoot without problems. So from my side it can be merged.

sawenzel:
travis exceeded the time limit again (but seems to build correctly in the 3 cases that were completed).

PR #543, opened 2017-04-26
Remove data members / virtual functions from FairMultiLinkedData_Inte…

sawenzel:
…rface

 * remove fVerbose data member which was not used
 * remove fInsertHistory member which is duplicated behind the link pointer
 * make accessor to link pointer not virtual

This PR achieves an 8 byte data reduction for this critical base class.

### DIFF
--- a/base/event/FairMultiLinkedData_Interface.cxx
+++ b/base/event/FairMultiLinkedData_Interface.cxx
@@ -6,10 +6,7 @@
  */
 
 #include "FairMultiLinkedData_Interface.h"
-
 #include "FairRootManager.h"            // for FairRootManager
-
-#include "TClonesArray.h"               // for TClonesArray
 
 #include <algorithm>                    // for find
 #include <iterator>                     // for distance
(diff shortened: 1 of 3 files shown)


base/event/FairMultiLinkedData_Interface.cxx
```diff
@@ -184,12 +180,11 @@ void FairMultiLinkedData_Interface::AddInterfaceData(FairMultiLinkedData_Interfa
 	if (data->GetPointerToLinks() != 0) {
 		AddLinks(*data->GetPointerToLinks());
 	}
-	SetInsertHistory(fInsertHistory);
+	SetInsertHistory(GetPointerToLinks()->GetInsertHistory());
```
sawenzel: @TobiasStockmanns : can you please verify the logic of this line?

TobiasStockmanns: The logic is correct. This change can be done.

PR #547, opened 2017-04-27
Reduce size of FairLink to 15 byte

TobiasStockmanns:
Reducing the size of FairLink down to 15 byte.
This limits the number of file links down to 254 and for types to 65534.
At the moment I do not check if values larger than that are given due to performance concerns.


sawenzel:
Note that the size is likely not 15byte but 16 (due to padding).

sawenzel:
That being said: We could even put unsigned short for the fileID too and have the same memory consumption.

PR #548, opened 2017-04-28
Look for nanomsg on custom paths then system

dberzano:
Same as #510


dberzano:
Hmm the Travis build failed because it took too long :-/

dennisklein:
> Hmm the Travis build failed because it took too long :-/

It always times out lately after 16/17 o'clock :( I will restart it in the morning if I remember.

PR #552, opened 2017-05-03
Less string work to retrieve branchID; special cached branchID for MC…

sawenzel:
…Track

 * The FairLink management uses a lot of string queries/operations which
   have visible effects on the CPU performance

 * This commit achieves 3 separate things:
   - it makes the query string->branchID faster
   - it makes the query avoidable in many cases by introducing
     a special branchId for MCTrack (since MCTrack has a special meaning
     in the framework and is used/required in many places)
   - The new caching mechanism of the MCTrack branchID is used in FairMultiLinkedData

### DIFF
--- a/base/event/FairMultiLinkedData.cxx
+++ b/base/event/FairMultiLinkedData.cxx
@@ -132,7 +132,8 @@ void FairMultiLinkedData::AddLink(FairLink link, Bool_t bypass, Float_t mult)
 
   if (fPersistanceCheck == kFALSE ||
       link.GetIndex() < 0 ||
-      ioman->CheckBranch(ioman->GetBranchName(link.GetType())) == 0) {
+      ( link.GetType() != ioman->GetMCTrackBranchId() &&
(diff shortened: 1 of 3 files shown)


sawenzel:
@TobiasStockmanns : Can you please review these changes?

TobiasStockmanns:
Test fails because the MCTrackBranchId is wrong. It gets overwritten whenever a task tries to add a "MCTrack" branch even if it is already in the BranchNameList.

The correct code should be:

Int_t  FairRootManager::AddBranchToList(const char* name)
{
    if(fBranchNameList->FindObject(name)==0) {
        fBranchNameList->AddLast(new TObjString(name));
        fBranchSeqId++;
        if(strcmp(name, "MCTrack") == 0) {
             fMCTrackBranchId = fBranchSeqId-1;
         }
    }
    // check if we are setting the MCTrack Branch
    
    return fBranchSeqId;
}

sawenzel:
@TobiasStockmanns : Well spottet. The new code should be within the protection indeed. I will update the PR accordingly.

PR #553, opened 2017-05-04
fix rpath generation on macOS

dennisklein:
This is part of the resolution of AliceO2Group/AliceO2#343.

The code is copied from DDS, I tested it on a `10.11.6` mac. It basically adds the `LC_RPATH` entries for the dylib itself and for all dependencies, that have an install name based on `@rpath` themselves.

This issue became visible after I have cleaned up the overlinking of `libFairMQ`. It can still be masked by CMake and `DYLD_LIBRARY_PATH`, but essentially the rpath entries were missing for all the libraries in FairRoot, at least when built with `aliBuild`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,12 @@ before_install:
         env;
         brew update;
         if test "${_system_version}" = "10.11"; then
-          brew install gcc;
+          brew install --force-bottle gcc@6;
+          sudo ln -s $(brew --prefix gcc@6)/bin/gfortran-6 /usr/local/bin/gfortran;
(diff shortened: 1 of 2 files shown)


dennisklein:
do not merge yet, looking into the failing tests..

dennisklein:
2 days ago the homebrew gcc formula was updated to now track version 7.1. Looks like it breaks our fortran based dependencies. I tested by rerunning old nightly dev builds on travis osx which have passed successfully in the past and are now failing. I will try pinning the homebrew formula to gcc 6 or whatever was before.

fuhlig1:
The PR was merged with 68f2237e4f66b1268b44eee2cf2d129439aa3eee.

PR #555, opened 2017-05-06
pin gcc to version 6 on travis osx

dennisklein:


dennisklein:
closing, this commit is also contained in #553

PR #560, opened 2017-05-09
add missing link dependency

dennisklein:
Should fix `build/O2/o2-dev-fairroot` check in AliceO2Group/AliceO2#370.


dennisklein:
We really need to get `FairSoft` migrated to `aliBuild`. I can't see this kind of bugs, because `FairSoft` installs all headers in the same directory.

PR #567, opened 2017-05-15
Inproc and ipc unit tests

dennisklein:
opened for discussion

### DIFF
--- a/fairmq/FairMQDevice.h
+++ b/fairmq/FairMQDevice.h
@@ -1,19 +1,23 @@
 /********************************************************************************
- *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ * Copyright (C) 2012-2017 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
- *              This software is distributed under the terms of the             * 
- *         GNU Lesser General Public Licence version 3 (LGPL) version 3,        *  
+ *              This software is distributed under the terms of the             *
+ *              GNU Lesser General Public Licence (LGPL) version 3,             *
  *                  copied verbatim in the file "LICENSE"                       *
  ********************************************************************************/
-/**
- * FairMQDevice.h
- *
- * @since 2012-10-25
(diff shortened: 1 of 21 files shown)


dennisklein:
@ktf Could the api used in this [example](https://github.com/dennisklein/FairRoot/blob/inproc_and_ipc_unit_tests/fairmq/test/protocols/_push_pull_multipart.cxx) solve your testing usecase, you mentioned in the dropbox document? The plan is to make all of the Message factories available on `FairMQChannel`. The `inproc` addresses are zero copy and just pass the message pointers around, it can only be used within the same process. The `ipc` addresses are implemented via named pipes in the file system, so you can do IPC with it as the name suggests.

dennisklein:
@MohammadAlTurany If the travis tests succeed tomorrow and @rbx approves, this is ready to merge from my side.

fairmq/FairMQDevice.h
```diff
@@ -338,6 +332,7 @@ class FairMQDevice : public FairMQStateMachine, public FairMQConfigurable
 
     /// Creates stand-alone transport factory
     /// @param transport  Transport string ("zeromq"/"nanomsg"/"shmem")
+    [[deprecated("use 'auto FairMQTransportFactory::CreateTransportFactory(const std::string& type) -> std::shared_ptr<FairMQTransportFactory>' instead")]]
     static std::unique_ptr<FairMQTransportFactory> MakeTransport(const std::string& transport);
```
dennisklein: We found that this function was still used currently in AliceO2 code, so we just marked it deprecated. This `[[deprecate("msg")]]` attribute is standardized in C++14, but the attribute syntax `[[ ]]` was already introduced in [C++11](http://en.cppreference.com/w/cpp/language/attributes). Worst case should be, that a C++11 compiler will just ignore it (which is only guaranteed in C++17 standard though).

dennisklein:
Hm, I will fix the warnings for gcc4. It should imho only warn if you use the deprecated code.

dennisklein:
I have removed the deprecation marker and will add it in a later PR.

dennisklein:
rebased to `dev`

PR #571, opened 2017-05-17
Fixed Tutorial3 to check for FairRootManager singleton instance exist…

karabowi:
…ance.

### DIFF
--- a/base/sim/FairRootManagerSim.cxx
+++ b/base/sim/FairRootManagerSim.cxx
@@ -44,7 +44,7 @@ FairRootManagerSim::FairRootManagerSim()
   fRootManager = FairRootManager::Instance();
   if ( ! fRootManager ) {
     std::cout << "Creating FairRootManager" << std::endl;
-    fRootManager = new FairRootManager();
+      fRootManager = FairRootManager::Instance();
   } else {
     std::cout << "Got FairRootManager existing" << std::endl;
   }
--- a/base/sim/FairRootManagerSimMT.cxx
+++ b/base/sim/FairRootManagerSimMT.cxx
@@ -93,7 +93,7 @@ FairRootManagerSimMT::FairRootManagerSimMT()
   if ( ! fRootManager ) {
     LOG(ERROR) << "FairRootManager::Instance() does not exist, creating new instance in " 
       << fId << " " << this << FairLogger::endl;
-    fRootManager = new FairRootManager();
+      fRootManager = FairRootManager::Instance();
   }
 
   if ( fgDebug ) {
--- a/base/steer/FairRootManager.cxx
+++ b/base/steer/FairRootManager.cxx
@@ -71,7 +71,8 @@ FairRootManager* FairRootManager::Instance()
 {
 // Returns singleton instance.
 // ---
-  return fgInstance;
+    if (!fgInstance) fgInstance = new FairRootManager;
+    return fgInstance;
 }
 //_____________________________________________________________________________
 
--- a/base/steer/FairRootManager.h
+++ b/base/steer/FairRootManager.h
@@ -54,9 +54,7 @@ class TIterator;
 class FairRootManager : public TObject
 {
   public:
-    /**ctor*/
-    FairRootManager();
-    /**dtor*/
+      /**dtor*/
     virtual ~FairRootManager();
      Bool_t             AllDataProcessed();
     /** Add a branch name to the Branchlist and give it an id*/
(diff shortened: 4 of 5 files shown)


examples/advanced/Tutorial3/reconstruction/FairTestDetectorRecoTask.cxx
```diff
@@ -97,7 +97,9 @@ void FairTestDetectorRecoTask::Exec(Option_t* /*opt*/)
         TVector3 dpos(1 / TMath::Sqrt(12), 1 / TMath::Sqrt(12), 1 / TMath::Sqrt(12));
 
         FairTestDetectorHit* hit = new ((*fHitArray)[ipnt]) FairTestDetectorHit(-1, -1, pos, dpos);
-        hit->AddLink(FairLink(-1, FairRootManager::Instance()->GetEntryNr(), FairRootManager::Instance()->GetBranchId("FairTestDetectorDigi"), ipnt));
+        if ( FairRootManager::Instance() != NULL ) {
```
dennisklein: Prefer [`nullptr`](http://en.cppreference.com/w/cpp/language/nullptr). `NULL` is no mistake here though.

dennisklein: `FairRootManager` has actually a strange implementation of the singleton pattern. Usually, `Instance()` returns a reference and then you don't need this check.

Btw, what is the meaning of your code, if you skip the `hit->AddLink` line? Is it really optional?

dennisklein: @MohammadAlTurany The caller of `Instance()` should really not worry about the existance of the singleton object. Can we rather fix the singleton implementation of `FairRootManager`? Best would be to return a reference, but that will break existing code. The default constructor of `FairRootManager` should also be private, but that may also break existing code, I assume.

The best non-breaking solution I can think of is
```cpp
FairRootManager* FairRootManager::Instance()
{
// Returns singleton instance.
  if (fgInstance == nullptr)
  {
    fgInstance = new FairRootManager();
  }
  return fgInstance;
}
```

But even with the above change the code is not safe. The initialization in the ctor `fgInstance = this;` is dangerous. For example, if the caller of the ctor puts the object on a stack which goes out of scope before the program exits, each call to `FairRootManager::Instance()` will eventually point to random memory.

dennisklein: and note, that the above example as well as the current ctor are not thread safe.

dennisklein: ok, on some platforms they may be, if thread local storage is available.

MohammadAlTurany: The singleton implementation does not really follow the known rules (private ctor and static instance that create the object by the first call). I started a branch few months ago to correct this (This is also valid for the FairRunXXX classes) but never continue! may be it is time to correct it.
With thread local storage the thread safety issue is solved (FairGenericRootManager.cxx)

MohammadAlTurany: I just create a PR with the correct implementation for the FairRootManager (#575)

>  But even with the above change the code is not safe. The initialization in the ctor fgInstance = this; is dangerous. For example, if the caller of the ctor puts the object on a stack which goes out of scope before the program exits, each subsequent call to FairRootManager::Instance() will eventually point to random memory.
@dennisklein "Instance()" is static so this is not an issue.

karabowi: Nice discussion. The problem here was that we use the FairTestDetectorRecoTask both in FairRoot (where we have access to input tree via FairRootManager and can set the link) _and_ in FairMQ (where we do not even want to create the FairRootManager and thus cannot set the link). I was quite happy that the FairRootManager::Instance() does not create new FRM, but only returns the pointer to already existing singleton.

MohammadAlTurany: @karabowi  I agree that this create a problem in this special case. Anyway we have to:

- Find another way to decide if we are in the MQ or ROOT world
- Set the links also in the MQ world.

karabowi:
Hi to everyone.
We just has a discussion (Mohammad, Denis, Radek) about the issue, and our proposition is:
- change FairRootManager::Instance() to create singleton;
- add a member to FairTask (like Bool_t fStreamProcessing; other suggestions welcomed), which will steer the calls to the FairRootManager depending on whether it is run in FairRoot or FairMQ.
I will implement the second change tomorrow if noone objects after Mohammad merges his pull req.

PR #574, opened 2017-05-17
Various Coverity fixes

dennisklein:


fairmq/nanomsg/FairMQSocketNN.cxx
```diff
@@ -82,7 +82,10 @@ FairMQSocketNN::FairMQSocketNN(const string& type, const string& name, const str
 
 #ifdef NN_RCVMAXSIZE
     int rcvSize = -1;
-    nn_setsockopt(fSocket, NN_SOL_SOCKET, NN_RCVMAXSIZE, &rcvSize, sizeof(rcvSize));
+    if (nn_setsockopt(fSocket, NN_SOL_SOCKET, NN_RCVMAXSIZE, &rcvSize, sizeof(rcvSize)) != 0)
```
ktf: Shouldn't this throw, actually?

rbx: It could, but I think error message is enough. The transport is still usable if this fails, with max message size of 1024kB.

PR #575, opened 2017-05-18
Correct singleton implimentation for FairRootManager

MohammadAlTurany:
Make the ctor of FairRootManager private

### DIFF
--- a/base/steer/FairRootManager.cxx
+++ b/base/steer/FairRootManager.cxx
@@ -71,7 +71,8 @@ FairRootManager* FairRootManager::Instance()
 {
 // Returns singleton instance.
 // ---
-  return fgInstance;
+    if (!fgInstance) fgInstance = new FairRootManager;
+    return fgInstance;
 }
 //_____________________________________________________________________________
 
--- a/base/sim/FairRootManagerSim.cxx
+++ b/base/sim/FairRootManagerSim.cxx
@@ -44,7 +44,7 @@ FairRootManagerSim::FairRootManagerSim()
   fRootManager = FairRootManager::Instance();
   if ( ! fRootManager ) {
     std::cout << "Creating FairRootManager" << std::endl;
-    fRootManager = new FairRootManager();
+      fRootManager = FairRootManager::Instance();
   } else {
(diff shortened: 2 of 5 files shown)


fuhlig1:
The problem here is that the change might break a lot of user code. In each macro you have code like   `FairRunSim* run = new FairRunSim();` which will not work any longer. Unfortunately I don't see any way to use the old (wrong) and the new (correct) implementation of the singleton with the same user code.

dennisklein:
The danger with public ctor is when people would do something like
```cpp
void Init() {
  FairRootManager ioman;
  mgr.Register(...);
} // ioman goes out of scope and the singleton object is destructed.

void Exec() { // is called after Init()
  FairRootManager* ioman = FairRootManager::Instance();
  // without this PR: ioman points to random memory
  // with fixed Instance() only: ioman in Init and in Exec are different objects
}
```

As proposed in this PR the above usage is not possible and also `Instance()` can never return `nullptr`. If you want the safer interface we have to hide the ctors from user code.

I believe the changes in this PR will fix the singleton implementation. I would still also remove the assignment in the now private ctor, it is no longer needed.

If this would be massive breakage and you are confident people will most of the time put their managers on the heap, then it might be cheapest overall to leave the ctor public..

base/steer/FairRootManager.cxx
```diff
@@ -71,7 +71,8 @@ FairRootManager* FairRootManager::Instance()
 {
```
dennisklein: If you decide to break user code by hiding the ctor now anyways, you might consider returning a `FairRootManager&` here as well, which is a cleaner interface because it must point to a valid object which the singleton factory guarantees.
```cpp
FairRootManager& FairRootManager::Instance() {
  if (!fgInstance) fgInstance = new FairRootManager;
  return *fgInstance;
}
```

MohammadAlTurany: With this PR no user code is broken at all, the IO Manager is created by the Framework and the user are accessing it via the FairRootManager::Instance() which is still returning a pointer as before. I think if we decide to change all the other Singletons (which will definitely break the user code!) then we improve this further.

Issue #573, opened 2017-05-17
[FairMQ] Add documentation


dennisklein:
- [x] Extend the Transport matrix for supported address formats.
- [x] Add a section which describes the various Message factories.
- [x] Add a section which gives examples and hints on how to write unit tests
- [x] Document NewPoller factory.
- [x] Document Channel.
- [x] Add Development/Testing/MultipleDevices section.
- [x] Add description of the JSON format for FairMQParser.
- [x] Document ParameterMQServer.
- [x] Extend threading documentation (state handlers run in their own thread).
- [x] Document common devices.
- [x] Document introspection options.

rbx:
Close for now, since all the above docs are now added. Can re-open and extend the list if something else becomes available.

PR #580, opened 2017-05-22
Update docs and add more test coverage

dennisklein:
work in progress for #573, open for discussion


fairmq/nanomsg/FairMQTransportFactoryNN.cxx
```diff
@@ -63,7 +63,8 @@ FairMQPollerPtr FairMQTransportFactoryNN::CreatePoller(const FairMQSocket& cmdSo
 
 void FairMQTransportFactoryNN::Shutdown()
 {
-    nn_term();
+    // nn_term();
+    // see https://www.freelists.org/post/nanomsg/Getting-rid-of-nn-init-and-nn-term,8
 }
```
rbx: Does it cause any issues? I would prefer to leave `nn_term()` in. Even though timeouts on socket transfer operations should make sure none of our sockets hang when the state machine wants to terminate, better to be on the safe side. It is same semantic as `zmq_ctx_shutdown()`. The post by Sustrik is 3 years old, things might have changed or change in the future. They still have it documented and used in their tests.

dennisklein: Yes, it causes a nn test not to runnable twice. Once you call `nn_term()`, you cannot ever create a socket any more.

rbx: So it's like shutting down the context, but in nanomsg case without a way to re-create it. We only do this at device exit. I guess the issue is when having multiple devices (or standalone transports) in a process?

dennisklein: In the tests I create and destroy multiple nn transports in series. Before this PR it worked because I did not call `Shutdown`. But I need to call it for shmem and for correctness.

dennisklein: There used to be `nn_init`, but it's gone. `nn_term` is just optional.

rbx: I guess we can try leaving it out, but need to keep an eye that it really doesn't cause any termination failures.

dennisklein: Have a look how it is implemented. It just sets a global flag. Havent checked the socket yet, but cant imagine it can do too much.

rbx: Yeah, I saw it, also didn't look deeper how they use the flag. But anyway, the fact that it is documented as optional for something that we solve differently makes it fine to remove. None of our sockets hang if device wants to exit. This call is just a secondary safety. And the fact that they have it global for all sockets will give us trouble if we need more than one context.

PR #581, opened 2017-05-23
FairTask - added new member variable (fStreamProcessing), set by default

karabowi:
to kFALSE. When running in FairMQ, it should be SetStreamProcesssing(kTRUE).

Modified examples/advanced/Tutorial3, examples/MQ/9-PixelDetector and examples/MQ/9a-PixelDetector
accordingly.


dennisklein:
This is a great article on the conventions on how to format git commit messages: https://chris.beams.io/posts/git-commit/

tldr; write a single header line with about 50-60 characters, then put an empty line and then add your detailed description.

PR #588, opened 2017-05-26
Always have access to FairMQProgOptions from device

rbx:
- Create FairMQProgOptions within a device if it has been started without access to one.
- Test devices without supplied config & two in a process.

### DIFF
--- a/fairmq/FairMQDevice.cxx
+++ b/fairmq/FairMQDevice.cxx
@@ -21,6 +21,10 @@
 
 #include <boost/algorithm/string.hpp> // join/split
 
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_generators.hpp>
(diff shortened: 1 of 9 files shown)


dennisklein:
Do you plan to add the `TestReceiver.cxx` and `TestSender.cxx` to the helper binary too? If not, I would move the files just inside the `fairmq/test/device` folder since they would not be related to the helper.

dennisklein:
And we should really rename the `.cxx` files to `.h`. I messed this up in the helper.

rbx:
yeah, i'll move to `fairmq/test/device`, will be cleaner that way.

rbx:
Moved the test devices.

And trying if changing `std::async(...)` to `std::async(std::launch::async, ...)` helps with the failing tests.

PR #605, opened 2017-07-03
base/MQ and Lmd example cleanup

rbx:
Cleanup of base/MQ and Lmd example for FairMQProgOptions use, naming and formatting.


rbx:
rebased/solved conflicts.

PR #606, opened 2017-07-06
Add version member to FairMQDevice, settable via constructor, connect to `--version`

rbx:


fairmq/runFairMQDevice.h
```diff
@@ -73,6 +73,11 @@ int main(int argc, const char** argv)
 
         int result = runStateMachine(*device, config);
 
+        if (config.Count("version"))
+        {
+            pluginManager->ForEachPlugin([](fair::mq::Plugin& plugin){ std::cout << "plugin: " << plugin.GetName() << ": " << plugin.GetVersion() << std::endl; });
```
dennisklein:
```cpp
std::cout << "plugin: " << plugin << std::endl;
```
would give you the same string I also print in the debug log, it contains also maintainer and homepage. If the formatting is not good, you can change it in the overloaded `operator<<` in `Plugin.h`. I would prefer this unless you don't want maintainer and homepage in the `--version` output.

fairmq/tools/runSimpleMQStateMachine.h
```diff
@@ -30,6 +32,15 @@ inline int runStateMachine(TMQDevice& device, FairMQProgOptions& cfg)
         return 0;
     }
 
+    if (cfg.Count("version"))
+    {
+        std::cout << "User device version: " << device.GetVersion() << std::endl;
+        std::cout << "FAIRMQ_INTERFACE_VERSION: " << FAIRMQ_INTERFACE_VERSION << std::endl;
```
dennisklein: Remind me which interface `FAIRMQ_INTERFACE_VERSION` is versioning again?

fairmq/tools/runSimpleMQStateMachine.h
```diff
@@ -30,6 +32,15 @@ inline int runStateMachine(TMQDevice& device, FairMQProgOptions& cfg)
         return 0;
     }
 
+    if (cfg.Count("version"))
+    {
+        std::cout << "User device version: " << device.GetVersion() << std::endl;
```
dennisklein: If we do not print the FairRoot version here, we might call this just `device version`, what do you think?

fairmq/tools/runSimpleMQStateMachine.h
rbx: It is the only FairMQ version we have now, incremented when there are breaking changes. We should replace this with a semantic version at some point, unless we stick to FairRoot version.

fairmq/runFairMQDevice.h
rbx: Sure, sounds good, i'll change it as you suggest.

fairmq/tools/runSimpleMQStateMachine.h
rbx: I think "User ..." is more clear because it prints the version user gives to a device that he derives from FairMQDevice, maybe a better name?. The FairMQDevice version is currently FAIRMQ_INTERFACE_VERSION, which is also printed, but this we can replace with FairRoot version or our own FairMQ semver.

PR #620, opened 2017-08-24
Functional tests for MQ examples

rbx:
Add functional tests for MQ examples.
For long-running examples stop conditions are added for use by the tests.


rbx:
Investigating the failed tests.

PR #636, opened 2017-09-26
Ability to register/retrieve data of any type with IO manager

sawenzel:
This commit provides a major extension to the FairRootManager, enabling
the user to register branches based on arbitrary types.
In particular, it is now possible to register stl containers as branch
entries, extending the current TObject* or TCollection* APIs.

Instant gratification:
```
// some source registering data
auto ioman = FairRootManager::Instance();
std::vector<CustomData>* data;
ioman->RegisterAny("mybranch", data, ...);
```

```
// some consumer task asking for the data
auto ioman = FairRootManager::Instance();
auto mydata = ioman->InitObjectAs<std::vector<CustomData> const*>("mybranch");
```

The example "simulation/Tutorial2" was extended accordingly showing use cases
with persistent and non-persistent branches.

The IO to ROOT trees works as before. It was checked that splitting of branches
based on std::vectors still works the same way as for TClonesArrays.

The user has to create dictionaries for any entry she wants to put into
a persistent branch.

Further notes:
  a) the new mechanism does not use folders; Hence the persistent branches
     have to be created explictly (done in FairMCApplication for the moment)
  b) the commit is incompletely in many ways; It demonstrates correct working
     for a simple FileSource
  c) Existing API is fully preserved. The user can use the traditional API as well
     as the new API at the same time.
  d) The new API imposes that input data obtained from InitObjectAs is unmodifiable

### DIFF
--- a/base/sim/FairMCApplication.cxx
+++ b/base/sim/FairMCApplication.cxx
@@ -1080,6 +1080,9 @@ void FairMCApplication::InitGeometry()
     rootManager->TruncateBranchNames(outTree, "cbmroot");
     rootManager->SetOutTree(outTree);
 
+    // create other branches not managed by folder
+    rootManager->CreatePersistentBranchesAny();
+
     // Explicit registration of stack in MT mode
     // (should be removed when problem with "Register" is resolved)
     // Create branch in outTree manually
--- a/base/source/FairFileSource.cxx
+++ b/base/source/FairFileSource.cxx
@@ -891,7 +891,7 @@ void FairFileSource::ReadBranchEvent(const char* BrName)
 //_____________________________________________________________________________
 void   FairFileSource::ReadBranchEvent(const char* BrName, Int_t Entry)
 {
-
+    fCurrentEntryNo = Entry;
     if ( fInTree ) {
         fInTree->FindBranch(BrName)->GetEntry(Entry);
         fEventTime = GetEventTime();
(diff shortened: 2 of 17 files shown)


sawenzel:
@MohammadAlTurany , @fuhlig1, @matthiasrichter : Please let me know what you think. This PR might need further work but I consider it a fair foundation for discussions/review. It adds a feature I discussed a lot in the past and which will help reduce memory footprint for ALICE.

matthiasrichter:
Looks very nice. As just discussed, not all the things are obvious because of peculiarities of ROOT I/O, that's why I would add a few comments to the examples why the pointer-to-pointer must be used.

sawenzel:
@MohammadAlTurany, @fuhlig1 : Can you please take a look?

MohammadAlTurany:
Hallo @sawenzel 
if you replace the line:

`fAnyBranchMap[brname]=std::make_unique<TypeAddressPair const>(ot, pt, (void*)obj);
`
with:
`fAnyBranchMap[brname]=std::unique_ptr<TypeAddressPair const> (new TypeAddressPair(ot, pt,(void*)obj));`
Then it will work without C++14, FairRoot is still requiring only c++11.

Mohammad

sawenzel:
@MohammadAlTurany : Yes will make this change. Thanks.

sawenzel:
@MohammadAlTurany : I updated the PR

sawenzel:
@MohammadAlTurany , @rbx : I took out `#include FairLogger.h` from FairRootManager.h which was the origin of the LOG shadowing problem. Now this PR should compile fine.

sawenzel:
Added one commit making timing set after FairFileSource::ReadBranchEvent consistent with the behaviour of FairFileSource::ReadEvent; Please review and let me know if this is correct.

Issue #645, opened 2017-10-11
Improve protobuf detection


ktf:
At the moment the only way to correctly specify protobuf is to do something along the lines of:

```
  -DPROTOBUF_INCLUDE_DIR=$(dirname $(which protoc))/../include                \
  -DPROTOBUF_PROTOC_EXECUTABLE=$(which protoc)                                \
  -DPROTOBUF_LIBRARY=$(dirname $(which protoc))/../lib/libprotobuf.$SONAME    \
```

This is because at the moment the lookup strategy excludes system locations and the above is the only way to have something which works both in the case of our own protobuf and for the system one. Would it be possible to change things so that:

- If `ProtoBuf_DIR` is defined, protobuf is looked up there (and the build fails if not found).
- If `ProtoBuf_DIR` is not defined, something like the above logic is used to determine the system protobuf.

dennisklein:
#632 will address this issues. I enables lookup in system paths and if `PROTOBUF_ROOT` is defined, it searches there first.

After the next FairRoot release (this month) we will merge #632 into `dev`. Can this wait until then or do you need it faster? Btw, #632 will change the logic for all dependencies as described above.

MohammadAlTurany:
Hi Giulio,
Could you check if the PR #646 solve this issue

dennisklein:
The upstream FindProtobuf.cmake does not offer a hint variable, that's why it was copied in the past and why it is synched with upstream in #632.

MohammadAlTurany:
solved in  #652

PR #646, opened 2017-10-11
use Findprotobuf from cmake instead of our own

MohammadAlTurany:
Use the Findprotobuf delivered by Cmake instead of our self-written one. Check if this will solve the issue #645


MohammadAlTurany:
As this is solved better in another PR (#632) I close it.

PR #652, opened 2017-10-12
Look for protobuf on custom paths then system

dberzano:
* Enforce presence of `protoc` (the protobuf compiler) too
* Similar to #510 and #548

### DIFF
--- a/cmake/modules/FindProtobuf.cmake
+++ b/cmake/modules/FindProtobuf.cmake
@@ -136,17 +136,13 @@ endfunction()
 function(_protobuf_find_libraries name filename)
    find_library(${name}_LIBRARY
        NAMES ${filename}
-       PATHS ${SIMPATH}/lib ${ProtoBuf_DIR}/lib
-       NO_DEFAULT_PATH
-       NO_CMAKE_SYSTEM_PATH
+       HINTS ${SIMPATH}/lib ${ProtoBuf_DIR}/lib
    )
    mark_as_advanced(${name}_LIBRARY)
 
    find_library(${name}_LIBRARY_DEBUG
        NAMES ${filename}
-       PATHS ${SIMPATH}/lib ${ProtoBuf_DIR}/lib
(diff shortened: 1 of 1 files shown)


dennisklein:
@dberzano I have no objections here, but just to give a headsup what will come soon: In #632 I already changed the behaviour of all find modules to also search in system paths. Also I introduced a CMake variable matching the exported variable in your alidist recipes. So, in this case instead of `ProtoBuf_DIR` I used `PROTOBUF_ROOT`.

dberzano:
Thanks @dennisklein I saw your PR. Just to be clear:

CMake's _de facto_ convention (_i.e._ for most `FindPackageName`s) is to pass `PackageName_DIR` variables, not `PACKAGENAME_ROOT` ones, and to rely on what you explicitly pass (_e.g._ with `-D`) as opposed to picking things from the environment.

What I think is that:

* `ProtoBuf_DIR` and all other `PackageName_DIR` should be kept as they are and not deprecated in favour of `PACKAGENAME_ROOT` ones.
* No `PACKAGENAME_ROOT` variable should be picked from the environment when configuring a project (following the principle of "explicit is better than implicit").

My two cents :-) I'll keep an eye on #632!

dennisklein:
> (...) as opposed to picking things from the environment.

> No PACKAGENAME_ROOT variable should be picked from the environment when configuring a project (following the principle of "explicit is better than implicit").

I fully agree with the notion, but it might not be possible to support all the needs of our clients by not picking up dependency hints from the environment. I will try to make everyone aware of this issue again and we will see, if I can convince them to support your requirement.

> CMake's de facto convention (i.e. for most FindPackageNames) is to pass PackageName_DIR variables

I was not aware of this convention, thx. I see no reason not to support both variable naming schemes. If people agree to waive environment pickups, I would probably drop the aliBuild variable naming scheme again from #632 and stick with the CMake convention only.

PR #656, opened 2017-10-17
Add paths to HINTS in FindProtobuf only if defined

dberzano:
Fixes a problem where undefined variables resulted in replacing search paths as
`${SIMPATH}/lib` to `/lib`, leading to unintended behaviours. This patch also
uses _either_ `${SIMPATH}` or `${ProtoBuf_DIR}` as hint, and only if they are
found as directories, in order to avoid cross-picking protobuf components. If
none is set, everything will be obtained from system paths, consistently, as
`HINTS` will be empty.

### DIFF
--- a/cmake/modules/FindProtobuf.cmake
+++ b/cmake/modules/FindProtobuf.cmake
@@ -136,13 +136,15 @@ endfunction()
 function(_protobuf_find_libraries name filename)
    find_library(${name}_LIBRARY
        NAMES ${filename}
-       HINTS ${SIMPATH}/lib ${ProtoBuf_DIR}/lib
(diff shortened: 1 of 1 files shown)


dennisklein:
What do you think about the following implementation?

```cmake
find_library(${name}_LIBRARY
  NAMES ${filename}
  HINTS ${SIMPATH} ${ProtoBuf_DIR}
  PATH_SUFFIXES lib
)
```

I believe it does what we want (slightly more, because it searches the HINTS directory and HINTS/lib), but would be much shorter.

dberzano:
I was not aware of the `PATH_SUFFIXES` option! @dennisklein I always learn something new from you :-) I'll update the PR soon.

PR #665, opened 2017-10-24
Add alfa-ci build jobs

dennisklein:
In principle this is how it could look like. What do you think?

Note: When a requested label in the build matrix is not available on any online build agent, the build job will be queued indefinitely by jenkins.

TODO
* [x] Make sure the build stages fails, if one of the unit tests failed. As one can [see here](https://alfa-ci.gsi.de/blue/organizations/jenkins/FairRootGroup%2FFairRoot/detail/PR-663/71/pipeline/12) or in the yellow box below, all build agents had failing tests, but some reported success.
* [x] Observe agent disk usage. I believe builds are currently not cleaned up. This could lead to excessive disk usage over time. Find a plugin or a config option to add cleanup.
* [x] ~Find out, if one can retrieve the submit id of cdash. Would be nicer to have the github links directly point to the right cdash page instead of just the dashboard.~ It is not possible, see https://cmake.org/pipermail/cmake/2014-September/058538.html

### DIFF
--- a/Dart.sh
+++ b/Dart.sh
@@ -69,14 +69,16 @@ chip=$(uname -m | tr '[A-Z]' '[a-z]')
 # environment variables used by ctest
 SYSTEM=$arch-$chip
 if test -z $CXX ; then
-  COMPILER=gcc;
-  GCC_VERSION=$(gcc -dumpversion)
+  if [ "$arch" == "darwin" ]; then
+    COMPILER=$(clang --version | head -n 1 | cut -d' ' -f1,2,4 | tr -d ' ')
+  else
+    COMPILER=gcc$(gcc -dumpversion)
+  fi
 else
-  COMPILER=$CXX;
-  GCC_VERSION=$($CXX -dumpversion)
+  COMPILER=$CXX$($CXX -dumpversion)
 fi
 
-export LABEL1=${LINUX_FLAVOUR}-$SYSTEM-$COMPILER$GCC_VERSION-fairroot_$GIT_BRANCH-fairsoft_$FAIRSOFT_VERSION
+export LABEL1=${LINUX_FLAVOUR}-$chip-$COMPILER-FairRoot_$GIT_BRANCH-FairSoft_$FAIRSOFT_VERSION
 export LABEL=$(echo $LABEL1 | sed -e 's#/#_#g')
 
 # get the number of processors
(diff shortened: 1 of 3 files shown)


dennisklein:
It would be nicer to split the labels into its components. But then we would need some way, that the build job can actually set this environment in a portable way. I don't want to have host specific paths in the Jenkinsfile.

MohammadAlTurany:
We also need to test the different configurations:

1. Standard
2. with DDS
3. Reco only (Switch off the simulation)

fuhlig1:
Concerning your last point about linking github to cdash. Another solution would be to link to the corresponding Jenkins page.

dennisklein:
> for the time being I would also keep the travis build to compare results. I would remove the travis config later.

Done.

> I see the results on CDash which is what we want. On the other hand I would like to check the Logs on the Jenkins-Server. Unfortunately I wasn't able to find them. After I wrote this I found them. For me it looks like the stuff is only displayed when using the Blue Ocean Plugin.

> Concerning your last point about linking github to cdash. Another solution would be to link to the corresponding Jenkins page.

By default, jenkins only reports a single cummulative status to github called **continuous-integration/jenkins/pr-head**. Its Details link points always to the jenkins web page.

The status of the form **alfa-ci/...** I have added. They report the status of each parallel build job. Once they finished, I set the Details link to the cdash FairRoot dashboard currently. The idea is, that one can see immediately which build jobs failed in the github interface and jump right into the logs on cdash, or alternatively - by clicking on the cummulative status - to the jenkins logs (from there you can select each build job by clicking the symbols in the pipeline diagram).

So, you propose to just have all Details links point to the jenkins logs? Then I would go further and just remove the github notifications per build job and just leave the cummulative one. What do you think?

dennisklein:
I patched `ctest`, which ignored the `NO_PROXY` environment variable (lxbuild05 could not send reports to cdash). The PR is pending: https://gitlab.kitware.com/cmake/cmake/merge_requests/1409.

@MohammadAlTurany I discussed today with @fuhlig1 how to get the multiple configurations tested as you suggested. We will add those incrementally, once we setup more build agents and defined a proper build matrix.

dennisklein:
I will now look also into why some tests are failing, but these are seperate issues. I will open separate PRs.

PR #669, opened 2017-10-27
Install DDS plugin

dennisklein:
Some small fixes to properly install DDS plugin and handle the case where there are multiple default routes in linux (happens on my notebook, when wired and wireless connected at the same time).


dennisklein:
Not sure how to get line buffered/flushed log output on MacOS 10.11. Is there a system setting on mac for such things?

PR #671, opened 2017-11-01
Fix broken tests

dennisklein:
Some of the tests produce characterwise interleaved output which breaks
the ctest condition check. This commit serializes the output.


dennisklein:
does not work

PR #679, opened 2017-11-20
Update for MT simulation

ihrivnac:
- Removed use of FairGenericRootManager, FairRootManagerSim and FairRootManagerSimMT
  and merged locking for thread-unsafe code in FairRootManager
- Fixed filling of sensistive volume map (fVolMap) in FairMCApplication in MT mode
- Made static data in FairModule thread local
- Code clean-up in FairRootManager, FairMCApplication, FairModule
  (Fixed indentions, removed tabulations)

### DIFF
--- a/base/steer/FairRootManager.cxx
+++ b/base/steer/FairRootManager.cxx
@@ -69,21 +69,13 @@ using std::pair;
 using std::set;
 
 namespace {
-  // Define mutexes per operation for the functions
-  // which are supposed to be called from VMC applications
+  // Define mutexes per operation which modify shared data
   TMCMutex createMutex = TMCMUTEX_INITIALIZER;
   TMCMutex deleteMutex = TMCMUTEX_INITIALIZER;
-  TMCMutex registerMutex = TMCMUTEX_INITIALIZER;
-  TMCMutex fillMutex  = TMCMUTEX_INITIALIZER;
-  TMCMutex tmpFillMutex  = TMCMUTEX_INITIALIZER;
-  TMCMutex writeMutex = TMCMUTEX_INITIALIZER;
-  TMCMutex closeMutex = TMCMUTEX_INITIALIZER;
 }
 
 //_____________________________________________________________________________
 Int_t   FairRootManager::fgCounter = 0;
-Bool_t  FairRootManager::fgIsFillLock = true;
-std::vector<Bool_t>* FairRootManager::fgIsFillLocks = 0;
 
 //_____________________________________________________________________________
 FairRootManager* FairRootManager::Instance()
@@ -133,30 +125,23 @@ FairRootManager::FairRootManager()
     fListOfBranchesFromInputIter(0),
     fListOfNonTimebasedBranches(new TRefArray()),
     fListOfNonTimebasedBranchesIter(0),
-    fId(0),
-    fDebugMT(false)
+    fId(0)
 {
-  if ( fDebugMT ) {
-    LOG(INFO) << "FairRootManager::FairRootManager: going to lock "
-      << this << FairLogger::endl;
-  }
+  LOG(DEBUG) << "FairRootManager::FairRootManager: going to lock "
+    << this << FairLogger::endl;
+
   TMCAutoLock lk(&createMutex);
 
   // Set Id
   fId = fgCounter;
 
   // Increment counter
-  if ( ! fgCounter ) {
-    fgIsFillLocks = new std::vector<Bool_t>();
-  }
   ++fgCounter;
-  fgIsFillLocks->push_back(true);
 
   lk.unlock();
-  if ( fDebugMT ) {
-    LOG(INFO) << "Released lock and done FairRootManager::FairRootManager in "
-      << fId << " " << this << FairLogger::endl;
-  }
+
+  LOG(DEBUG) << "Released lock and done FairRootManager::FairRootManager in "
+    << fId << " " << this << FairLogger::endl;
 }
 
 //_____________________________________________________________________________
@@ -164,12 +149,6 @@ FairRootManager::~FairRootManager()
 {
 //
   LOG(DEBUG) << "Enter Destructor of FairRootManager" << FairLogger::endl;
-
-  if ( fDebugMT ) {
-    LOG(INFO) << "FairRootManager::~FairRootManager: going to lock "
-      << fId << " " << this << FairLogger::endl;
-  }
-  TMCAutoLock lk(&deleteMutex);
 
   // delete fOutTree;
   if(fOutFile) {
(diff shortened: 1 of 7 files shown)


base/steer/FairRootManager.cxx
```diff
@@ -139,8 +194,109 @@ FairRootManager::~FairRootManager()
...
+  LogMessage("Done Fill");
+
+  if ( fgIsFillLock ) {
+    // the access to TFile and TTree needs to be locked only until
+    // __after__ the first Fill
+    (*fgIsFillLocks)[fId] = false;
+    Bool_t isDoneAll = true;
+    Int_t counter = 1;
+    while ( isDoneAll && counter < fgCounter ) {
+      if ( (*fgIsFillLocks)[counter++] ) {
```
dennisklein: I would like to avoid the post-increment operator here (it creates a copy of counter and it is harder to read). Could you please change it to something like:

```cpp
  if ( (*fgIsFillLocks)[counter] ) {
    isDoneAll = false;
    break;
  }
  ++counter;

File: base/steer/FairRootManager.cxx [hunk 2]
```diff
@@ -67,6 +70,24 @@ using std::list;
...
+namespace {
+  TMCMutex createMutex2 = TMCMUTEX_INITIALIZER;
+  TMCMutex deleteMutex2 = TMCMUTEX_INITIALIZER;
+  TMCMutex registerMutex2 = TMCMUTEX_INITIALIZER;
+  TMCMutex fillMutex2  = TMCMUTEX_INITIALIZER;
+  TMCMutex tmpFillMutex2  = TMCMUTEX_INITIALIZER;
+  TMCMutex writeMutex2 = TMCMUTEX_INITIALIZER;
+  TMCMutex closeMutex2 = TMCMUTEX_INITIALIZER;
+  TMCMutex getMutex2 =  TMCMUTEX_INITIALIZER;
+}
```
dennisklein: 1. Could you please add some comments which document which shared data these mutexes are intended for? From the names they seem to be tied to the functions, which does not make much sense to us at the moment.

2. Is there any particular reason against using `std::mutex` (and `std::unique_lock`/`std::lock_guard` for the locks)?

base/steer/FairRootManager.cxx
```diff
@@ -139,8 +194,109 @@ FairRootManager::~FairRootManager()
...
+    LOG(INFO) << message << " in " << fId << " " << this << FairLogger::endl;
+  }
+}
+
+//_____________________________________________________________________________
+void  FairRootManager::FillWithTmpLock()
+{
+/// Fill the Root tree.
+
+  LogMessage("Going to lock for Fill");
```
dennisklein: Can you please move the log message inside the `#if .. #endif`? There are multiple places where this needs to be applied. Thank you.

base/steer/FairRootManager.cxx
ihrivnac: There is added a comment on the top of  mutexes definition; I have also changed the mutexes names as the '2'  was added in a different use case and makes no sense here.
This code is needed in the context of Geant4 MT simulation; there is used the same technology and locking method as in Geant4; see more in:
http://geant4.web.cern.ch/geant4/UserDocumentation/UsersGuides/ForToolkitDeveloper/html/ch02s14.html
It is planned to move to C++11 thread functions in Geant4 next year, so my plan is to follow then this development in Geant4 VMC. So for the time being I prefer to keep the same technique.

base/steer/FairRootManager.cxx
ihrivnac: Ok, it is changed.

base/steer/FairRootManager.cxx
rbx: Can all these mutexes be replaced just by a single one? Or am I missing something?

dennisklein: To illustrate a bit more: Access to the member `TTree* FairRootManager::fOutTree` is synchronized with different mutexes, for example `fillMutex`, `fillTmpMutex`, and `deleteMutex`. This does not make sense to us. This implementation will still have data races, if one thread call `FairRootManager::FillWithLock` and another `FairRootManager::FillTmpLock` at the same time.

Unless, of course, you make sure that this never happens from the caller's site. But if this external synchronisation is in place, then you can just use a single mutex in all functions without a penalty (and it would be thread-safe without external synchronisation).

The only reason to have multiple mutexes really is, if you have independent data members that can be locked separately.

base/steer/FairRootManager.cxx
```diff
@@ -296,12 +491,20 @@ TClonesArray* FairRootManager::Register(TString branchName, TString className, T
     outputArray = fActiveContainer[branchName];
     Register(branchName, folderName, outputArray, toFile);
   }
+
+  LogMessage(TString("Done Register (3)") + branchName);
+#if ( ROOT_VERSION_CODE >= ROOT_VERSION(6,9,3) )
```
rbx: If I understand correctly, if ROOT version is below of what you check here, the code will have races.

Feel free to bump the required version of ROOT in the main CMakeLists.txt to 6.10.0:
```
find_package(ROOT 6.00.00 REQUIRED) 
```
to

```
find_package(ROOT 6.10.00 REQUIRED) 
```
Then you can remove all the version checks in the code.

base/steer/FairRootManager.cxx
ihrivnac: First, please, note that in MT applications, there is one Root file and tree defined per thread, so we do not access the same three object by the operation, but its Root internal globals which gets modified in these operations. The problem with a global lock is that we get easily dead-locks. This mechanism of very local locks works fine in tests.
And also, the functions FillWithLock() and FillWithTmpLock() are private, the only public functions are Fill() and FillImpl(). When using just FillImpl(), no locks are applied; this function is used in MC application when running with sequential MCs.

base/steer/FairRootManager.cxx
ihrivnac: Ok, I was not sure, if you are ok to set the minimal version to 6.10.00 globally.
The ifs inside code would keep the class compiling with older Root versions, but it would not be usable with MT. Do you want I commit this change in the branch?

base/steer/FairRootManager.h
```diff
@@ -385,6 +396,24 @@ class FairRootManager : public TObject
...
+    // methods
+    void  LogMessage(const TString& message);
+    void  FillWithLock();
+    void  FillWithTmpLock();
+    void  FillWithoutLock();
+
+    // global static data members
+    static  Int_t    fgCounter;         // The counter of instances
+    static  Bool_t   fgIsFillLock;      // The if the Fill should be locked
+    static  std::vector<Bool_t>* fgIsFillLocks; // The info per thread if the Fill should be locked
```
sawenzel: does this need to be a pointer?

base/steer/FairRootManager.cxx
```diff
@@ -204,21 +341,28 @@ TFile* FairRootManager::OpenOutFile(const char* fname)
...
 template<typename T>
-void FairRootManager::RegisterImpl(const char* name, const char *folderName, T* obj, Bool_t toFile) {
+void FairRootManager::RegisterImpl(const char* name, const char *folderName, T* obj, Bool_t toFile)
+{
+/// a common implementation for Register
+
   FairMonitor::GetMonitor()->RecordRegister(name,folderName,toFile);
 
+  LogMessage("Going to lock for Register (1)");
+  TMCAutoLock lk(&registerMutex);
```
sawenzel: What is the advantage of using `TMCAutoLock` versus a lock from the C++11 standard? In the latter case one could use a resource managing paradigm (RAII) that unlocks the code at function exit automatically. See here for a discussion:

http://en.cppreference.com/w/cpp/language/raii

sawenzel:
Does this code already protect the new `RegisterAny` functions? I guess this should also be done in the same PR.

base/steer/FairRootManager.cxx
dennisklein: The answer is already in this thread in one of the folded review subthreads above. As far as I could see [`TMCAutoLock`](https://root.cern.ch/doc/master/TMCAutoLock_8h_source.html) is already RAII capable.

base/steer/FairRootManager.h
ihrivnac: This gives us full control on creating/deleting this object; see also the O2 coding guidelines (Global variables initialization) which explicitly forbid static data members of std::vector type.

ihrivnac:
Yes, RegisterAny is protected.

base/steer/FairRootManager.h
sawenzel: Ok ... but do you need full control? Do you prefer calling "delete" vs let the compiler clean it up?

I could do not find a statement that forbids static std::vector ... (if the O2 coding guidelines apply at all for FairRoot).

MohammadAlTurany: just to clarify things: The O2 coding guidelines does not apply at all to FairRoot

ihrivnac: Yes, sure, but the rule was added to the guidelines for a good reason, and this reason is also explained here. See Decision, and then the example in extra details section.
So even if the O2 guidelines do not apply to FairRoot, I would prefer to keep this code as it is.

ihrivnac:
Hi,
Is there any action expected from me? If so, let me know, please.
The links in "Details" point to  cdash with all tests, so I don't know how to find the failures relevant to my PR.
Thank you,

MohammadAlTurany:
Hi Ivana,
No, nothing needed from your side at the moment! I installed G4 MT and would like to do some tests before merging this PR.
best,
Mohammad

ihrivnac:
Ok, thanks. But note that to get O2 working with MT there will be needed another pull request for O2, which I have on hold, as it will require this update in FairRoot. But what should be tested in FaiRoot with this PR is that it does not break your tests in sequential mode, event when running with Geant4 built with MT enabled.

MohammadAlTurany:
Hi Ivana,
It seems that we have a deadlock when we run the time based simulation with your PR , I am trying to localise the problem and see how to solve it! I see this when I use the Geant4 without multithreading.
Before we solve this I cannot merge this PR.

base/steer/FairRootManager.cxx
dennisklein: Just for my curiosity, do you know which Root internal globals are not thread-safe? Is there any chance we could apply the locking further down the call stack to achieve even better lock locality?

What percentage of cycles used by `TTree::Fill()` are actually spent on accessing the Root internal globals? I would imagine we are quite wasteful by locking the whole function call at FairRoot level, aren't we?

ihrivnac:
Hi Mohammad,
In fact this system of locking was developed three years ago, when VMC was migrated to MT, and as it was working fine with the VMC tests, it was never reviewed since then.
 I tried to run now  O2 MT simulation without these  locks and everything seems to run fine. I tested a few runs of 100 events with 16 threads. (That confirms what ROOT reports about making a progress in thread-safety, especially for core libraries.)  The only locks, which should stay are those, which modify the fgCounter static data, but these should be safe. I will make a clean-up of the class and commit the update in the branch.

MohammadAlTurany:
Hi Ivana, 
Great! Thanks I will wait for your update before continue testing.
best,
Mohammad

PR #692, opened 2017-12-01
Make members of FairMCApplication protected

sawenzel:
This allows to inherit from FairMCApplication and access
the data members directly (as not accessible through getters).


sawenzel:
@MohammadAlTurany : As discussed offline.

PR #308, opened 2016-03-10
examples: finer grained dependencies

sbinet:
This CL allows to build the MQ examples which don't depend on any simulation.

PR #696, opened 2017-12-06
Added FairModule::fMC data member and more:

ihrivnac:
- This avoid accessing VMC via thread-local variables and fixes the AliceO2 issue #308
- AliModule::kGeoSaved renamed in fGeoSaved to be consistent class data members
  naming in FairRoot


sawenzel:
Maybe AliModule should be replaced by FairModule in the commit message.

ihrivnac:
Indeed, the right log message should be FairModule ... Thanks :)

MohammadAlTurany:
Merged Manually after correcting the comment text

PR #699, opened 2017-12-07
FairMQ: Add ability to say how much of the allocated message was used

rbx:
- Add ability to set used message size: `msg.SetUsedSize(size_t size)` (unused space will be freed for reuse (for transports that support this)).
  *shmem (managed)* will shrink used buffer and free the unused space for reuse.
  *zeromq* will create a new message with a view on the subset of the original buffer.
  *nanomsg* will "chop off" the unused part, however it is only freed after message is cleanup.
- Add tests for the above.
- Disable irrelevant tests output to workaround output scrambling on MacOS 10.11.
- Remove `GetMessage`/`SetMessage` methods from the general FairMQMessage interface - these are purely transport details and should not be exposed in the interface.


fairmq/nanomsg/FairMQMessageNN.h
```diff
@@ -33,21 +33,23 @@ class FairMQMessageNN : public FairMQMessage
...
+    size_t GetSize() const override;
 
-    virtual void SetMessage(void* data, const size_t size);
+    bool SetUsedSize(const size_t size) override;
 
-    virtual FairMQ::Transport GetType() const;
+    void SetMessage(void* data, const size_t size) override;
 
-    virtual void Copy(const std::unique_ptr<FairMQMessage>& msg);
+    FairMQ::Transport GetType() const override;
```
dennisklein: lost return type

dennisklein: forget it, i cannot read :P

fairmq/zeromq/FairMQMessageZMQ.cxx
```diff
@@ -12,60 +12,71 @@
  * @author D. Klein, A. Rybalchenko, N. Winckler
  */
 
-#include <cstring>
-#include <cstdlib>
 
 #include "FairMQMessageZMQ.h"
 #include "FairMQLogger.h"
+#include "fairmq/Tools.h"
```
dennisklein: I would prefer `#include <fairmq/Tools.h>` here. `""` mean here, look first in `fairmq/zeromq/fairmq/Tools.h`.

Or use `#include "../Tools.h"`. This should not be done in header files though.

fairmq/zeromq/FairMQMessageZMQ.cxx
```diff
@@ -12,60 +12,71 @@
  * @author D. Klein, A. Rybalchenko, N. Winckler
  */
 
-#include <cstring>
-#include <cstdlib>
 
 #include "FairMQMessageZMQ.h"
```
dennisklein: Although this is correct usage of `""`, I would use `#include <zeromq/FairMQMessageZMQ.h>`, or later, once we refactored all classes to namespaces `#include <fairmq/zeromq/Message.h>`

fairmq/zeromq/FairMQMessageZMQ.h
```diff
@@ -31,26 +32,33 @@ class FairMQMessageZMQ : public FairMQMessage
...
+    void ApplyUsedSize();
 
-    virtual FairMQ::Transport GetType() const;
+    void SetMessage(void* data, const size_t size) override;
 
-    virtual void Copy(const std::unique_ptr<FairMQMessage>& msg);
+
+    FairMQ::Transport GetType() const override;
+
+    void Copy(const std::unique_ptr<FairMQMessage>& msg) override;
```
dennisklein: At some point we should fix these `const unique_ptr &` signatures. `const FairMQMessage&` will do just fine, or? https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/

fairmq/zeromq/FairMQMessageZMQ.h
```diff
@@ -31,26 +32,33 @@ class FairMQMessageZMQ : public FairMQMessage
...
 
-    virtual ~FairMQMessageZMQ();
+    ~FairMQMessageZMQ() override;
 
   private:
-    zmq_msg_t fMessage;
+    bool fUsedSizeModified;
+    size_t fUsedSize;
+    std::unique_ptr<zmq_msg_t> fMsg;
+    std::unique_ptr<zmq_msg_t> fViewMsg; // view on a subset of fMsg (treating it as user buffer)
```
dennisklein: Do we need to store `fViewMsg`? Could we just create it on the fly when sending?

rbx: Sending could still fail/timeout and we should keep the view around for retry/cleanup, even though the size cannot be changed anymore after first send.

fairmq/zeromq/FairMQMessageZMQ.cxx
rbx: I would rather do it as a separate PR, together with the namespace changes. For now I apply it only to new includes.

fairmq/zeromq/FairMQMessageZMQ.cxx
rbx: fixed.

fairmq/zeromq/FairMQMessageZMQ.h
rbx: agree, we should move away from that passing. Since this is breaking I will keep the old method and mark it as deprecated.

rbx: Done. I updated only the Copy method. For Send/Receive I want to try what we discussed with passing with ownership transfer, but that would be a different PR.

fairmq/FairMQMessage.h
```diff
@@ -29,15 +29,15 @@ class FairMQMessage
...
+    virtual void* GetData() const = 0;
+    virtual size_t GetSize() const = 0;
 
-    virtual void SetMessage(void* data, size_t size) = 0;
+    virtual bool SetUsedSize(const size_t size) = 0;
 
     virtual FairMQ::Transport GetType() const = 0;
 
-    virtual void Copy(const std::unique_ptr<FairMQMessage>& msg) = 0;
+    virtual void Copy(const std::unique_ptr<FairMQMessage>& msg) __attribute__((deprecated("Use 'Copy(const FairMQMessage& msg)'"))) = 0;
```
matthiasrichter: I think its difficult to force such a hard API change. Can the old API be supported for a while? There is anyhow no ownership transferred here, so the unique pointer can be dereferenced and the new API be called.

rbx: Why is it problematic?
It is not really forced, since it is just marked as deprecated and the old version works (at least for a while). The warning should help you switch to new API.
Cannot imagine an easier way to change API.

matthiasrichter: In the present case, the API change breaks the O2 CI, and there is no way to fix this consistently, because the CI does checks with both the dev and the master, but only one version of the API is allowed.

rbx: We will keep the old version at least until the end of next release. You can switch to the new API once the master is updated, there is no rush to change this now in your code.

Or how else would you do it?

matthiasrichter: true, then everything is correct, actually the CI is broken because we treat the warnings as errors. Thanks for the clarification.

Somehow I was expecting the wrapper for the old API to be in the header file itself, which made me confused.

PR #704, opened 2017-12-22
Single Logger implementation for FairLogger & FairMQLogger

rbx:
This introduces `Logger` class that is used by:
- `FairMQLogger` (uses `Logger` directly, header just for backwards-compatibility)
- `FairLogger` (minimal wrapper class to maintain some of the special `FairLogger` features and backwards-compatibility).

TODOs:

* [x] Provide all severity and verbosity levels used by both old loggers (with some output enchantments).
* [x] Change all severity names to lowercase. Uppercase is still available for backwards-compatibility.
* [x] Remove usage of `MQLOG` macro, previously used to avoid conflict of both old loggers.
* [x] Make `info` default severity (default in `FairMQLogger` was `DEBUG`) and `low` the default verbosity (default in `FairMQLogger` was `medium`).
* [x] Provide a hook after `FATAL` message, FairLogger uses it to throw an exception and call `gSystem->StackTrace()`.
* [x] Refactor LOG calls that relied on `FairLogger` not adding `endl` automatically.
* [x] Remove `FairLogger::endl` usage from code, which now does basically nothing.
* [x] Removed `TMCThreadLocal` from FairLogger instance. `Logger` is already thread safe.
* [x] Fix FairMQ tests: fix captured stream.
* [x] Fix failing `FairLogger` tests.
* [x] Do more tests with root macros to make sure nothing broke.
* [x] Mark the printf-style interface of FairLogger as deprecated.

The new Logger is backwards-compatible, at least from what I could test so far.


rbx:
While testing the new implementation I noticed some performance degradation with Boost compared to FairLogger. The runtime of the entire FairRoot test suite almost doubled with the new implementation.

First thing I tried to improve it is to disable all customizations on top of boost to make sure it is nothing that I added on top - the performance remained unchanged.

The discussion [here](https://sourceforge.net/p/boost-log/discussion/710021/thread/47dd83b3/) suggests that the timestamps that boost calculates are not very efficient and also logging of entries that should be hidden (e.g. TRACE&DEBUG calls during INFO severity) takes considerable amount of time (all attributes are calculated even if a log line will not be sinked).

I setup a small test that logs all severities & verbosities followed by 32 million log lines in a "hidden" severity (so they should be skipped as fast as possible).

With default boost log setup test runs in **12 seconds**.

Making a custom calculation for the timestamps in the formatters using `std::chrono` reduced this time to **8 seconds**.

Further I tried to make a custom check in the macro that checks severity before even giving a line to boost - this reduced the time to **1.6 seconds**.

At this point already considerable amount of customizations have been added where not much is left for boost to do (at least from the features that we use), but the runtime of the FairRoot test suite is still almost double.

I tried a custom logger implementation without boost ([branch](https://github.com/rbx/FairRoot/tree/logger-experimental)) just with `cout` and `fstream`. With it the small test completes in just **0.1 seconds**. And the test suite runs as fast as before. It now has all of the features we require.

Here are some additional test results:

|                      | old (two loggers) | new boost       | new custom |
|:---------------------|:-----------------:|:---------------:|:----------:|
| compile time         | 7.1 min           | 9.2 min         | 6.9 min    |
| test suite runtime   | 1.7 min           | **2.9** min     | 1.7 min    |
| small test runtime | -                 | **12 (8, 1.6)** sec | **0.1** sec    |
| bin dir size         | 292.4 MB          | 293.8 MB        | 279.9 MB   |
| lib dir size         | 146.3 MB          | 173.2 MB        | 132.9 MB   |

My guess as to why the runtime of the test suite is so much slower with boost is that cling has to parse all the boost log headers on each execution. But I have not verified this.

Considering these test results (and also the fact that at least one more customization on top of boost would be necessary to provide the required special handling of the FATAL severity) I think we should ditch Boost.Log and use a custom implementation. It has all the features that we use, unless we foresee need of some of the more unique boost log features (log file rotation, ...?).

rbx:
As discussed I merged my experimental logger branch here.
From what I can tell this is now feature complete.
I updated the TODO list above to reflect all the changes.

ktf:
I am trying this in ALICE O2 but this is not backward compatible, apparently. I get:

```
sw/BUILD/O2-latest/log
Scanning dependencies of target SubframeBuilderDevice
[ 32%] Building CXX object Utilities/DataFlow/CMakeFiles/SubframeBuilderDevice.dir/src/runSubframeBuilderDevice.cxx.o
/mnt/mesos/sandbox/sandbox/sw/SOURCES/O2/0_O2_DEV_FAIRROOT/0/Framework/Core/src/runDataProcessing.cxx: In function 'int doChild(int, char**, const o2::framework::DeviceSpec&)':
/mnt/mesos/sandbox/sandbox/sw/SOURCES/O2/0_O2_DEV_FAIRROOT/0/Framework/Core/src/runDataProcessing.cxx:250:13: error: 'fair::mq::logger' has not been declared
   fair::mq::logger::ReinitLogger(false);
             ^~~~~~
make[2]: *** [Framework/Core/CMakeFiles/Framework.dir/src/runDataProcessing.cxx.o] Error 1
```

It would be good if changes like this are tested against O2 and announced in the Framework meeting.

Any idea on how to solve the above issue?

rbx:
@ktf sorry, you're right, I didn't test with O2. The backwards-compatibility is for the LOG macro and FairLogger interfaces - I wasn't aware anyone is using the FairMQ logger methods except in our internal code. Will check more carefully next time.

Anyway the replacement for the above call is `fair::Logger::SetConsoleColor(false);`.

ktf:
Great, thanks. I am in the process of updating my PR.

PR #716, opened 2018-01-24
Remove boost::regex dependency from tests, use std

rbx:


fuhlig1:
Cool. I missed that regex made it to the standard.

rbx:
@fuhlig1 but only since c++11, and the old code was written before we supported c++11 :)

PR #718, opened 2018-01-26
Logger: Allow adding custom sinks

rbx:
Example use in logger/loggerTest.cxx:141

Plus some refactoring to simplify the code.


dennisklein:
Could you add a paragraph to the FairMQ markdown docs explaining how to use this customization point? That would be awesome :) I also should sit down and catch up on missing docs ;)

rbx:
Good idea, will do :)

rbx:
Added some docs.

PR #722, opened 2018-02-01
Seems -inlineInputHeader does not work with #pragma once headers

klenze:
please consider removing it.


fuhlig1:
The problem here is that the flag changes the generation of the dictionaries. To be able to remove the flag we have to test the dictionary generation on all supported platforms. I am not sure what is the effect if the flag is removed.

klenze:
The inlineInputHeader flag was added in 0289d1e6aa948b7c5109d21b0a6c55b29cd6796a, the general ROOT6 support commit. There is no indication in git that the flag was added to fix a bug. 

https://sft.its.cern.ch/jira/browse/ROOT-6468 mentions that this flag might be required when one wants to generate relocatable binaries while feeding rootcint with absolute path header files. 

However, with our cmake setup, no absolute path headers are passed to rootcint:
Running 
find . -iname generate\*.sh | xargs grep -e ' /[^ ]*\.h'
in build returns no lines. 
(Beside, include pathes used during compilation are stored as absolute pathes in the generated G__*Dict.cxx file (string array includePathes) regardless of the settingRootcint. Go figure.)

Crashing when encountering the de-facto standard #pragma once -- which was working with the ROOT5 version -- seems like a bug to me. Removing the flag is a workaround. CI seems to indicate that removing the flag does not break the build. Thus, I would kindly ask you to merge the commit. If this should break the build for HP-UX or something you can always roll back. :-)

fuhlig1:
Hi Dennis, Hi klenze,

thanks for investigating this issue further. If it not used an needed but breaks in existence of #pragma once we should remove the flag. I checked back why this flag was added but probably never needed. The idea when ROOT6 started was to put everything you need to compile a macro into the dictionary, including the required headers. To my knowledge this project of modules is not finished yet such that the flag is not needed.

So I think we can safely merge the pull request.

Ciao

Florian

fuhlig1:
Merged with 8f7269d8750e1197512d208090e8361a59a289ac. Rebase was needed.

PR #740, opened 2018-03-19
dds plugin: handling for same channel names from multiplied collections

rbx:


fairmq/PluginServices.h
```diff
@@ -166,6 +166,8 @@ class PluginServices
 
     // Config API
 
+    auto CountProperty(const std::string& key) -> int { return fConfig->Count(key); }
```
dennisklein: I would not expose the `variables_map::count` interface here. What you need is just a test, if the key is present, maybe something like: 

```cpp
auto PropertyExists(const std::string& key) -> bool { return fConfig->Count(key) > 0; }
```

`count` is specific to how often the key appeared on the command line.

rbx: Fixed as suggested.

PR #748, opened 2018-03-27
FairMQ: Fix regression bug introduced with PR#728

dennisklein:
@ihrivnac Hi, I believe this should fix the issue you reported. It was an oversight in my last PR that got merged couple of days ago.


dennisklein:
@MohammadAlTurany The test failure is the usual one, that sometimes fails, it is unrelated to the change here.

PR #755, opened 2018-04-18
Move logger and fairmq to external packages

### LINKED ISSUE #736: Make a special repository for FairMQ
# Split FairMQ from FairRoot
## Introduction
Students at the Amsterdam University of Applied sciences will in the future learn how to work with clusters and load balancers. For this they learn how to use Ansible, Zookeeper and we would like them to learn FairMQ as well. For this purpose we use the Raspberry Pi 3.
whatsapp image 2018-02-28 at 15 57 51

Due the limited hardware capabilities, it's not practical (and for now impossible) to run a full version of FairRoot on a Raspberry Pi, since we do not need the full version of FairRoot we suggest to place FairMQ in a separate repository with its own build and installation scripts.

## Why use a separate repository?
I already forked the code and created a separate [repository](https://github.com/valvy/PiFairMQ) containing FairMQ's code which was tested on the Raspberry Pi 3. It is faster to compile on the 'limited' hardware and doesn't completely overwhelm a small SD-card.

If FairMQ receives its own dedicated repository, students can always use the latest version of FairMQ while I don't have to backport all features from the FairRoot SDK. Also if a students find an issue they can push it to a central repository instead of both to my repository and to the FairRoot SDK.

dennisklein:
You can test this currently with FairRootGroup/FairSoft#69.

This is the minimum work to do to get it working. There is additional cleanup to do, but I suggest to do this in the context of the CMake modernization PR.

Resolves #736


dennisklein:
@karabowi Do you have any idea, what the problem with `ex9a_static` test here is? https://cdash.gsi.de/testDetails.php?test=4275361&build=176097

I have seen this error couple of times already in the past. Somehow the post condition fails from time to time.

PR #756, opened 2018-04-24
add compiler check to exclude wrong error for the same symlinks: g++ c++

kgertsenberger:
the incorrect error "different compilers" has been encountered Ubuntu 17.10, 17.04, SLC 6... where OS has symlink c++ pointed to g++


dennisklein:
CMake seems to offer a portable `diff` with `cmake -E compare_files file1 file2`. If it works, it might be preferrable even.

kgertsenberger:
Ok, i will try this portable solution.

вт, 24 апр. 2018 г., 16:41 Dennis Klein <notifications@github.com>:

dennisklein:
> Ok, i will try this portable solution.

@kgertsenberger Just don't spend too much time on it ;) I believe, we would also be happy to merge your proposed solution with `diff`.

kgertsenberger:
Unfortunately, i have no MacOS, so i tested it only on Linux. 
It is not problem for me to test different solutions especially if it is maximally portable.
Thank you for your suggestion.

MohammadAlTurany:
No problem, we have enough MACs to test it, just send the PR and it will be tested on MAC.

PR #757, opened 2018-04-25
additional compiler checking to exclude wrong error for the same symlinks: g++ c++

kgertsenberger:


kgertsenberger:
Good morning.
Does my pull request cause the crash of FairRoot?
I don't understand what does it mean:
51/62 Test #52: run_rutherford_TGeant4 ...........***Failed  Required regular expression not found.Regex=[Macro finished successfully

fuhlig1:
I don’t think that your pull request is the problem. Sometimes something in the initialisation phase went wrong.
Did you try to run the single test alone?

Ciao

Florian

Issue #750, opened 2018-04-09
Create base class for propagators


TobiasStockmanns:
FairRoot has a particle propagator called FairRKPropagator. It would be nice to have a base class for particle propagators in general to be able to select different propagators.

kresan:
implemented in #766

PR #766, opened 2018-05-08
Added base abstract class for track propagation algorithms.

kresan:
solves #750


kresan:
agree, is implemented in #767

Issue #771, opened 2018-05-15
FairRun::GetOutputFile removed breaks backward compatibility


TobiasStockmanns:
FairRun::GetOutputFile() was removed in the dev branch.
This method is used in PandaRoot which therefore crashes and cannot be compiled.

Is there a subsititute for this method?

dennisklein:
I believe this function was replaced by the `FairSink` facility. Try setting a `FairFileSink`.

karabowi:
Dear Tobias,

With the new interface, the FairSink, FairRoot may store the output data using other means than the output file (f.e. it may send via FairMQ). Therefore it is not guaranteed that there will even be output file to return.
Thus, the method GetOutputFile is now gone. If you want to store another data, like histograms,
you should create a separate file.

TobiasStockmanns:
We will change PandaRoot accordingly but it would be nice if the interface of a class changes that this change is done in two stages.
First giving a warning that this method will not be available in the future and, just in the second stage, remove it.

dennisklein:
I believe this issue is resolved by reintroducing the removed API as deprecated. Reopen, if it doesn't work or anything is missing.

PR #772, opened 2018-05-15
Ease interface change

dennisklein:
@karabowi According to #771 the removed `FairRun::GetOutputFile()` member function is used in PandaRoot. Should we put it back as deprecated like the setters?


dennisklein:
@karabowi Changed the message.

Issue #777, opened 2018-05-18
FairGeanePro should derive from new FairPropagator


TobiasStockmanns:
To use FairGeanePro also as a FairPropagator it should derive from the new base class.
In addition some of the propagation types like propagate to PCA should become part of the base class.

PR #781, opened 2018-05-28
changes to FairPropagator

kresan:
- Inherit FairGeanePro from FairPropagator.
- Additional functions in FairPropagator interface.


kresan:
fix for #777

PR #792, opened 2018-06-11
Fixed MT simulation broken with the introduction of FairSink

ihrivnac:


karabowi:
Dear Ivana,

Would it be possible to change the name UserOutputFileName?
It suggests to any user that he should expect output in such file.
In fact the function only affects the multithreaded worker,
so maybe something like G4WorkerFileName?
Also, wouldn't it be better to move the functionality from FairRun to FairRunSim,
as it only affects simulation?

yours,
radek

ihrivnac:
Dear Radek,
I understand your point, but changing the name as you proposed would be misleading, as the data member added in FairRun represents really the file name defined by the user when he sets the sink. Also, as the sink is set in FairRun, then the data member is needed in the base class and so it is natural to put setter and getter to this class.
The file name is the primary information which is set by the user when defining the sink, so it makes sense to keep it outside the contained objects; and for MT it is needed, as we need to keep this info set by the user on the master up to the moment when the worker sink can be created. So though these functions were added for MT and are used in MT, they should be available in the FairRun & FairSink framework. You may also imagine another use case, that a user first sets the file name to FairRun class, and use it later in his code to create and set the sink object.
Does the above make sense for you?
Cheers,
Ivana

karabowi:
Hi,
OK, let's just follow that.
yours,
radek

Issue #793, opened 2018-06-13
Problems with pthread in FairRoot (issue moved from GSI Forum)


MohammadAlTurany:
This issue was reported originally on [GSI Forum](https://forum.gsi.de/index.php?t=post&reply_to=22714&rid=0&SQ=fce68b0229469336830d8da30244d99d&).

> Hi everybody,
>
>after successfully configuring the current version of FairSoft on CentOS 7, I tried to install FairRoot. Running cmake resulted in the following error:
>
>CMake Error at cmake/modules/FindGo.cmake:49 (message):
>Go build for go-processor failed!
>Call Stack (most recent call first):
>examples/advanced/GoTutorial/CMakeLists.txt:6 (build_go_package)
>
>
>-- Configuring incomplete, errors occurred!
>I attached the corresponding CMakeError and CMakeOutput logs. The issue seems to be something related to pthread from what I can gather from the logs. pthread is installed and is also found by cmake:
>
>-- Looking for pthread.h
>-- Looking for pthread.h - found
>-- Looking for pthread_create
>-- Looking for pthread_create - not found
>-- Looking for pthread_create in pthreads
>-- Looking for pthread_create in pthreads - not found
>-- Looking for pthread_create in pthread
>-- Looking for pthread_create in pthread - found
>-- Found Threads: TRUE 
>
>Did anyone already encounter this issue? Or is running the current version of FairRoot on CentOS 7 and knows how I can get it up and running? 
>Thanks for your help!
>Cheers,
>Benjamin

[CMakeError.log](https://github.com/FairRootGroup/FairRoot/files/2097540/CMakeError.log)
[CMakeOutput.log](https://github.com/FairRootGroup/FairRoot/files/2097541/CMakeOutput.log)

dennisklein:
Looking at the logs, I agree with Mohammad's assessment that this is a problem with the Go installations. Our CMake just finds the go executable in the `PATH` environment variable and later calls `go get` in the two go tutorial directories. This is not related to the pthread detection also performed by CMake.

Can you try the following:

```
FairRoot$ cd examples/advanced/GoTutorial/src/go-processor
FairRoot/examples/advanced/GoTutorial/src/go-processor$ go get
```

sbinet:
FWIW, I tried the following on lxplus7:
```
$> which go
/usr/bin/go

$> go version
go version go1.9.4 linux/amd64

$> go env
GOARCH="amd64"
GOBIN=""
GOEXE=""
[...]

Issue #694, opened 2017-12-05
`examples/advanced/GoTutorial` leaves source directory dirty


dennisklein:
Go compilation seems to download files and leaves them in the source directory. Files that are not tracked via git need to be created in the out-of-source build directory.

dennisklein:
33b76b2bbf9fceb482e6f0e163e9fe1871244f54 removes the Go example, so this can be closed.

PR #794, opened 2018-06-14
cmake/modules: fix Go execute_process recipe

sbinet:


dennisklein:
@sbinet Could you, by any chance, also address #694?

sbinet:
assign it to me (so I don't forget about it)
but to fix it, I'll need to up my CMake game. (something I am rather reluctant to, as Go doesn't impose on me any build system :P)

dennisklein:
Thx. Need to see how to assign you, probably you need to become a member of this repo first.

Coming back to this solution, I still have questions: What I do not understand, is the reasoning for this change. People observed build problems on gcc 4.8 platforms. Did it work there before? If I understand you correctly, you concluded, that a new cmake version behaves differently, will this change also work on other gcc platforms? So, what is the root problem people are observing on which platforms?

sbinet:
my take on this is that the `CMakeOutput.log` and `CMakeError.log` were just red herrings.
and the `pthreadS` thing was just "normal" plumbing to detect whether on should link against `pthread` or `pthreads`.

`cmake` would fail because `go get` would either return `0` (and `FindGo` would interpret this as failure) or "not `0`" (because _e.g._ of the `GOPATH` issue) and `FindGo` would properly interpret this as failure.

AFAICT, `gcc-4.8` (or otherwise) is completely irrelevant to the issue at hand: Go binaries usually do not link against any C/C++.

dennisklein:
> my take on this is that the CMakeOutput.log and CMakeError.log were just red herrings.
and the pthreadS thing was just "normal" plumbing to detect whether on should link against pthread or pthreads.

I agree.

> cmake would fail because go get would either return 0 (and FindGo would interpret this as failure) or "not 0" (because e.g. of the GOPATH issue) and FindGo would properly interpret this as failure.

So it was broken before?

> AFAICT, gcc-4.8 (or otherwise) is completely irrelevant to the issue at hand: Go binaries usually do not link against any C/C++.

If you look in your `go env` output, there are quite a lot of gcc related variables. So far, I don't think we understand the relationship between go and gcc, do we?

dennisklein:
> (because e.g. of the GOPATH issue)

Which issue? On my machine (Fedora27) I can execute the listing I posted in #793 without any problems (and it is what cmake would do). My hope was we could establish on which platforms this does not work and then fix the cmake accordingly. Looking at the proposed change in this PR I am still puzzled about what has changed to before, or was it broken all along?

sbinet:
> If you look in your go env output, there are quite a lot of gcc related variables. So far, I don't think we understand the relationship between go and gcc, do we?

yes, this is used for when somebody is using `cgo` (the Go <-> C FFI).
`sbinet-alice/fer` is pure-Go (I have implemented (a subset of) ZeroMQ in pure-Go, and nanomsg had already a pure-Go package).
so there's no C involved.

> or was it broken all along?

that's quite possible (my CMake-fu is very weak (sorry about that) so I may have borked something) but I don't think so.
the `GOPATH` issue I was hinting at was when somebody had no `$GOPATH` defined.
Going to `examples/advanced/GoTutorial/src/go-processor` and then issueing `go get`, would result in an error.
Indeed, starting with Go-1.8, when no `$GOPATH` was defined, Go would default to `$HOME/go`.
but then, running `go get` from within `go-processor` would make Go try to install the resulting binary under `$HOME/go/bin` which is a build config smell for the `go` command.

dennisklein:
> yes, this is used for when somebody is using cgo (the Go <-> C FFI).
sbinet-alice/fer is pure-Go (I have implemented (a subset of) ZeroMQ in pure-Go, and nanomsg had already a pure-Go package).
so there's no C involved

And from a quick google search I just learned there is also `gccgo` which is a go frontend for gcc. But that is not, what is used, when issuing a `go get` AFAIK.

Still confused about the `GOPATH` behaviour, but judging purely from the cmake code, the proposed change should be fine (my first guess was it is equivalent, but you did the testing and apparently its not).

MohammadAlTurany:
Hi,
> @MohammadAlTurany Can you verify?
Yes. This change make it works on Ubuntu 14.04 with go 10.03 and gcc 4.8 

About the discussion above, I am sure it is related to the gcc, on the same machine if I use the gcc 6.2 all works fine.

PR #797, opened 2018-07-03
Do not prevent building with --std=c++17

dberzano:


dberzano:
Hi all, we're trying to bump our compiler to GCC v7.3 and C++ standard to C++17 but the current CMake configuration of FairRoot prevents us from doing that. Would it be possible to lift the ban by merging this PR? Thanks!

Reference: alisw/alidist#1214

dberzano:
@dennisklein is it really still building `continuous-integration/jenkins/pr-head`? Looks a bit stuck to me 🙂

PR #809, opened 2018-08-21
Option String & ListOftTasks

RalfKliemt:
Added option String to FairRun. Experiments should handle parsing by themselves.
Iterative retrieving of task names for task list.

### DIFF
--- a/base/steer/FairRun.cxx
+++ b/base/steer/FairRun.cxx
@@ -75,18 +75,37 @@ FairRun::~FairRun()
 //_____________________________________________________________________________
 
 //_____________________________________________________________________________
-void FairRun::AddTask(FairTask* t)
+void FairRun::AddTask(FairTask* task)
 {
-  fTask->Add(t);
-  fNTasks++;
-  fFileHeader->AddTaskClassName(t->ClassName());
+  fTask->Add(task);
+  StoreTaskNames(task);
 }
 //_____________________________________________________________________________
-void FairRun::SetTask(FairTask* t)
+void FairRun::StoreTaskNames(const FairTask* task)
+{
+  if(task) {
+    fFileHeader->AddTaskClassName(task->ClassName());
+    fNTasks++;
+
+    const TList* subtasks = task->GetListOfTasks();
+    if(subtasks) {
+      for(const auto subtask : *subtasks) {
+        StoreTaskNames(static_cast<FairTask*>(subtask));
+      }
+    }
+  }
+}
+//_____________________________________________________________________________
+void FairRun::SetTask(FairTask* task)
 {
   delete fTask;
-  fTask = t;
-  fFileHeader->AddTaskClassName(t->ClassName());
+  fTask = task;
+  fNTasks=0;
+  TList* tasklist = fFileHeader->GetListOfTasks();
+  if(tasklist) {
+    tasklist->Clear();
+    StoreTaskNames(task);
+  }
 }
 //_____________________________________________________________________________
 void FairRun::CreateGeometryFile(const char* geofile)
--- a/base/steer/FairRun.h
+++ b/base/steer/FairRun.h
@@ -1,8 +1,8 @@
 /********************************************************************************
  *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
  *                                                                              *
- *              This software is distributed under the terms of the             * 
- *              GNU Lesser General Public Licence (LGPL) version 3,             *  
+ *              This software is distributed under the terms of the             *
+ *              GNU Lesser General Public Licence (LGPL) version 3,             *
  *                  copied verbatim in the file "LICENSE"                       *
  ********************************************************************************/
 #ifndef FAIRRUN_H
@@ -49,6 +49,7 @@ class FairRun : public TNamed
      * Add a FAIRTask to the simulation or analysis
      */
     virtual void       AddTask(FairTask* t);
+    virtual void       StoreTaskNames(const FairTask* t);
     virtual void       SetTask(FairTask* t);
     /**
      * Initialize the Simulation or analysis
@@ -96,13 +97,13 @@ class FairRun : public TNamed
       return (static_cast<Int_t>(fRunId));
     }
 
-	/** 
+	/**
      * Set the Run ID
 	 */
     void SetRunId(UInt_t runId)
 	{
-	  fRunId = runId; 
-    }    
+	  fRunId = runId;
+    }
 
     /**Get the detector specific run header*/
     FairEventHeader*  GetEventHeader();
@@ -154,6 +155,12 @@ class FairRun : public TNamed
     //** Mark/Unmark event to be filled into output. Default is TRUE. */
     void MarkFill(Bool_t flag) { fMarkFill = flag; }
 
+    //** Get option string */
+    TString GetOptions(){return fOptions;};
+
+    //** Set option string */
+    void SetOptions(const TString& s){fOptions=s;};
+
     // vvvvvvvvvv depracted functions, replaced by FairSink vvvvvvvvvv
     /**
      * Set the output file name for analysis or simulation
@@ -198,6 +205,8 @@ class FairRun : public TNamed
     FairSink*                fSink;
     /**Output file name set by user*/
     TString                  fUserOutputFileName;
+    /**Options for derived classes, to be set & parsed by user*/
+    TString                  fOptions;
     /**Run Id*/
     UInt_t                   fRunId;//!
     /** true for Anaylsis session*/


base/steer/FairRun.cxx
```diff
@@ -79,14 +79,26 @@ void FairRun::AddTask(FairTask* t)
 {
   fTask->Add(t);
   fNTasks++;
+  StoreTaskNames(t);
+}
+//_____________________________________________________________________________
+void FairRun::StoreTaskNames(FairTask* t)
+{
+  if(t==NULL) return;
```
dennisklein: Prefer `if(t == nullptr)` or just `if(!t)`.

base/steer/FairRun.cxx
```diff
@@ -79,14 +79,26 @@ void FairRun::AddTask(FairTask* t)
 {
   fTask->Add(t);
   fNTasks++;
+  StoreTaskNames(t);
+}
+//_____________________________________________________________________________
+void FairRun::StoreTaskNames(FairTask* t)
```
dennisklein: `void FairRun::StoreTaskNames(const FairTask* t)`

base/steer/FairRun.cxx
```diff
@@ -79,14 +79,26 @@ void FairRun::AddTask(FairTask* t)
...
+{
+  if(t==NULL) return;
   fFileHeader->AddTaskClassName(t->ClassName());
+  if (t->GetListOfTasks()==NULL) return;
+  for(TObject* t2: *(t->GetListOfTasks()) )
+  {
+    StoreTaskNames((FairTask*)t2);
+  }
+  return;
```
dennisklein: * Let's not call `GetListOfTasks()` twice.
* Downcast here with `static_cast` - safer than C-cast.

```cpp
const auto tasks = t->GetListOfTasks();
if(t)
{
  for(const auto task : *tasks)
￼   {
￼     StoreTaskNames(static_cast<FairTask*>(task));
￼   }
}
```

base/steer/FairRun.h
```diff
@@ -49,6 +49,7 @@ class FairRun : public TNamed
      * Add a FAIRTask to the simulation or analysis
      */
     virtual void       AddTask(FairTask* t);
+    virtual void       StoreTaskNames(FairTask* t);
```
dennisklein: `virtual void       StoreTaskNames(const FairTask* t);`

base/steer/FairRun.h
```diff
@@ -154,6 +155,12 @@ class FairRun : public TNamed
     //** Mark/Unmark event to be filled into output. Default is TRUE. */
     void MarkFill(Bool_t flag) { fMarkFill = flag; }
 
+    //** Get option string */
+    TString GetOptoins(){return fOptions;};
```
dennisklein: No unit tests, no catching typos ;)

base/steer/FairRun.h
```diff
@@ -154,6 +155,12 @@ class FairRun : public TNamed
...
+
+    //** Set option string */
+    void SetOptions(TString s){fOptions=s;};
```
dennisklein: Yes, `TString` has short string optimization, but I believe, it is still better to pass `const TString&` instead of by value.

base/steer/FairRun.cxx
```diff
@@ -79,14 +79,26 @@ void FairRun::AddTask(FairTask* t)
...
 }
 //_____________________________________________________________________________
 void FairRun::SetTask(FairTask* t)
 {
   delete fTask;
   fTask = t;
-  fFileHeader->AddTaskClassName(t->ClassName());
+  StoreTaskNames(t);
```
dennisklein: Since `fTask` is deleted, shouldn't you also clear the previously stored task names?

RalfKliemt: This is the same logic as before. I concur, though to clear the list. However, everything that was inside before will be gone, too.

base/steer/FairRun.cxx
RalfKliemt: NULL is being used elsewhere in the same file. I'd leave it at that.

base/steer/FairRun.cxx
dennisklein: True, maybe @MohammadAlTurany or @fuhlig1 have a comment here, if the previous behaviour should be preserved or if clearing the class names is the better way to go.

base/steer/FairRun.cxx
```diff
@@ -79,14 +79,28 @@ void FairRun::AddTask(FairTask* t)
...
   fNTasks++;
+  StoreTaskNames(t);
+}
+//_____________________________________________________________________________
+void FairRun::StoreTaskNames(const FairTask* t)
+{
+  if(NULL==t) return;
   fFileHeader->AddTaskClassName(t->ClassName());
+  if (NULL==t->GetListOfTasks()) return;
+  for(TObject* t2: *(t->GetListOfTasks()) )
```
dennisklein: Why do you want to call `GetListOfTasks` twice here?

base/steer/FairRun.cxx
```diff
@@ -79,14 +79,29 @@ void FairRun::AddTask(FairTask* t)
...
+//_____________________________________________________________________________
+void FairRun::StoreTaskNames(const FairTask* t)
+{
+  if(NULL==t) return;
   fFileHeader->AddTaskClassName(t->ClassName());
+  TList* l=t->GetListOfTasks();
+  if (NULL==l) return;
+  for(TObject* t2: *l )
+  {
+    StoreTaskNames(static_cast<FairTask*>(t2));
```
MohammadAlTurany: Here you need also to add "fNTasks++" to count the subtasks as the counter in AddTask was only increased by one.

base/steer/FairRun.cxx
```diff
@@ -79,14 +79,29 @@ void FairRun::AddTask(FairTask* t)
...
+  return;
 }
 //_____________________________________________________________________________
 void FairRun::SetTask(FairTask* t)
 {
   delete fTask;
   fTask = t;
-  fFileHeader->AddTaskClassName(t->ClassName());
+  TList* l = t->GetListOfTasks();
+  if(NULL != l) l->Clear();
```
MohammadAlTurany: I think we have also to reset the counter here (fNTasks=0) when the list is cleared.

dennisklein:
This is your current proposal:

```cpp
void FairRun::StoreTaskNames(const FairTask* t)
{
  if(NULL==t) return;
  fFileHeader->AddTaskClassName(t->ClassName());
  fNTasks++;
  TList* l=t->GetListOfTasks();
  if (NULL==l) return;
  for(TObject* t2: *l )
  {
    StoreTaskNames(static_cast<FairTask*>(t2));
  }
  return;
}
```

Why no spaces? Sometimes space after `if`, sometimes not...
Why no speaking variable names? FairRoot has no strict coding standard, true, but you made the consistency argument about `NULL` macro usage earlier yourself. Then pls apply the consistency argument consistently...

Compare your code to this:

```cpp
void FairRun::StoreTaskNames(const FairTask* task)
{
  if(task) {
    fFileHeader->AddTaskClassName(task->ClassName());
    fNTasks++;

    const auto subtasks = task->GetListOfTasks();
    if(subtasks) {
      for(const auto subtask : *subtasks) {
        StoreTaskNames(static_cast<FairTask*>(subtask));
      }
    }
  }
}
```

I am sorry for being so pedantic, but if we never start producing readable and maintainable code... We have modern C++ now for almost a decade, and choosing shitty single character variable names with numbers was bad practice even in the 70s. Your commit message has 3 typos in a single line!! Why do you treat programming like garbage? I will stop reviewing this PR now, it is not good for my heart.

dennisklein:
p.s.: All the questions in my last post are rethorical.

MohammadAlTurany:
Ok, come down @dennisklein, What you wrote about C++ is valid for this file and many others in FairRoot, many of these files are more than 10 years old.  So you cannot put the blame on Ralf. We should rewrite these files before we ask the people making PR to do it in proper C++.

dennisklein:
Noone will ever rewrite anything. The only realistic chance to improve existing codebases like this, is to increase quality step by step whenever one touches a piece. 

The function I am talking about is new code written in 2018. But well, its not my code, if you want to merge this, have fun, I hope I never have to touch code like this in the future... oh wait...

MohammadAlTurany:
Ok, this is a good starting point. We have to define coding rules and then we can even force them via test on the PR and reject whatever violate the rules. In AliceO2 it is already the case.

RalfKliemt:
Ahoi, Looks like I started something here. 
I concur that having coding rules is a good thing. 

On the subject at hand, I see your point in readability. I'm no particular fan of the "auto" types and the C++11 style loops. 

I'll revisit the stuff on Friday and I'll squash the history, too. 
Keep the tips and hits coming, even the pedantic ones.

Cheers! 
  Ralf

dennisklein:
> I'm no particular fan of the "auto" types

I hope, the example shows first of all how formatting and better variable names can make things much more readable. And in the "auto lines", the `const` is the important one. Make everything `const` by default unless you find reasons why things should be mutable (C++ standard screwed the default here, we should rather have a `nonconst`/`var`/`mutable` keyword instead). In the end I was mainly triggered by this:

```
if(NULL==t) return;
 ^ no space
if (NULL==l) return;
  ^ space
for(TObject* t2: *l )
    ^ no space     ^ space
```

or

```
TList* l=t->GetListOfTasks();
       ^^^
```

or

What is the logic of naming the argument of the function `t` (as in Task) and the running variable of the for loop `t2`? It is not just a second Task, is it - they are many children/sub tasks? I got the feeling you didn't even care that anyone else will have to read this code at some point, together with all the typos.. I am glad, you care !

But you want to see the world burn a little bit, don't you ;)

> and the C++11 style loops.

I did not propose those ;)

MohammadAlTurany:
Now it looks much better!

PR #824, opened 2018-09-24
Fix ambiguous method call in BaseMQFileSink

dennisklein:


dennisklein:
CI does not work any more for this old release.

PR #830, opened 2018-10-15
Alignment module

rklasen:
This adds a simple interface for apply misalignment to a detector geometry. Please see the readme.md for examples and details.


base/steer/FairRun.h
```diff
@@ -222,6 +228,10 @@ class FairRun : public TNamed
 
     Bool_t                   fMarkFill; //!
 
+    FairAlignmentHandler fAlignmentHandler;
+
+    void AlignGeometry() const;
+
     ClassDef(FairRun ,4)
```
dennisklein: Shouldn't we increase the schema version, too?

base/sim/FairMCApplication.cxx
```diff
@@ -1007,9 +1007,8 @@ void FairMCApplication::ConstructGeometry()
     delete particleIter;
   }
   fModIter->Reset();
-  while((Mod = dynamic_cast<FairModule*>(fModIter->Next()))) {
-    Mod->ModifyGeometry();
-  }
+
+  fRun->AlignGeometry();
```
dennisklein: 1. Have you considered calling `fRun->AlignGeometry()` in the [TVirtualMCApplication::MisalignGeometry() hook](https://root.cern.ch/doc/master/classTVirtualMCApplication.html#a5f0c903039eb1e4407e7a9a22dbddcad)?
2. Not calling our `ModifyGeometry()` hook any more will break some examples, won't it? If the new alignment module proposed here is intended to replace the `FairModule::ModifyGeometry()` hook, then there is more work needed, I am afraid (deprecate the interface and update the examples). If you do not want to update the examples, then we should still call it to not break existing code (probably it should also be moved to the `TVirtualMCApplication::MisalignGeometry()` hook!?)

base/steer/FairRun.h
```diff
@@ -222,6 +228,10 @@ class FairRun : public TNamed
 
     Bool_t                   fMarkFill; //!
 
+    FairAlignmentHandler fAlignmentHandler;
+
+    void AlignGeometry() const;
```
dennisklein: Let's just make this new member function `public`, so we do not need the explicit tight coupling of `FairRun` and `FairMCApplication` introduced in line `base/steer/FairRun.h#34`.

base/steer/FairRun.h
MohammadAlTurany: For consistancy YES, technically we never save the FairRun so it does not heart but it it shows the reader that something change.

rbx:
You can modernize the `LOG` calls if you are on release FairSoft may18 and FairRoot v18 or later. With earlier releases you have to follow the old API.

base/sim/FairMCApplication.cxx
spflueger: 1. No did not know that existed. It looks like that would be the right place to call it! Why have you not used that before and called `Mod->ModifyGeometry();` from within `ConstructGeometry()`?
2. I think it should replace the `FairModule::ModifyGeometry()` hook, since I do not see the need of ModifyGeometry() anymore. Is that ok for you? Then we would deprecate it and update the examples (and yes move the `fRun->AlignGeometry();` call to the `TVirtualMCApplication::MisalignGeometry()` hook).

base/steer/FairRun.h
spflueger: Hmm, I wanted to avoid users from calling `AlignGeometry()` of FairRun. What is so bad about making FairMCApplication a friend class?

base/steer/FairRun.h
spflueger: Ah yes we will also update that

spflueger:
> You can modernize the `LOG` calls if you are on release FairSoft may18 and FairRoot v18 or later. With earlier releases you have to follow the old API.

Ok, we will have a look. I thought we already did that.

base/sim/FairMCApplication.cxx
spflueger: I just checked how you handle the deprecation. As far as I saw you just use a `LOG(WARNING)` message at runtime, is that correct?

rbx: for deprecation warning, use `__attribute__((deprecated("message")`, which will warn at compile time if the marked method is used.
Example:
`void ModifyGeometry() __attribute__((deprecated("Use AlignGeometry instead")))`;

spflueger: Ok, that looks better. Btw, does FairRoot require any c++ standard, I believe c++11 or?

base/steer/FairRun.h
dennisklein: I just don't see the tight relationship of `FairRun` and `FairMCApplication`. Maybe we could move the friend declaration to `FairRunSim`?

base/sim/FairMCApplication.cxx
dennisklein: > Ok, that looks better. Btw, does FairRoot require any c++ standard, I believe c++11 or?

Correct.

fuhlig1: >     1. No did not know that existed. It looks like that would be the right place to call it! Why have you not used that before and called `Mod->ModifyGeometry();` from within `ConstructGeometry()`?

Probably because I overlooked the existence of this hook. 
 
>     2. I think it should replace the `FairModule::ModifyGeometry()` hook, since I do not see the need of ModifyGeometry() anymore. Is that ok for you? Then we would deprecate it and update the examples (and yes move the `fRun->AlignGeometry();` call to the `TVirtualMCApplication::MisalignGeometry()` hook).

Here I completely agree. To my knowledge the ModifyGeometry is only used in the FairRoot examples.

rklasen: My current workaround for this is indeed to run the macro twice.

1. Call the Init() function, get misalignment parameters and store them to disk (ascii file etc.) but don't run the actual simulation
2. Read the parameters from disk before calling Init(), pass them to the FairAlignHandler and then call Init() to apply the misalignment.

I updated the Tutorial4 to reflect that and removed all the code that is no longer neccessary.

base/steer/FairRun.h
rklasen: I would prefer to leave the friend declaration in the FairRun. For most of the simulation macros we use FairRunSim, that's true, but we also need the alignment functionality in the Digi macros, which use a FairRunAna, not a FairRunSim. Since both of these inherit from FairRun, is seems best to include the functionality in FairRun.

The reason we need it in the Digi macros is that the digitizer must know the sensors are in the wrong place, otherwise it will just take the mc points that come from the misaligned geometry and calculate digi hits that are outside the pixel detector (and will subsequently have out of bounds row and col values).

rklasen:
I updated the Tutorial4 and corresponding tut4Detector files. at the moment, this is the only practical way to generate misalignment matrices and apply them: https://github.com/FairRootGroup/FairRoot/pull/830#discussion_r226270565

other than that, the FairAlignmentHandler should work now.

base/steer/FairRun.h
dennisklein: In this case, I would recommend to just remove it completely and declare the member function public.

But in the end, I am not the decider in this matter ;)

spflueger: I don't like the friend class that much either, as it suggests the code as a whole maybe bad design. But we could not come up with something better, unless you redesign large portions of FairRoot or ROOT.
Making the function public is just moving the problem from one part to another. I mean we can put a wrapper around the AlignGeometry() to reject calls that would lead to problems (which the user could make). But thats not super pretty either.
We can just explain our idea or way of thinking. You guys can discuss internally and let us know what you would like most. Did you read my general comment about the how to read the alignment parameters?

spflueger:
I wanted to inform you that I used this PR together with Pandaroot and the geant simulations fail. Im not sure if this is something we introduced with this PR or something changed with the new Fairroot that my Pandaroot instance is not calling Fairroot correctly. Does the current Fairroot/dev work correcly?

fuhlig1:
Hi,

I have checked the current FairRoot dev branch with the PR applied and for me everything compiles and all the macros run without any problems. I did not yet check if the results are as expected.

rklasen:
Which macros did you run? We only tested our Lumi macros. Also, can you make a `git diff` between your currently working branch and this branch, or are they identical? Thanks!

fuhlig1:
> Am 13.02.2019 um 15:28 schrieb R Klasen <notifications@github.com>:
> 
> Which macros did you run?

I run the macros from 

examples/simulation/Tutorial4/macros

which are in FairRoot.

> We only tested our Lumi macros. Also, can you make a git diff between your currently working branch and this branch, or are they identical? Thanks!

I used the current FairRoot dev version from GitHub and applied the changes from the pull request.

spflueger:
So you run them from the install directory? For me the macros cannot resolve the FairRoot classes. Do I have to export more env variables? Or how do I run them?

Btw, the current dev branch does not compile on my system either. Is that true/normal?

MohammadAlTurany:
@spflueger on which system you run? and which version of FairSoft you use?

spflueger:
Hi Mohammad, it is on the cluster in Mainz (himster2) with faisoft may18.
However I tested it recently on my local machine and it ran fine. So
something seems to be wrong with the installation on the himster2.

Why are some of the tests for this pull requests failing? Is that normal? I
cant see what exactly is failing or why. Can the pull request be merged?

MohammadAlTurany:
Hi Stefan,
The errors are not related to your PR. and yes I will merge it now!
best,
Mohammad

PR #831, opened 2018-10-22
Example of histogram server, implemented in a device

kresan:
A device sending multiple histograms to the generic server. Server publishes the histograms using THttpServer. Implementation with the generic CanvasDrawer.

---

Checklist:

### DIFF
--- a/examples/MQ/histogramServer/FairMQExHistoDevice.cxx
+++ b/examples/MQ/histogramServer/FairMQExHistoDevice.cxx
@@ -1,0 +1,76 @@
+/********************************************************************************
+ *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ *                                                                              *
+ *              This software is distributed under the terms of the             *
+ *              GNU Lesser General Public Licence (LGPL) version 3,             *
+ *                  copied verbatim in the file "LICENSE"                       *
+ ********************************************************************************/
+#include <chrono>
+#include <thread>
+
+#include "FairMQExHistoDevice.h"
+#include "RootSerializer.h"
+
+#include "TH1F.h"
+#include "TH2F.h"
+#include "TRandom3.h"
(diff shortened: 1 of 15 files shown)


dennisklein:
@kresan With which version of FairMQ have you developed this example? We should determine the minimum version needed and then require this one in FairRoot.

examples/MQ/histogramServer/FairMQExHistoDevice.cxx
```diff
@@ -0,0 +1,76 @@
...
+    , fRandom(new TRandom3(0))
+    , fArrayHisto(new TObjArray())
+{
+}
+
+FairMQExHistoDevice::~FairMQExHistoDevice() {}
+
+void FairMQExHistoDevice::InitTask()
+{
+    fh_histo1 = new TH1F("histo1", "", 200, -0.5, 1.5);
```
rbx: Can all fh_histos be value members instead of pointers?
TH1F seems to have `operator=` defined, so instead of putting them on the heap here, we can do:
```C++
fh_histo1 = TH1F("histo1", "", 200, -0.5, 1.5);
```

kresan:
@rbx Drawing histograms on a canvas is absolutely optional. The http server can display just single histograms, as a list. I can imagine such simple use case

I will replace the pointers with objects, no problem. I think nothing speak against having objects also for histograms.

dennisklein:
> @dennisklein I have used FairMQ version 1.2.3 from FairSoft may18

Hm, but this means, that the PR here really does not compile. Can you verify, that you did not forget to commit any changes?

dennisklein:
If you have a look e.g. at these errors: https://cdash.gsi.de/viewBuildError.php?buildid=190199. `RootDeserializer` does indeed not exist in your branch.

kresan:
I have changed to RootSerializer, now it compiles

@rbx requested changes are in, please have a look

PR #834, opened 2018-10-30
MCStepLogger

karabowi:
Implementation of MCStepLogger to monitor stepping calls in VMC.
Originally developed by Sandro Wenzel for AliceO2.

---

Checklist:

* [ ] Rebased against `dev` branch

### DIFF
--- a/fairtools/CMakeLists.txt
+++ b/fairtools/CMakeLists.txt
@@ -9,6 +9,8 @@
 # the array .
 # The extension is already found.  Any number of sources could be listed here.
 
+add_subdirectory (MCStepLogger)
+
 Set(INCLUDE_DIRECTORIES
  ${CMAKE_SOURCE_DIR}/fairtools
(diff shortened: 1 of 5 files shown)


sawenzel:
Alternatively, I am wondering if we could make this code an external (somewhat configurable) package so that FairRoot VMC users could use it?

MohammadAlTurany:
@sawenzel Do you mean VMC users? As it is now all FairRoot users can use it!

sawenzel:
Well. To be precise, in the current version all users of a FairMCApplication can use it but that can be a configuration (in fact the name of the VMCApplication class has to be given as the only parameter to the source code). I just wanted to know if this code could be shared rather than copied ... but we can do so also later in case you want to go ahead.

MohammadAlTurany:
I think we can use it now as it is (all experiment using FairRoot can use it as is) and we can extend it later if needed.

PR #838, opened 2018-10-31
Add example of setting the random seed.

kresan:
An example of how to set a random seed for both Geant3 and Geant4.
Implemented in examples/simulation/Tutorial1

---

Checklist:

### DIFF
--- a/examples/simulation/Tutorial1/macros/run_tutorial1.C
+++ b/examples/simulation/Tutorial1/macros/run_tutorial1.C
@@ -98,6 +98,10 @@ void run_tutorial1(Int_t nEvents = 10, TString mcEngine = "TGeant3", Bool_t isMT
   // ------------------------------------------------------------------------
 
   // -----   Initialize simulation run   ------------------------------------
+  UInt_t randomSeed = 123456;
+  TRandom3 random(randomSeed);
+  gRandom = &random;
+
   run->Init();
   // ------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


examples/simulation/Tutorial1/macros/run_tutorial1.C
```diff
@@ -98,7 +98,13 @@ void run_tutorial1(Int_t nEvents = 10, TString mcEngine = "TGeant3", Bool_t isMT
   // ------------------------------------------------------------------------
 
   // -----   Initialize simulation run   ------------------------------------
+  UInt_t randomSeed = 123456;
+  TRandom3 random(randomSeed);
+  gRandom = &random;
+
   run->Init();
+
+  TVirtualMC::GetMC()->SetRandom(&random);
```
fuhlig1: Isn't it enough to use

gRandom->SetSeed(123456) ?

kresan: The method TVirtualMC::SetRandom stores the pointer in both gRandom and fRandom variables. If the second one is also used by the transport engine, then it is not enough.

fuhlig1:
That is true but in the constructor of TVirtualMC you have

> fRandom = gRandom;

which sets fRandom to the already existing gRandom. So doing all the changes to gRandom before calling Init() should be enough to properly set everything.

kresan:
@fuhlig1 it is changed now to modifying the gRandom before Init()

PR #841, opened 2018-11-05
Add example of simulation executable with options.

kresan:
Compile simulation macro of Tutorial1 into executable with command-line options.
---

Checklist:


rbx:
Not sure about root i/o, but having values in members is simpler & faster than navigating the variables_map on each access (hashtable lookup, boost::any conversion and potential copy of string vectors). Also it is tricky to modify the variables_map after `store()` has been called, although i guess here it doesn't matter.

PR #846, opened 2018-11-12
rename libMCStepLogger.so to libFairMCStepLogger.so

sawenzel:
This is necessary in order to avoid library name duplication
(with AliceO2 where this code was picked from) and to avoid problems
loading libraries within ROOT.


sawenzel:
This is an urgent fix. Please merge asap.

PR #850, opened 2018-11-21
Fix and update the project template.

kresan:
- Add test of the project template compilation. This requires default build command in Dart.sh "make install".

---

Checklist:


kresan:
@fuhlig1 I reverted back to "make" as build command in FairBase_test.cmake
I have changed to the following scheme:
- templates will be added as subfolder only if CMAKE_INSTALL_PREFIX was set by user
- the test itself installs FairRoot and then builds the project template

dennisklein:
Hi, a rebase against current `dev` should restore alfa-ci checks (The `dev_mt` check on Linux is failing at the moment).

PR #879, opened 2019-03-09
Add badges and improve make output

MohammadAlTurany:
- Add license badge 
- Use the FairMQ style implemented by Dennis in FairRoot


cmake/modules/FairMacros.cmake
```diff
@@ -320,19 +326,62 @@ ENDIF(FAIRROOT_FOUND)
...
   Set(FAIRLIBDIR ${CMAKE_BINARY_DIR}/lib)
-  
 ENDIF(FAIRROOT_FOUND)
-
 Set(LD_LIBRARY_PATH  ${FAIRLIBDIR} ${LD_LIBRARY_PATH})
-
-
 EndMacro (SetBasicVariables)
+################################################################################
+macro(find_package2 qualifier pkgname)
```
dennisklein: Use this improved version of the `find_package2` macro: https://github.com/FairRootGroup/FairMQ/blob/master/cmake/FairMQLib.cmake#L296-L341

I allows to write this (in top level `CMakeLists.txt`)

```cmake
set(FairRoot_Boost_COMPONENTS thread system timer program_options random filesystem chrono exception regex serialization log log_setup atomic date_time signals)
list(APPEND FairRoot_Boost_COMPONENTS ${FairMQ_Boost_COMPONENTS})
list(REMOVE_DUPLICATES FairRoot_Boost_COMPONENTS)
find_package(Boost 1.67 COMPONENTS ${FairRoot_Boost_COMPONENTS})
find_package2(PUBLIC Boost VERSION 1.67 COMPONENTS ${FairRoot_Boost_COMPONENTS})
```

as just

```cmake
find_package2(PUBLIC Boost
  VERSION 1.67 ${FairMQ_Boost_VERSION}
  COMPONENTS thread system timer program_options random filesystem chrono exception regex serialization log log_setup atomic date_time signals ${FairMQ_Boost_COMPONENTS}
)
```

It will pick highest listed version and automatically make the components list unique. I was also thinking about a syntax like this:

```cmake
find_package2(PUBLIC Boost
  VERSION 1.67
  COMPONENTS thread system timer program_options random filesystem chrono exception regex serialization log log_setup atomic date_time signals
  ADD_REQUIREMENTS_OF FairMQ
)
```

or similar. What do you think?

dennisklein: This reminds me, we should also add `find_package2` to FairLogger, which now has an optional dependency to Boost as well. With [generate_package_dependencies()](https://github.com/FairRootGroup/FairMQ/blob/master/cmake/FairMQLib.cmake#L213-L239) one can generate a cmake string, that can be exported in the cmake package, which defines `${FairLogger_Boost_VERSION}` and `${FairLogger_Boost_COMPONENTS}` - basically for any `find_package2` call, if `PUBLIC` and `VERSION`/`COMPONENTS` are specified. Made an issue FairRootGroup/FairLogger#14.

Then it would read

```cmake
find_package2(PUBLIC Boost
  VERSION 1.67 ${FairMQ_Boost_VERSION} ${FairLogger_Boost_VERSION}
  COMPONENTS thread system timer program_options random filesystem chrono exception regex serialization log log_setup atomic date_time signals ${FairMQ_Boost_COMPONENTS} ${FairLogger_Boost_COMPONENTS}
)
```

or

```cmake
find_package2(PUBLIC Boost
  VERSION 1.67
  COMPONENTS thread system timer program_options random filesystem chrono exception regex serialization log log_setup atomic date_time signals
  ADD_REQUIREMENTS_OF FairMQ FairLogger
)
```

Acutally, I like second one even better, will add it at some point.

dennisklein: This topic has **another aspect**: The reason for exporting the public package dependency versions/components, is obviously, that we do not have to hardcode any requirements of the dependent package. But what about the list of dependencies itself? Currently, we hardcode it. For Boost, it is somewhat hidden, because we depend on it directly in FairRoot, FairMQ, and FairLogger. E.g. FairMQ also exports a variable [`${FairMQ_PACKAGE_DEPENDENCIES}`](https://github.com/FairRootGroup/FairMQ/blob/master/cmake/FairMQLib.cmake#L218) that could be used to loop over all public FairMQ dependencies and "`find_package`" them, even if FairRoot does not directly depends on them. 

So far, so good, but this is not enough. Imagine multiple dependencies exporting such a list of their dependencies and they overlap partly. Then one would first have to "fuse" those lists together preserving their topological properties (doable, if the order is the original call order !currently not the case! those lines would be a [bug](https://github.com/FairRootGroup/FairMQ/blob/master/cmake/FairMQLib.cmake#L263-L264)) and figure out a call order together with the direct dependencies FairRoot "finds" explicitely. All doable, but it is not yet done.

dennisklein: FairRootGroup/FairMQ#142 FairRootGroup/FairMQ#143 FairRootGroup/FairMQ#144

MohammadAlTurany:
@dennisklein Thanks, I will adapt to your suggestions.

cmake/modules/FairMacros.cmake
MohammadAlTurany: My understanding here is to clean and simplify the CMake output so that the users see in one shot what he has. The real issue of dependancies between packages is solved in AliBuild or FairSoft, OR?

dennisklein: Those tools will build, install and decide on the package versions. But the public package dependencies must be discovered by each project for the whole subtree.

E.g. if a dependency uses a symbol of yet another library in its header, that is included in a FairRoot source file. Then, we only know about the target name, that we need to link against. But that target is not defined unless we `find_package` the righr dependency.

MohammadAlTurany: agree

dennisklein: In a way, it is a standard way of doing, what the `config.sh` does.

cmake/modules/ROOTMacros.cmake
```diff
@@ -210,7 +210,7 @@ MACRO (GENERATE_ROOT_TEST_SCRIPT SCRIPT_FULL_NAME)
                   )
   ENDIF(FAIRROOTPATH)
 
-  EXEC_PROGRAM(/bin/chmod ARGS "u+x  ${new_path}/${shell_script_name}")
+  EXEC_PROGRAM(/bin/chmod ARGS "u+x  ${new_path}/${shell_script_name}" OUTPUT_VARIABLE tmp)
```
fuhlig1: Instead of exec_program which is deprecated use execute_process which also allows to suppress the output to stdout

execute_process(COMMAND /bin/chmod u+x ${new_path}/${shell_script_name}
                             OUTPUT_QUIET)

cmake/modules_old/FindCLHEP.cmake
```diff
@@ -1,8 +1,8 @@
  ################################################################################
```
fuhlig1: Do we still need this Find module? We use the CLHEP which comes with Geant4.

cmake/modules_old/FindGENERATORS.cmake
```diff
@@ -9,7 +9,7 @@ if (GENERATORS_LIBRARY_DIR)
   SET (GENERATORS_LIBRARY_DIR GENERATORS_LIBRARY_DIR-NOTFOUND)
```
fuhlig1: Do we still need this Find module?

CMakeLists.txt
```diff
@@ -5,34 +5,21 @@
...
-  cmake_policy(SET ${p} NEW)
-  endif()
-endforeach()
-
+set(CMAKE_INSTALL_MESSAGE LAZY)
 # Set project version
 SET(FAIRROOT_MAJOR_VERSION 18)
 SET(FAIRROOT_MINOR_VERSION 0)
-SET(FAIRROOT_PATCH_VERSION 1)
+SET(FAIRROOT_PATCH_VERSION 3)
```
fuhlig1: We are already at patch version 6 for the v18.0 branch. So for the development branch we should choose a higher number or decide on a different numbering scheme.

cmake/modules/CheckCompiler.cmake
```diff
@@ -201,44 +207,45 @@ if (CMAKE_SYSTEM_NAME MATCHES Darwin)
       Execute_Process(COMMAND gfortran -print-file-name=libgfortran.dylib
                       OUTPUT_VARIABLE FORTRAN_LIBDIR
                      )
-      Get_Filename_Component(FORTRAN_LIBDIR ${FORTRAN_LIBDIR}
+      Get_Filename_Component(--FORTRAN_LIBDIR ${FORTRAN_LIBDIR}
```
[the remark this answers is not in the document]
MohammadAlTurany: No meaning, this is what happened if you find and replace!!

cmake/modules/ROOTMacros.cmake
MohammadAlTurany: Ok, will do

CMakeLists.txt
MohammadAlTurany: Ok, I will change it to 7 now.

fuhlig1:
The problem I see is that we create a new patch we will have again a conflict. Somehow we have to distinguish patches from the development.

PR #881, opened 2019-03-12
fixed config.sh:FAIRROOTPATH="", replaced with build/install path

klenze:
FairRootConfig.sh did set FAIRROOTPATH="", as did config.sh, which meant one had to set it manually after sourcing that file. This should fix it. 

cmake/modules/WriteConfigFile.cmake looks for config.sh.in in FAIRROOTPATH, so instead of setting that, I have introduced a new variable TARGETFAIRROOTPATH. 

No changes in WriteConfigFile.cmake, so it should not affect downstream users of that macro (e.g. r3broot) as long as they use their own config.sh.in.

Checklist:


dennisklein:
Can you give an example, where you need to set `FAIRROOTPATH` manually, if you are **not** working in a downstream experiment repo?

(If you are working from an experiment repo, then `FAIRROOTPATH` is meant to be set by you, so that the build system can find FairRoot.)

klenze:
> (If you are working from an experiment repo, then FAIRROOTPATH is meant to be set by you, so that the build system can find FairRoot.)

I do not understand this. FairRootConfig.sh sets many variables to the current FairRoot installation:
ROOT_INCLUDE_PATH, VMCWORKDIR, FAIRLIBDIR, LD_LIBRARY_PATH.
So if you are sourcing FairRootConfig.sh, you are already committed to that version of FairRoot. Why not set FAIRROOTPATH as well, instead of requiring it to be set explicitly? Or is there a use case where you would recommend sourcing FairRootConfig.sh from on FairRoot installation and then setting FAIRROOTPATH to a different installation?

For comparison, thisroot.sh sets ROOTSYS (instead of just amending PATH, LD_LIBRARY_PATH etc and leaving it to the user to set ROOTSYS explicitly), and I think this is the correct behavior. 

I am probably just misunderstanding when  I should source config.sh/FairRootConfig.sh at all. If I have an experiment repo, I guess I am supposed to set SIMPATH, FAIRROOTPATH instead. If I want to run FairRoot binaries (e.g. fairroot-config , parmq-server), sourcing the environment is also insufficient because it won't adjust the PATH. So the use case for FairRootConfig.sh is just when I want to commit to use the software in fairsoft, like cmake, root, geant? Should it not be in FairSoft then?

Personally, I would the idea to commit to use one version of FairSoft plus FairRoot by sourcing a single script, instead of explicitly running:

export SIMPATH=...
export PATH=$SIMPATH/bin:$PATH #for cmake
source $SIMPATH/bin/thisroot.sh
source  $SIMPATH/bin/geant.sh
export FAIRROOTPATH=... #for exp repos

If my patch was merged, FairRootConfig.sh would be the that script. When is it useful to set FAIRROOTPATH explicitly to "" (the current behavior, in my experience) instead?

Cheers, 
   Philipp

dennisklein:
The canonical workflow with downstream experiment repos should be like this:
* **BUILD TIME**: Define `SIMPATH` and `FAIRROOTPATH` env variables, when building the experiment repo.
* **RUN TIME**: Source the generated `config.sh` script in the build directory (it should also set `SIMPATH` and `FAIRROOTPATH` at this point, if not, consider it a bug). As far as I can see, (not all of) the experiment repos seem to have a proper installation target. So, you basically work from the build directory.

The `FairRootConfig.sh` should not be needed, if you are working with a downstream experiment repo. All the necessary environment should be set by the downstream `config.sh` (@fuhlig1 Correct me, if I am telling bogus here.)

fuhlig1:
You are right. The variable FAIRROOTPATH is only needed for downstream experiment. I don’t see where it is needed wehen you only work within a FairRoot installation.
That ROOT uses the ROOTSYS variable is because the use this information internally at several places. I am not aware that we do such a thing in FairRoot.

The reason why the variable is in the config script is because the same template is used by the downstream experiments.

klenze:
Ok, thanks for clearing this up for me.

PR #889, opened 2019-04-25
Add support for TVirtualMCSensitiveDetector.

kresan:
Inherit FairModule from TVirtualMCSensitiveDetector, which allows to use VMC stepping.
Tested with Tutorial1.
Requires changes in detector interface.


sawenzel:
Could you please summarize shortly what is the idea behind this change? Will it be backward compatible/optional or will it lead to deep changes?

kresan:
It is still work in progress. Idea is to simplify handling of sensitive detectors in FairRoot. Currently foreseen as a default solution. Would require changes in FairDetector interface, which are not backward compatible. Changes are in ProcessHits method.

kresan:
@MohammadAlTurany , Looking into it again, I have realised that the experiments will need to change ProcessHits method in every detector class

karabowi:
can you shortly describe, what the changes would have to be?

kresan:
The declaration:
```
virtual Bool_t ProcessHits(FairVolume* v = 0) = 0;
```
will be changed to:
```
virtual void ProcessHits()=0;
```

karabowi:
Looking at the code, you've also moved the function from FairDetector to FairModule.
What is the reason to remove its argument and return value?
Can't it stay it used to be?

MohammadAlTurany:
This change I needed only if you went to use the sensitive detector  or did I miss something here?

fuhlig1:
I thought there was a wrapper in FairRoot which calls the old function if the new one isn't available.

kresan:
@karabowi We used to take care about bookkeeping of sensitive volumes in FairRoot. Idea of this PR is to let VMC do it. The new function is VMC requirement.
@MohammadAlTurany I thought we want to get read of the parts where we have to handle sensitive volumes on our own completely. Does it make sense to keep both options?
I agree to Florian that a wrapper can help here.
But, we have to be aware, that FairVolume parameter of the overloaded methods might have been used in the experiment code. How do we handle this with the wrapper?

MohammadAlTurany:
One could get the current volume and the corresponding FairVolume and call the old method, at least for some time to allow the experiments to move smoothly to the new one!

karabowi:
With the current state of this branch, there seems to be no any MC points stored in the output files.

kresan:
@karabowi thanks for testing. Probably something went wrong while rebasing. The whole thing will be revisited completely. I will open another PR soon.

base/sim/FairMCApplication.h
```diff
@@ -209,6 +215,16 @@ class FairMCApplication : public TVirtualMCApplication
...
 
+    /**
+     * Add module to the list of sensitive detectors.
+     */
+    void AddSensitiveModule(std::string volName, FairModule* module);
+
+    /**
+     * Method introduced temporarily. It should go awway with DEPRACATED Bool_t FairDetector::ProcessHits()
+     */
+    FairVolume* GetFairVolume();
```
ChristianTackeGSI, line 226: What about using C++ deprecations?

```suggestion
    /**
     * \deprecate Method introduced temporarily. It should go awway with DEPRACATED Bool_t FairDetector::ProcessHits()
     */
    [[deprecated]] FairVolume* GetFairVolume();
```

(And doxygen deprecations in addition?)

ChristianTackeGSI (commented):
Just some more random comments.

base/sim/FairDetector.cxx
```diff
@@ -124,6 +125,22 @@ void FairDetector::Initialize()
...
 }
 
+// -------------------------------------------------------------------------
+
+void FairDetector::ProcessHits()
+{
+    static const auto printOnce = [this] {
+        LOG(warning) << "Calling depracated Bool_t FairDetector::ProcessHits(FairVolume* vol = 0) for \"" << GetName()
+                     << "\".";
+        LOG(warning) << "        Replace with void FairDetector::ProcessHits(). ";
```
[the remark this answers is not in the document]
ChristianTackeGSI: Okay, having looked at it again… Now I understand more!

Maybe rephrase to somethig like "Please override / implement ProcessHits on " << GetName() << " instead of the old ProcessHits(FairVolume*)".

fuhlig1:
I started testing the Pull Request for CbmRoot and had several problems which are spread over the whole application stack. For my tests I used FairSoft apr21p2 and FairRoot v18.6.7 as baseline since these are the default versions used by CbmRoot. To compare to the PR I cherry-picked the changes from the PR into v18.6.7.

Compilation of the patches FairRoot as well as CbmRoot worked but when running out transport tests all of them failed with an assertion in the Initialize() function of one of our detector classes. The assertion checked that the function was only called once but obviously it was called twice. 
When deriving FairModule from TVirtualMCSensitiveDetector the Initialize() function of Detectors/Modules is called from VMC and we need to remove calling the function from out FairMCApplication.
After the change in FairRoot and several fixes in CbmRoot I was able to run the tests without errors. When comparing the results produced with the patched and unpatched versions they were identical for most of our detector systems but for two of them only part of the MCPoints were produced.
After two days of debugging I probably found the problem in TGeant3. If I am correct the issue is already fixed with v4-0 which is the default for FairSoft apr22. I am currently compiling  the new FairSoft version to confirm that with this new version the problem is fixed.

Preliminary comparison of execution times of the patched and unpatched FairRoot versions don't show any observable difference.

ChristianTackeGSI:
@fuhlig1
Could you add this double-calling-prevention assert to some of the FairRoot tests/examples please? Maybe open it in a new PR, so that we get this test before we merge this PR?

fuhlig1:
The requested test case is described in #1195 and implemented with #1196.

fuhlig1:
@kresan, 

sorry but I think I messed up the PR a minute ago. There was a conflict which I tried to solve using the web editor but it seems to me that instead off doing a rebase the webpage did a merge. Could you please have a look.

kresan:
@fuhlig1, I reverted back your merge action. Now we need to resolve conflicts properly.

fuhlig1:
@kresan,

thanks. Could you solve the conflicts. I fear if I do it from the web frontend we run into the same problem.

ChristianTackeGSI:
We noticed in this morning's meeting, that we should consider raising the minimum required geant3 version to the one containing some fixes relevant to this pull request.

If we do so, it should be done in the CMakeLists.txt.

ChristianTackeGSI:
Also before we merge this one, we probably should do a rebase/cleanup of the commits.

And before we merge it, we should get clear on our release planning (this currently targets 19.2, not 19.0).

base/sim/FairMCApplication.cxx
```diff
@@ -1338,4 +1291,29 @@ void FairMCApplication::UndoGeometryModifications()
...
+void FairMCApplication::ConstructSensitiveDetectors()
+{
+    std::map<std::string, FairModule*> cloneVolumeMap;
+
+    for (auto const& x : fMapSensitiveDetectors) {
+        std::string volName = x.first;   //.substr(0, x.first.find("#", 0));
+        if (volName.find('#') != std::string::npos) {
+            volName = volName.substr(0, volName.find("#", 0));
+            std::map<std::string, FairModule*>::iterator it;
+            it = cloneVolumeMap.find(volName);
```
ChristianTackeGSI, line 1303:
```suggestion
            auto it = cloneVolumeMap.find(volName);
```

base/sim/FairMCApplication.cxx
```diff
@@ -599,6 +548,8 @@ void FairMCApplication::Stepping()
         }
     }
     if (fRadLenMan || fRadMapMan) {
+        Int_t copyNo;
+        Int_t id = fMC->CurrentVolID(copyNo);
         id = fMC->CurrentVolID(copyNo);
```
ChristianTackeGSI, line 553: You're assigning twice the same?
```suggestion
        Int_t id = fMC->CurrentVolID(copyNo);
```

base/sim/FairMCApplication.cxx
```diff
@@ -1338,4 +1291,29 @@ void FairMCApplication::UndoGeometryModifications()
     gGeoManager->ClearPhysicalNodes(kFALSE);
 }
 
-ClassImp(FairMCApplication);
+void FairMCApplication::ConstructSensitiveDetectors()
+{
+    std::map<std::string, FairModule*> cloneVolumeMap;
+
+    for (auto const& x : fMapSensitiveDetectors) {
+        std::string volName = x.first;   //.substr(0, x.first.find("#", 0));
```
ChristianTackeGSI, line 1299: Could you try a string_view here?

```suggestion
        std::string_view volName = x.first;   //.substr(0, x.first.find("#", 0));
```

base/sim/FairMCApplication.cxx
```diff
@@ -1338,4 +1291,29 @@ void FairMCApplication::UndoGeometryModifications()
...
+                continue;
+            cloneVolumeMap[volName] = x.second;
+            LOG(debug) << "FairMCApplication::ConstructSensitiveDetectors really do " << volName;
+        }
+        LOG(debug) << "FairMCApplication::ConstructSensitiveDetectors really do " << volName;
+        TVirtualMC::GetMC()->SetSensitiveDetector(volName, x.second);
+    }
+}
+
+void FairMCApplication::AddSensitiveModule(std::string volName, FairModule* module)
```
ChristianTackeGSI, line 1315: Could you try a string_view here as well?
```suggestion
void FairMCApplication::AddSensitiveModule(std::string_view volName, FairModule* module)
```

base/sim/FairMCApplication.h
```diff
@@ -319,6 +336,10 @@ class FairMCApplication : public TVirtualMCApplication
     /** Current state */
     FairMCApplicationState fState;   //!
 
+    /** List of sensitive detectors.
+     * To be used with TVirtualMCSensitiveDetector. */
+    std::map<std::string, FairModule*> fMapSensitiveDetectors;
+
```
ChristianTackeGSI, line 342: Can you move this into the `private:` section? Or do we need it as protected?

base/sim/FairModule.h
```diff
@@ -15,6 +15,7 @@
 #include "FairLogger.h"
 #include "FairRun.h"         // for FairRun
 #include "FairRuntimeDb.h"   // for FairRuntimeDb
+#include "TVirtualMCSensitiveDetector.h"
```
ChristianTackeGSI, line 18: This is not an include of FairRoot, right? Please use `<>`.
```suggestion
#include <TVirtualMCSensitiveDetector.h>
```

examples/advanced/propagator/src/FairTutPropDet.cxx
```diff
@@ -110,7 +111,7 @@ Bool_t FairTutPropDet::ProcessHits(FairVolume* vol)
         stack->AddPoint(kTutProp);
     }
 
-    return kTRUE;
+    return;
```
ChristianTackeGSI, line 114: Why a `return` at the end of a function?

```suggestion
```

ChristianTackeGSI (commented):
I have added a few minor comments.

In reality, I would like to discuss the whole design with people interested. Because I think, it's making things more and more complex and unmaintainable in the long run. It took me a few iterations to start to understand why things are done this way. And I have some ideas / questions about this.

base/sim/FairMCApplication.cxx
karabowi: Ain't that easy. Should I change the whole map? 
```
FairRoot/base/sim/FairMCApplication.cxx:1301:38: error: no matching member function for call to 'find'
            auto it = cloneVolumeMap.find(volName);
/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk/usr/include/c++/v1/map:1405:14: note: candidate function not viable: no known conversion from 'std::string_view' (aka 'basic_string_view<char>') to 'const std::map<std::string, FairModule *>::key_type' (aka 'const std::string') for 1st argument
    iterator find(const key_type& __k)             {return __tree_.find(__k);}
```

base/sim/FairMCApplication.cxx
karabowi: Same here:
```
FairMCApplication.cxx:1315:27: error: no viable overloaded operator[] for type 'std::map<std::string, FairModule *>' (aka 'map<basic_string<char, char_traits<char>, allocator<char>>, FairModule *>')
    fMapSensitiveDetectors[volName] = module;
    ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
```

base/sim/FairMCApplication.cxx
ChristianTackeGSI: No, leave the map as is. I was just hoping for the string_view to be a little more efficient. But it needs to be converted to a std::string for those errors anyway.

Thanks for trying it!

base/sim/FairMCApplication.cxx
ChristianTackeGSI: Thanks again for trying it!

PR #896, opened 2019-06-26
[WIP] Get boost provided cmake configs to work

ktf:
This is what I need to get FairRoot to build with boost 1.70.0 and automatically generated targets.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,15 +206,21 @@ EndIf(FAIRSOFT_EXTERN)
 if(NOT DEFINED Boost_NO_SYSTEM_PATHS)
   Set(Boost_NO_SYSTEM_PATHS TRUE)
 endif()
-#Set(Boost_NO_BOOST_CMAKE TRUE)
+
+if(NOT DEFINED Boost_NO_BOOST_CMAKE)
+  set(Boost_NO_BOOST_CMAKE TRUE)
+endif()
+
 #Message("-- Looking for Boost ...")
 # If an older version of boost is found both of the variables below are
 # cached and in a second cmake run, a good boost version is found even
 # if the version is to old.
 # To overcome this problem both variables are cleared before checking
 # for boost.
-#Unset(Boost_INCLUDE_DIR CACHE)
-#Unset(Boost_LIBRARY_DIRS CACHE)
+if (Boost_NO_BOOST_CMAKE)
+  Unset(Boost_INCLUDE_DIR CACHE)
+  Unset(Boost_LIBRARY_DIRS CACHE)
+endif()
 
 find_package2(PUBLIC Boost
   VERSION 1.67 ${FairMQ_Boost_VERSION}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -221,6 +221,10 @@ find_package2(PUBLIC Boost
   COMPONENTS thread system timer program_options random filesystem chrono exception regex serialization log log_setup atomic date_time ${FairMQ_Boost_COMPONENTS}
 )
 
+if (NOT Boost_INCLUDE_DIRS)
+  get_target_property(Boost_INCLUDE_DIRS Boost::serialization INTERFACE_INCLUDE_DIRECTORIES)
+  message(FATAL "${Boost_INCLUDE_DIRS}")
+endif()
 
 If (Boost_FOUND)
   Set(Boost_Avail 1)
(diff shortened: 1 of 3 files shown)


ktf:
Any chance any of this can be picked up upstream?

CMakeLists.txt
```diff
@@ -13,7 +13,8 @@
 # and the subproject (i.e: CbmRoot, Pandaroot,... etc).
 
 # Check if cmake has the required version
-CMAKE_MINIMUM_REQUIRED(VERSION 3.11.0 FATAL_ERROR)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.12.0 FATAL_ERROR)
+set(BUILD_SHARED_LIBS ON)
```
dennisklein: While this solves the issue, [this option has more implications](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) which I would prefer to avoid - better set it as a `-D` command line option when configuring FairRoot. Also, it has been [acknowledged as a bug by upstream](https://github.com/boostorg/boost_install/issues/5) and fixed for next release.

ktf: Ok.

CMakeLists.txt
```diff
@@ -205,21 +206,25 @@ EndIf(FAIRSOFT_EXTERN)
...
 
 find_package2(PUBLIC Boost
   VERSION 1.67 ${FairMQ_Boost_VERSION}
-  COMPONENTS thread system timer program_options random filesystem chrono exception regex serialization log log_setup atomic date_time signals ${FairMQ_Boost_COMPONENTS}
+  COMPONENTS thread system timer program_options random filesystem chrono exception regex serialization log log_setup atomic date_time ${FairMQ_Boost_COMPONENTS}
 )
 
+if (NOT Boost_INCLUDE_DIRS)
+  get_target_property(Boost_INCLUDE_DIRS Boost::serialization INTERFACE_INCLUDE_DIRECTORIES)
+  message(FATAL "${Boost_INCLUDE_DIRS}")
```
ktf: This obviously is spurious but you get the idea...

dennisklein: Yup, this should not hurt though, so no objections here.

CMakeLists.txt
```diff
@@ -205,21 +206,25 @@ EndIf(FAIRSOFT_EXTERN)
 if(NOT DEFINED Boost_NO_SYSTEM_PATHS)
   Set(Boost_NO_SYSTEM_PATHS TRUE)
 endif()
-Set(Boost_NO_BOOST_CMAKE TRUE)
+#Set(Boost_NO_BOOST_CMAKE TRUE)
```
dennisklein: Here I would prefer something like

```cmake
if(NOT DEFINED Boost_NO_BOOST_CMAKE)
  set(Boost_NO_BOOST_CMAKE TRUE)
endif()
```

That way you can override our default in the alidist recipe via `-D`.

CMakeLists.txt
```diff
@@ -205,21 +206,25 @@ EndIf(FAIRSOFT_EXTERN)
...
 #Message("-- Looking for Boost ...")
 # If an older version of boost is found both of the variables below are
 # cached and in a second cmake run, a good boost version is found even
 # if the version is to old.
 # To overcome this problem both variables are cleared before checking
 # for boost.
-Unset(Boost_INCLUDE_DIR CACHE)
-Unset(Boost_LIBRARY_DIRS CACHE)
+#Unset(Boost_INCLUDE_DIR CACHE)
+#Unset(Boost_LIBRARY_DIRS CACHE)
```
dennisklein: This section including the comment we should put inside a `if(Boost_NO_BOOST_CMAKE)` block I think.

CMakeLists.txt
dennisklein: FYI, with [CMake 3.15.0-rc2](https://gitlab.kitware.com/cmake/cmake/merge_requests/3438) and later this is addressed upstream.

aphecetche: if the intent was to break then that should be message(FATAL_ERROR XXX) as message(FATAL XXX) just printouts FATALXXX

ktf: @aphecetche yes, that was the intent, but I should have not committed it in any case, so it's fine...

CMakeLists.txt
ktf: Ok.

CMakeLists.txt
ktf: @dennisklein Yeah, but I am stuck to 3.13...

dennisklein:
I included the work from here in #898 which supersedes this PR. Let's continue the discussion there.

PR #899, opened 2019-06-28
CMake: Fix yaml-cpp lookup

dennisklein:
Separate PR to be included in the release as discussed.


dennisklein:
The failing tests on demac016
https://cdash.gsi.de/testDetails.php?test=5495304&build=207952
https://cdash.gsi.de/testDetails.php?test=5495305&build=207952
https://cdash.gsi.de/testDetails.php?test=5495525&build=207957
https://cdash.gsi.de/testDetails.php?test=5495526&build=207957

are all because of an outdated CMake version. Geant4 10.5.1 seems to require CMake 3.14. See also https://gitlab.kitware.com/cmake/cmake/commit/acfe53c58817c662b935fbe0f0443de298371731

dennisklein:
The last error https://cdash.gsi.de/testDetails.php?test=5495831&build=207964 indicates a broken FairSoft installation ?!
```
objc[170]: Class RunStopper is implemented in both /Users/alfaci/fairsoft/dev/lib/root/libCore.6.16.so (0x1065f1388) and /Users/turany/fairsoft/install/release/FairSoft/may18/lib/root/libCore.6.12.so (0x10c9c6618). One of the two will be used. Which one is undefined.
```

The user `alfaci` should really not see libraries in the home directory of user `turany`. Will reinstall those environments next week. Maybe this fixes also the compile warnings of sort
```
ld: warning: text-based stub file /System/Library/Frameworks//CoreAudio.framework/Versions/A/CoreAudio.tbd and library file /System/Library/Frameworks//CoreAudio.framework/Versions/A/CoreAudio are out of sync. Falling back to library file for linking.
```
which amount to over 500.

dennisklein:
@fuhlig1 @MohammadAlTurany I added the Geant3/4 data directories to the summary table. Only question that remains: should `G3SYS` point to `share/geant3/data` or just `share/geant3`?

FairRoot dep summary 2019-07-01 20-03-41

dennisklein:
@fuhlig1 Reinstalled the `dev` and `dev_mt` environments on demac016 with CMake 3.14.5 which fixed both issues described above. The warnings are still present though.

MohammadAlTurany:
@dennisklein it is needed for the gcalor package, the data files chetc.dat and xsneut.dat are expected in  $G3SYS/lib

fuhlig1:
Hi Dennis,

it has to point to $SIMPATH/share/geant3 and it expects a subdirectory data which contains the two files. So the G3SYS which you have in your screenshot is the correct one.

Ciao

Florian

PR #906, opened 2019-08-02
Hotfix v18.0 to support CMake 3.15

### LINKED ISSUE #905: CMake 3.15.1 crashes when building FairRoot v18.0.7
When using CMake 3.15.1 with FairRoot v18.0.7, CMake stops with the following error message

```
CMake Error at MbsAPI/CMakeLists.txt:19 (CHECK_SYMBOL_EXISTS):
  Unknown CMake command "CHECK_SYMBOL_EXISTS".
```

When using older CMake versions like 3.11.1 (part of FairSoft may18) or 3.13.4 (part of FairSoft jun19) the problem does not show up.

To reproduce the error use CMake 3.15.1 together with FairRoot v18.0.7

The expected behavior is that CMake runs without problems when using CMake 3.15.1 

The problem was seen with macosx 10.12 but I expect that the problem also occurs for other versions of macosx and also on Linux. 

The error do not show up with FairRoot v18.2.0 and FairRoot dev.

Looking into the problem one sees that the unknown CMake command "CHECK_SYMBOL_EXISTS" is defined in the file CheckSymbolExists.cmake. This file has to be loaded before usage by the command "include(CheckSymbolExists)" which is missing in FairRoot v18.0.7. In FairRoot v18.2.0 and FairRoot dev the include statement is in the main CMakeLists.txt. Most probably in older CMake versions the file is included implicitly by some other include which changed with the latest CMake version.

The fix is rather easy. Either add the missing include statement in the main CMakeLists.txt as done in FairRoot v18.2.0 or add it in the specific CMakeLists.txt which has the problem.

dennisklein:
Tested on Fedora 30 with CMake `3.15.1` only.

Resolve #905

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,7 @@ SET(VMCWORKDIR ${CMAKE_SOURCE_DIR}/examples)
 # any of the features of the new standard
 Set(CheckSrcDir "${CMAKE_SOURCE_DIR}/cmake/checks")
 include(CheckCXX11Features)
+include(CheckSymbolExists)
 
 # FairRoot only uses C++11 or later, so we need an compiler which supports C++11
 # Check if the used compiler support C++11. If not stop with an error message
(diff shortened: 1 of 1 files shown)


dennisklein:
CI wont work. But this is a different issue, we need to preserve the old environments and move the Jenkinsfile out of the repo.

PR #909, opened 2019-08-06
Fix YAML include dirs (should not be relative path resolving to /usr/include)

davidrohr:
Fix usage of YAML-CPP include dirs, when they are in /usr/include.
Otherwise, CMake will add `-isystem /usr/[SOMETHING]/../include` to the gcc command line, which breaks GCC for some known weird reason.

---

Checklist:


CMakeLists.txt
```diff
@@ -248,8 +248,9 @@ EndIf()
 
 find_package2(PUBLIC yaml-cpp)
 if(yaml-cpp_FOUND AND TARGET yaml-cpp AND YAML_CPP_INCLUDE_DIR)
+  get_filename_component(YAML_CPP_INCLUDE_TMP ${YAML_CPP_INCLUDE_DIR} ABSOLUTE)
```
dennisklein:
```suggestion
  get_filename_component(YAML_CPP_INCLUDE_DIR "${YAML_CPP_INCLUDE_DIR}" ABSOLUTE)
```

I think it is safe to overwrite the existing `YAML_CPP_INCLUDE_DIR` variable here - no need for a new name.

davidrohr: Fine with me, the PR is editable by maintainers, so please adjust to your needs.

dennisklein: Mh, from the looks upstream seems to have [made this workaround obsolete](https://github.com/jbeder/yaml-cpp/commit/ab5f9259a4e67d3fe0e4832bd407a9e596e2d884) entirely after `0.6.2` (no release though). Eventually this will become:

```cmake
if(yaml-cpp_FOUND AND TARGET yaml-cpp AND YAML_CPP_INCLUDE_DIR AND yaml-cpp_VERSION VERSION_LESS 0.6.2)
  get_filename_component(YAML_CPP_INCLUDE_DIR "${YAML_CPP_INCLUDE_DIR}" ABSOLUTE)
  set_target_properties(yaml-cpp PROPERTIES INTERFACE_INCLUDE_DIRECTORIES ${YAML_CPP_INCLUDE_DIR})
endif()
```

Wanna go there immediately?

dennisklein: > so please adjust to your needs.

will do

dennisklein:
@davidrohr It wouldn't let me force push to your branch. If you give me a thumbs up in #911, I will merge that, or if you want to get my changes from my fork (branch `davidrohr_PR2`) in this PR, I can merge this one. However you prefer :) Thx for the contrib!

davidrohr:
Weird, it is definitely marked as "allow edits from maintainers", no idea what is the problem.

dennisklein:
> "allow edits from maintainers", no idea what is the problem.

Had this problem many times in the past with Alexey too. It seems to be depending on some additional condition, maybe we need to be member of the same org or just force-pushes are not possible or the checkbox needs to be checked right away? sometimes it works for me, sometimes not..

PR #910, opened 2019-08-06
Hotfix v18.2.1: Fix builds against FairSoft installations without simulation engines

### LINKED ISSUE #907: FairRoot with light installation of FairSoft (w/o simulation engines)
FairSoft offers the possibility to be installed without simulation engines (Geant3 / Geant4); this can be choosen in the questionnaire conducted by configure.sh. However, FairRoot does not compile against such a "light" FairSoft (see below), which makes the feature somehow useless.

Are there plans to help this situation by making FairRoot compilable w/o simulation engines?

Here is the compilation error with FairRoot v18.2.0 against FairSoft jun19p1 (light):
```
[ 62%] Linking CXX shared library ../../lib/libFairMCStepLogger.dylib
[ 62%] Built target FairMCStepLogger
[ 62%] Generating G__FairFastSimDict.cxx, G__FairFastSimDict_rdict.pcm, ../../../lib/libFairFastSim.rootmap
input_line_12:19:10: fatal error: 'TG4VUserFastSimulation.h' file not found
#include "TG4VUserFastSimulation.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~
Error: /Users/vfriese/Software/fairsoft/jun19p1_nosim/bin/rootcint: compilation failure (/Users/vfriese/Software/fairroot/build/base/sim/fastsim/G__FairFastSimDictd69d857cf3_dictUmbrella.h)
make[2]: *** [base/sim/fastsim/G__FairFastSimDict.cxx] Error 1
make[1]: *** [base/sim/fastsim/CMakeFiles/FairFastSim.dir/all] Error 2
make: *** [all] Error 2
```

dennisklein:
@kresan I removed the `FairGeoPassivePar.h` header inclusion from the templates because it seems to not be used (it does not link against the corresponding library and additionally this header file appears to be part of an example only?!). Pls veto, if there is a preferred way to fix this.

Fixes #907


dennisklein:
@fuhlig1 Just realized this additionally needs to be merged into `dev` then.

fuhlig1:
And a similar patch should also go to v18_patches. This is the version used by CBM.

dennisklein:
> And a similar patch should also go to v18_patches. This is the version used by CBM.

Just tried with `v18.0.8` against a *light* installation of FairSoft `jun19p1`. It worked just fine for me, it seems to not be affected.

dennisklein:
@fuhlig1 I removed the version bump commit.

PR #911, opened 2019-08-06
Fix YAML include dirs cont.

dennisklein:
Continuation of #909

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,6 +247,10 @@ Else()
 EndIf()
 
 find_package2(PUBLIC yaml-cpp)
+# Workaround missing exported include directories
+# Upstream has fixed this in https://github.com/jbeder/yaml-cpp/commit/ab5f9259a4e67d3fe0e4832bd407a9e596e2d884
+# Once an upstream release includes the fix above, adapt the following condition accordingly
+# if(yaml-cpp_FOUND AND TARGET yaml-cpp AND YAML_CPP_INCLUDE_DIR AND yaml-cpp_VERSION VERSION_LESS 0.6.2)
 if(yaml-cpp_FOUND AND TARGET yaml-cpp AND YAML_CPP_INCLUDE_DIR)
   get_filename_component(YAML_CPP_INCLUDE_DIR "${YAML_CPP_INCLUDE_DIR}" ABSOLUTE)
   set_target_properties(yaml-cpp PROPERTIES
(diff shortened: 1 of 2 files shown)


davidrohr:
This looks OK for the CMakeList, but the path is also used in `fairtools/MCConfigurator/CMakeLists.txt`, which I had to change as well in #909

dennisklein:
> also used in fairtools/MCConfigurator/CMakeLists.txt, which I had to change as well in #909

I saw, the variable should propagate downwards. I'll double check.

davidrohr:
OK, you are right, since you overwrite the variable it will be OK.

dennisklein:
Hm, I don't even see your case, both my system and non-system installations have even conflicting include dirs^^

```
╭─dklein@gamma ~/projects/FairRoot ‹ruby-2.6.0› ‹davidrohr/PR2*›
╰─➤  grep YAML_CPP_INCLUDE_DIR /usr/lib64/cmake/yaml-cpp/yaml-cpp-config.cmake
set(YAML_CPP_INCLUDE_DIR "/usr/include/yaml-cpp")
╭─dklein@gamma ~/projects/FairRoot ‹ruby-2.6.0› ‹davidrohr/PR2*›
╰─➤  grep -B1 YAML_CPP_INCLUDE_DIR /home/dklein/fairsoft/dev_5b980cf/lib64/cmake/yaml-cpp/yaml-cpp-config.cmake
get_filename_component(YAML_CPP_CMAKE_DIR "${CMAKE_CURRENT_LIST_FILE}" PATH)
set(YAML_CPP_INCLUDE_DIR "${YAML_CPP_CMAKE_DIR}/../../../include")
```

Since we include `<yaml-cpp/yaml.h>` the include path my system installation (Fedora 30) provides is not even effective... weird packaging...

dennisklein:
Anyways, coming back to the `fairtools/MCConfigurator/CMakeLists.txt`, I checked the absolute path is used. Merging.

davidrohr:
Hm, I have 0.6.2 as well. the YAML CMake Config file `/usr/lib64/cmake/yaml-cpp/yaml-cpp-config.cmake` sets
`set(YAML_CPP_INCLUDE_DIR "${YAML_CPP_CMAKE_DIR}/../../../include")`
which becomes `/usr/include` (with some [something]/../). If that is used as system include path (which happened in `fairtools/MCConfigurator/CMakeLists.txt`, it breaks the compilation.

PR #918, opened 2019-08-24
Fix deps

dennisklein:
Follow-up on question in alisw/alidist#1768.

@ktf Does this help? I believe the `liblz4` dep you see comes in transitively from ROOT's `libCore` (It has a dep to `liblzma` on my ROOT installation, but IIRC this depends on the ROOT config and/or version)


ktf:
I am checking it in the other PR. Thank you for the "off peak" work.

PR #919, opened 2019-08-26
Few changes to allow running of Geant4 transport in multithreaded mode.

karabowi:
FairMCApplication - added mutex lock in CloneForWorker() function, otherwise unstable
FairDetector      - DefineSensitiveVolumes() function runs only in multithreaded mode
Pixel             - CheckIfSensitive() function is needed by FairDetector to make detector active
                    CloneModule() function need by G4MT to copy detector

---

Checklist:

 [ * ] Rebased against `dev` branch
 [ * ] My name is in the resp. CONTRIBUTORS/AUTHORS file
 [ * ] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/base/sim/FairDetector.h
+++ b/base/sim/FairDetector.h
@@ -79,13 +79,13 @@ class FairDetector : public FairModule
       return fDetId;
     }
 
-    virtual void DefineSensitiveVolumes();
-
   protected:
     /** Copy constructor */
     FairDetector(const FairDetector&);
     /** Assignment operator */
     FairDetector& operator= (const FairDetector&);
+
+    void DefineSensitiveVolumes();
 
     Int_t fDetId; // Detector Id has to be set from ctr.
     FairLogger* fLogger;  //! /// FairLogger
--- a/base/sim/FairDetector.h
+++ b/base/sim/FairDetector.h
@@ -79,6 +79,8 @@ class FairDetector : public FairModule
       return fDetId;
     }
 
+    virtual void DefineSensitiveVolumes();
+
   protected:
     /** Copy constructor */
     FairDetector(const FairDetector&);
--- a/base/sim/FairMCApplication.cxx
+++ b/base/sim/FairMCApplication.cxx
@@ -606,7 +606,7 @@ TVirtualMCApplication* FairMCApplication::CloneForWorker() const
   // and pass some data from master FairRunSim object
   FairRunSim* workerRun = new FairRunSim(kFALSE);
   workerRun->SetName(fRun->GetName()); // Transport engine
-  workerRun->SetUserOutputFileName(fRun->GetUserOutputFileName());
+  workerRun->SetSink(fRun->GetSink()->CloneSink());
 
   // Trajectories filter is created explicitly as we do not call
   // FairRunSim::Init on workers
@@ -637,9 +637,8 @@ void FairMCApplication::InitOnWorker()
 
   // Generate per-thread file name
   // and create a new sink on worker
-  TString workerFileName = fRun->GetUserOutputFileName();
-  fRootManager->UpdateFileName(workerFileName);
-  fRun->SetSink(new FairRootFileSink(workerFileName));
+  // moved to CloneForWorker and specific sink->CloneSink();
+
   fRootManager->InitSink();
 
   // Cache thread-local gMC
--- a/base/sim/FairMCApplication.cxx
+++ b/base/sim/FairMCApplication.cxx
@@ -812,7 +812,10 @@ void FairMCApplication::FinishEvent()
 {
 // User actions after finishing of an event
 // ---
-  LOG(debug) << "FairMCMCApplication::FinishEvent: " << fRootManager->GetInstanceId();
+  LOG(debug) << "[" << fRootManager->GetInstanceId() << " FairMCMCApplication::FinishEvent: " << fMCEventHeader->GetEventID() << " (MC " << gMC->CurrentEvent() << ")";
+  if ( gMC->IsMT() && fRun->GetSink()->GetSinkType() == kONLINESINK ) { // fix the rare case when running G4 multithreaded on MQ
+    fMCEventHeader->SetEventID(gMC->CurrentEvent()+1);
+  }
 
   // --> Fill the stack output array
   fStack->FillTrackArray();
--- a/base/sim/FairMCApplication.cxx
+++ b/base/sim/FairMCApplication.cxx
@@ -65,6 +65,9 @@ class TParticle;
 #include <stdlib.h>                     // for getenv, exit
 #include <utility>                      // for pair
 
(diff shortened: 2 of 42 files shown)


base/sim/FairDetector.h
```diff
@@ -99,6 +99,8 @@ class FairDetector : public FairModule
       return fDetId;
     }
 
+    virtual void DefineSensitiveVolumes();
```
sawenzel: I am wondering if this is this a new **public** interface of FairDetector? Are clients supposed to implement it?
(To me it looks like there is a default implementation in FairDetector and possibly a non-virtual private method would be sufficient here)

karabowi: This part has been implemented by Ivana to make the MT running
and removed some time ago, also by Ivana.
It seems however, that without this part the G4MT is not working.

sawenzel: Sure. There is no question if this part is necessary (it seems to be called from somewhere). I am just wondering if the scope of the function can be reduced to non-virtual private as otherwise users might start to override and thereby breaking the simulation.

karabowi: That makes sense. Thanks.

examples/simulation/Tutorial1/src/FairTutorialDet1.cxx
```diff
@@ -165,6 +165,15 @@ void FairTutorialDet1::Reset()
   fFairTutorialDet1PointCollection->Clear();
 }
 
+Bool_t FairTutorialDet1::CheckIfSensitive(std::string name)
```
rbx:
```suggestion
Bool_t FairTutorialDet1::CheckIfSensitive(const std::string& name)
{
  if (name.find("tutdet") != std::string::npos) {
    return kTRUE;
  }
  return kFALSE;
}
```
to avoid string copy/allocation.

examples/simulation/Tutorial1/src/FairTutorialDet1.h
```diff
@@ -83,6 +83,8 @@ class FairTutorialDet1: public FairDetector
 
     virtual FairModule* CloneModule() const;
 
+    virtual Bool_t      CheckIfSensitive(std::string name);
```
rbx: perhaps simply `IsSensitive` is a more concise name?
E.g.: `if (pixelDetector.IsSensitive()) { ... }`. Ignore this if the name has to be like you have it...

examples/simulation/Tutorial1/src/FairTutorialDet1.cxx
rbx: or even just `return name.find("tutdet") != std::string::npos;` if `bool` converts to `Bool_t`

base/sim/FairMCApplication.cxx
```diff
@@ -599,13 +602,14 @@ void FairMCApplication::PreTrack()
 //_____________________________________________________________________________
 TVirtualMCApplication* FairMCApplication::CloneForWorker() const
 {
+  mtx.lock();
```
rbx:
```suggestion
  std::unique_lock<std::mutex> lock(mtx);
```
Use this and remove mtx.unlock(); at the end, unique_lock will do lock/unlock for you in its constructor/destructor. This will make the locking exception-safe.

examples/MQ/pixelDetector/src/Pixel.cxx
```diff
@@ -184,6 +197,15 @@ void Pixel::ConstructGeometry()
     ConstructASCIIGeometry<PixelGeo, PixelGeoPar>(Geo, "PixelGeoPar");
 }
 
+Bool_t Pixel::CheckIfSensitive(std::string name)
```
rbx:
```suggestion
Bool_t Pixel::CheckIfSensitive(const std::string& name)
{
  return name.find("Pixel") != std::string::npos;
}
```

examples/simulation/Tutorial1/src/FairFastSimExample.cxx
```diff
@@ -197,6 +194,15 @@ FairTutorialDet1Point* FairFastSimExample::AddHit(Int_t trackID,
     return new (clref[size]) FairTutorialDet1Point(trackID, detID, pos, mom, time, length, eLoss);
 }
 
+Bool_t FairFastSimExample::CheckIfSensitive(std::string name)
```
rbx:
```suggestion
Bool_t FairFastSimExample::CheckIfSensitive(const std::string& name)
{
  return name == "fast_sim_vol";
}
```

examples/simulation/Tutorial1/src/FairFastSimExample2.cxx
```diff
@@ -182,6 +175,15 @@ FairTutorialDet1Point* FairFastSimExample2::AddHit(Int_t trackID,
     return new (clref[size]) FairTutorialDet1Point(trackID, detID, pos, mom, time, length, eLoss);
 }
 
+Bool_t FairFastSimExample2::CheckIfSensitive(std::string name)
```
rbx:
```suggestion
Bool_t FairFastSimExample2::CheckIfSensitive(const std::string& name)
{
  return name == "fast_sim_vol_n2";
}
```

karabowi: The problem is that the function API is quite difficult to change. It is being used by many experiments already. 
The guts of the function is a different thing - I do agree that the std::string is nowadays better than the TString.

rbx: :(, I see, then at least change the implementation.
Or perhaps keep the bad one and add new one, and add `__attribute__((deprecated("Use Bool_t FairFastSimExample2::CheckIfSensitive(const std::string& name)")))` to the old signature. Then at least people will see a warning and hopefully change it. But because it is overloaded through inheritance, maybe the warning will not be triggered.

rbx:
Please squashed the commits with same name when this is ready. Looks like we missed it on the last PR.

examples/MQ/pixelDetector/src/devices/FairMQRunDevice.cxx
```diff
@@ -45,30 +49,61 @@ void FairMQRunDevice::SendObject(TObject* obj, const std::string& chan) {
...
+                    else {
+                        LOG(fatal) << "Object " << ObjStr->GetString() << " NOT FOUND!!!";
+                    }
+                }
+            }
+        }
+        if ( parts.Size() > 0 ) {
+            std::unique_lock<std::mutex> lock(mtx);
+            Send(parts,mi.first.data());
+            mtx.unlock();
```
rbx: you don't need this line, because the `std::unique_lock` will automatically unlock in it's destructor.

PR #921, opened 2019-09-10
Build eventdisplay only if ROOT has opengl.

kresan:
Check for output of
`root-config --has-opengl`

Do not build eventdisplay if ROOT has no OpenGL.


CMakeLists.txt
```diff
@@ -323,7 +323,11 @@ If(Geant3_FOUND)
  add_subdirectory (geane)
  add_subdirectory (trackbase)
 EndIf ()
-add_subdirectory (eventdisplay)
+
+execute_process(COMMAND ${ROOTSYS}/bin/root-config --has-opengl OUTPUT_VARIABLE ROOT_HAS_OPENGL)
```
dennisklein:
```suggestion
execute_process(COMMAND ${ROOT_CONFIG_EXECUTABLE} --has-opengl OUTPUT_VARIABLE ROOT_HAS_OPENGL)
```

Have not tried it, but this should be available, if I read our `cmake/FindROOT.cmake` correctly.

Once we switch to the ROOT CMake package, the `execute_process` can be dropped, because it populates a variable `${ROOT_opengl_FOUND}` which can be used directly as `if(ROOT_opengl_FOUND)`.

CMakeLists.txt
```diff
@@ -323,7 +323,11 @@ If(Geant3_FOUND)
...
+If (${ROOT_HAS_OPENGL} MATCHES "yes")
+  add_subdirectory (eventdisplay)
+EndIf ()
```
dennisklein: If you want to be nice to the user, you could add an else case with `message(STATUS "Not building the eventdisplay because ROOT was built w/o OpenGL support")` or similar.

CMakeLists.txt
kresan: done. thanks

CMakeLists.txt
kresan: also done

PR #933, opened 2020-02-28
Modernize CMake phase 2

dennisklein:
Phase 2 will break experiment repos.

* [x] Overhaul top-level `CMakeLists.txt`
* [x] Modernize/Remove find modules
* [x] Switch to ROOT imported targets
* [x] Define components
* [x] Introduce CMake package with component support
* [ ] Update project templates
* [x] Change build switch logic to fail fast
* [x] Add `RUN_STATIC_ANALYSIS` mode
* [x] Update `FairRootConfig.sh`
* [x] Modernize CXX standard mgmt
* [x] Remove outdated checks
* [x] Refactor version generation logic
* [x] Find external VMC if no `ROOT::VMC` available, see #934
* [x] Bundle GTest
* [x] Install headers to `<prefix>/include/fairroot`
* [x] Rename libs to `libfairroot-<TARGET_LOWERCASE>`
* [x] Export some CMake functions into separate modules

PR #934, opened 2020-03-02
Add linking with VMC library.

kresan:
- In the next FairSoft release VMC will be a standalone package. Add it as dependency to the main CMakeLists.txt
- VMC of ROOT is deprecated. Link against VMCLibrary from VMC.
- CI will obviously fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,6 +190,8 @@ find_package2(PUBLIC Protobuf)
 find_package2(PUBLIC msgpack)
 find_package2(PUBLIC Flatbuffers)
 
+find_package2(PUBLIC VMC)
+
 find_package2(PUBLIC Geant3)
 find_package2(PUBLIC Geant4)
 
@@ -494,6 +496,9 @@ if(PROJECT_PACKAGE_DEPENDENCIES)
       endif()
     elseif(${dep} STREQUAL yaml-cpp)
       get_filename_component(prefix ${YAML_CPP_INCLUDE_DIR}/.. ABSOLUTE)
+  elseif(${dep} STREQUAL VMC)
+      get_target_property(vmc_include VMCLibrary INTERFACE_INCLUDE_DIRECTORIES)
+      get_filename_component(prefix ${vmc_include}/../.. ABSOLUTE)
     elseif(${dep} STREQUAL Geant4VMC)
       string(REPLACE ":" ";" geant4vmc_include ${Geant4VMC_INCLUDE_DIRS})
       list(GET geant4vmc_include 0 geant4vmc_include)
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -133,8 +133,9 @@ target_link_libraries(${target} PUBLIC
   Proof
   RIO
   Tree
-  VMC
   $<$<BOOL:${ROOT_GDML_FOUND}>:Gdml>
+
+  VMCLibrary
 )
 
 target_compile_definitions(${target} PRIVATE
--- a/base/sim/fastsim/CMakeLists.txt
+++ b/base/sim/fastsim/CMakeLists.txt
@@ -43,9 +43,10 @@ target_link_libraries(${target} PUBLIC
 
   Geom # TGeoManager
   Core # Rtypes
-  VMC # TVirtualMC
   EG # TParticle
   Physics # TVector
+
+  VMCLibrary
 )
 
 fairroot_target_root_dictionary(${target}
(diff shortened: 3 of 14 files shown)


CMakeLists.txt
```diff
@@ -190,6 +190,8 @@ find_package2(PUBLIC Protobuf)
 find_package2(PUBLIC msgpack)
 find_package2(PUBLIC Flatbuffers)
 
+find_package2(PUBLIC Vmc)
```
dennisklein: Does this work? From the looks at the VMC project, it should be `find_package2(PUBLIC VMC)`?

base/CMakeLists.txt
```diff
@@ -110,6 +110,8 @@ target_include_directories(${target} PUBLIC
 
   # DELETE ME ONCE USING root targets
   ${ROOT_INCLUDE_DIR}
+
+  ${Vmc_INCLUDE_DIR}
```
dennisklein: Who sets this variable?

base/sim/fastsim/CMakeLists.txt
```diff
@@ -27,10 +27,13 @@ target_include_directories(${target} PUBLIC
 
   # TODO: DELETE ME ONCE USING root targets
   ${ROOT_INCLUDE_DIR}
+
+  ${Vmc_INCLUDE_DIR}
 )
 
 target_link_directories(${target} PUBLIC
   ${ROOT_LIBRARY_DIR}
+  ${Vmc_LIBRARY_DIR}
```
dennisklein: Same here, who sets this variable? Also it is not needed.

CMakeLists.txt
```diff
@@ -494,6 +496,9 @@ if(PROJECT_PACKAGE_DEPENDENCIES)
       endif()
     elseif(${dep} STREQUAL yaml-cpp)
       get_filename_component(prefix ${YAML_CPP_INCLUDE_DIR}/.. ABSOLUTE)
+    elseif(${dep} STREQUAL Vmc)
```
dennisklein: If the find_package argument is changed to `VMC`, then also change it here.

CMakeLists.txt
kresan: Fixed

base/CMakeLists.txt
kresan: Changed to VMC_INCLUDE_DIRS

base/sim/fastsim/CMakeLists.txt
kresan: Removed

CMakeLists.txt
kresan: Done

base/CMakeLists.txt
```diff
@@ -110,6 +110,8 @@ target_include_directories(${target} PUBLIC
 
   # DELETE ME ONCE USING root targets
   ${ROOT_INCLUDE_DIR}
+
+  ${VMC_INCLUDE_DIRS}
```
dennisklein: Are you sure this is needed?

If the imported `VMCLibrary` target does not define its `INTERFACE_INCLUDE_DIRECTORIES` property, we should define it on this level, see e.g. https://github.com/FairRootGroup/FairRoot/blob/4ab2d2b1f0392ad93b92026d70fe049c38665f7d/CMakeLists.txt#L271-L273 on how to do this right after the `find_package2` call in the top-level `CMakeLists.txt`.

Adding these kind of variables in each an every `CMakeLists.txt` is not good because it duplicates code and encourages the copy&paste mess we had in the past - ideally, a dependency of a library or executable should only be declared once per `CMakeLists.txt` in the `target_link_libraries` call.

kresan: removed VMC_INCLUDE_DIRS from sub-folders CMakeLists

dennisklein:
Ok, just installing vmc to the dev and dev_mt environments does not work, because it gets confused with the vmc builtin to ROOT, see e.g. https://cdash.gsi.de/testDetails.php?test=6369431&build=228726.

That means, we either need to bump the required ROOT version or add a check, if the detected ROOT ships its own vmc and then link against the old vmc instead of the new.

dennisklein:
An easy way to check is to delete `cmake/modules/FindROOT.cmake` and then

```cmake
find_package(ROOT OPTIONAL_COMPONENTS vmc)
if(ROOT_vmc_FOUND)
  set(vmc_lib ROOT::VMC)
else()
  find_package(VMC)
  set(vmc_lib VMCLibrary)
endif()
```

or similar, and then link against `${vmc_lib}`.

dennisklein:
@kresan In principle, I added this PR to #933.

kresan:
included in #933. closing

dennisklein:
Closing, implemented by #941.

PR #936, opened 2020-03-05
Link against the shared flatbuffers library

dennisklein:


dennisklein:
```
CMake Error at CMakeLists.txt:523 (get_target_property):
  get_target_property() called with non-existent target
  "flatbuffers::flatbuffers_shared".
```

@ChristianTackeGSI So, legacy FairSoft does not build the shared lib.

dennisklein:
closing, will be covered by #933.

PR #950, opened 2020-04-20
Improve adding container to the container factory

fuhlig1:
Add a function to add a container to the factory that also checks if the
container already exist. If the container already exist it is not added
again and kFALSE is returned. Otherwise it is added and kTRUE is returned.
The "normal" way to add containers is to use the TList data member directly
which will not report any errors when adding the container. Only when the
destructor of the runtime database is called one will see an error message

Error in <TList::Delete>: A list is accessing an object (0x55d2e2220550)
already deleted (list name = TList)

---

Checklist:

### DIFF
--- a/parbase/FairContFact.h
+++ b/parbase/FairContFact.h
@@ -55,6 +55,8 @@ class FairContFact : public TNamed
     FairParSet* getContainer(const char*);
     virtual FairParSet* createContainer(FairContainer*) { return 0; }
     virtual void activateParIo(FairParIo*) {}
+    /// @param[in] container Transfers ownership if return value is true
+    Bool_t AddContainer(FairContainer*);
 
   protected:
     TList* containers;   // all parameter containers managed by this factory
--- a/parbase/FairContFact.cxx
+++ b/parbase/FairContFact.cxx
@@ -218,3 +218,19 @@ void FairContFact::print()
         c->print();
     }
 }
+
+Bool_t FairContFact::AddContainer(FairContainer* cont)
+{
+    // Check if a container already exist in the List of containers
+    // If it alread exist print an error message and return kFALSE
+    // such that the user can handle the issue
+    if (nullptr != containers->FindObject(cont)) {
+        LOG(error) << "The container " << cont->GetName() << " already exist in the "
+                   << "container factory " << GetName() << ".\n"
+                   << "Duplicate container is not added.";
+        return kFALSE;
+    }
+
+    containers->Add(cont);
+    return kTRUE;
+}


fuhlig1:
Could you please check and merge.

parbase/FairContFact.h
```diff
@@ -56,6 +56,7 @@ class FairContFact : public TNamed
     FairParSet* getContainer(const char*);
     virtual FairParSet* createContainer(FairContainer*) {return 0;}
     virtual void activateParIo(FairParIo*) {}
+    Bool_t AddContainer(FairContainer*);
```
dennisklein, line 59: Please add a doxygen comment to clarify ownership semantics of the raw pointer, e.g.:

```suggestion
    /// @param[in] container Transfers ownership if return value is true
    Bool_t AddContainer(FairContainer* container);
```

dennisklein:
The need for this kind of *checking add function* is very worrisome. What does a return value `false` mean? It tells the caller that he is not the owner of the object (because the `FairContFact` instance is). But how can this be in the first place? The caller should already know this.

fuhlig1:
> The need for this kind of _checking add function_ is very worrisome. What does a return value `kFALSE` mean? It tells the caller that he is not the owner of the object (because the `FairContFact` instance is). But how can this be in the first place? The caller should already know this.

This only tells the caller that the framework could not add the object to the internal TList because an object with the same name already exist in the list. This is completely independent from the question who is the owner. 

The current way which is used all over the place in FairRoot and any derrived project is directly access the Internal TList (which is public) and call TList->Add directly. This works since years until you do a stupid copy and paste error. To use internal data members is a very bad idea but since the code is heavily used we can' change it without breaking all existing code. 

I agree that passing the raw pointer is not the desired  way. Since the function is not used anywhere yet I will try to implement a proper interface.

dennisklein:
> This only tells the caller that the framework could not add the object to the internal TList because an object with the same name already exist in the list.

True. Still, if it is possible that your calling code tries to add the same object (not just by name) twice to the `TList` (doesn't matter whether it is through `FairContFact` or directly on the `TList` member), it means that the calling code seems to have lost track whether it is the owner of the added object.

```
root [0] TList* l = new TList; // non-owning container
root [1] TObject* obj = new TObject; // caller owns it
root [2] l->Add(obj); // caller owns it
root [3] l->Add(obj); // caller owns it
root [4] delete l; // fine
root [5]
root [5] l = new TList;
root [6] l->SetOwner(kTRUE); // l now owning container
root [7] l->Add(obj); // caller does not own obj any more
root [8] l->Add(obj); // caller knows it should not call Add any more, because obj holds a non-owning pointer
root [9] delete l;
Error in <TList::Delete>: A list is accessing an object (0x5632af764490) already deleted (list name = TList)
```

I do not object this PR, especially if you intend to have a `TList` that actually represents a unique set of `TObject`s by name. From your issue description I got the impression you wanted to fix a double add of the same object (by reference), which would indicate a problem in the calling code.

PR #951, opened 2020-04-21
Adapt to new spack-based FairSoft

dennisklein:


dennisklein:
way too much broken. not going to happen like this.

PR #958, opened 2020-04-27
Fix mem mgmt of fLink member

### LINKED ISSUE #955: NULL protection missing
https://github.com/FairRootGroup/FairRoot/blob/e6cd7da7d60d0983cd0c1014a611376a1e8fe0c4/base/event/FairMultiLinkedData_Interface.h#L37

not urgent

dennisklein:
* Fix nullptr deref in dtor
* Fix memory leak in setter
* Clarify raw pointer semantics
* Add new setter overload

Resolves #955

---

It looks like we do not have test coverage of this code. @MohammadAlTurany @TobiasStockmanns @RalfKliemt Can you point me to some code/macro that calls the changed code, so we can gain some confidence that the code still does what it is expected to do?


karabowi:
Tested with Pandaroot - it works just as it was working before.

PR #962, opened 2020-04-29
Fix examples

dennisklein:
Attempt at fixing https://cdash.gsi.de/testDetails.php?test=6621457&build=234910


dennisklein:
This PR is serializing the whole test suite for stability gains and at the cost of a few minutes longer CI build times. I will have a look at a more fine-grained tuning at a later point in time (only some tests are moody when run in parallel).

PR #963, opened 2020-04-29
Format addendum

rbx:
`ClassDef` also should have a semicolon at the end, otherwise formatter will get lost in the following code.

Didn't see this in the first run because only few classes have code after `ClassDef`.


dennisklein:
pls rebase on current `dev` branch, thx

rbx:
done.

Issue #974, opened 2020-05-07
-Wstringop-overflow in Tutorial4


dennisklein:
On current [`dev`](https://github.com/FairRootGroup/FairRoot/commit/f6d0b62f3c76fbc997f716e864cde5ef4f13b40f) branch

```
../examples/simulation/Tutorial4/src/tools/FairTutorialDet4GeoHandler.cxx: In member function ‘Int_t FairTutorialDet4GeoHandler::VolId(const Text_t*) const’:
../examples/simulation/Tutorial4/src/tools/FairTutorialDet4GeoHandler.cxx:105:16: warning: ‘char* strncpy(char*, const char*, size_t)’ specified bound depends on the length of the source argument [-Wstringop-overflow=]
  105 |         strncpy(sname, name, len);
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~
../examples/simulation/Tutorial4/src/tools/FairTutorialDet4GeoHandler.cxx:101:27: note: length computed here
  101 |         Int_t len = strlen(name) - 1;
      |                     ~~~~~~^~~~~~
```

PR #978, opened 2020-05-14
fix issue #974 and some other warnings

MohammadAlTurany:
fix issue #974 and few other warnings


base/sim/FairGenericStack.h
```diff
@@ -98,8 +98,11 @@ class FairGenericStack : public TVirtualMCStack
     virtual void Register() {}
 
     virtual TClonesArray* GetListOfParticles() { return 0; }
-    virtual void SetParticleArray(TClonesArray* partArray) {}
-    virtual void SetParticleArray(TClonesArray* partArray, Int_t partFrom, Int_t partTo) {}
+    virtual void SetParticleArray(__attribute__((unused)) TClonesArray* partArray) {}
+    virtual void SetParticleArray(__attribute__((unused)) TClonesArray* partArray,
+                                  __attribute__((unused)) Int_t partFrom,
+                                  __attribute__((unused)) Int_t partTo)
+    {}
```
mattkretz, line 105: use `[[gnu::unused]]` or `[[maybe_unused]]` instead of non-standard attribute syntax

fuhlig1: or use

 virtual void SetParticleArray(TClonesArray*) {}
virtual void SetParticleArray(TClonesArray*, Int_t, Int_t) {}

MohammadAlTurany: Thanks! Ok I will use **[[maybe_unused]]**

mattkretz: Note that `[[maybe_unused]]` is C++17

examples/simulation/Tutorial4/src/tools/FairTutorialDet4GeoHandler.cxx
```diff
@@ -102,7 +102,7 @@ Int_t FairTutorialDet4GeoHandler::VolId(const Text_t* name) const
         if (name[len] != ' ') {
             return VolIdGeo(name);
         }
-        strncpy(sname, name, len);
+        memcpy(sname, name, len);
```
dennisklein, line 105: There are actually two more issues present in this code:
* buffer overflow if `name` is longer than 20 characters
* buffer underflow if `name` is empty

I suggest something like this (untested):

```diff
@@ -100,8 +100,5 @@
-        char sname[20];
-        Int_t len = strlen(name) - 1;
-        if (name[len] != ' ') {
-            return VolIdGeo(name);
-        }
-        strncpy(sname, name, len);
-        memcpy(sname, name, len);
-        sname[len] = 0;
-        return VolIdGeo(sname);
+        std::string volname(name);
+        while (!volname.empty() && volname.back() == ' ') {
+          volname.pop_back();
+        }
+        return VolIdGeo(volname.c_str());
```

This also does a proper `rtrim` which I assume was the original intention anyways (if not, replace the `while` with an `if`).

base/sim/FairGenericStack.h
MohammadAlTurany: > or use
> 
> virtual void SetParticleArray(TClonesArray*) {}
> virtual void SetParticleArray(TClonesArray*, Int_t, Int_t) {}

In some cases it did not work!

examples/simulation/Tutorial4/src/tools/FairTutorialDet4GeoHandler.cxx
MohammadAlTurany: Done

base/sim/FairGenericStack.h
MohammadAlTurany: I change it to  gnu-unused as we are still requiring C++11 and not 17.

base/sim/fastsim/FairFastSimModel.cxx
```diff
@@ -97,7 +97,7 @@ void FairFastSimModel::DoIt(const G4FastTrack& fastTrack, G4FastStep& fastStep)
             particle->GetPolarisation(polVect);
             G4ThreeVector pol(polVect.X(), polVect.Y(), polVect.Z());   // get polarisation
             G4double tim = particle->T() * 10.e9;                       // change from ns to s
-            G4double len = 0.;
+                                                                        //   G4double len = 0.;
```
mattkretz, line 100: If `len` is unused, better remove it. The next reader of the code will be rather puzzled, trying to understand the comment. :wink:

MohammadAlTurany: removed in #979

PR #980, opened 2020-05-19
Rename library name from libPassive to libExamplesPassive

karabowi:
Fixes the problem with duplicated library name in the
PandaRoot subproject when installing with spack.

---

Checklist:

[x] Rebased against `dev` branch
[x] My name is in the resp. CONTRIBUTORS/AUTHORS file
[x] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)


rbx:
Can we make all example libs have the same name pattern?

Now there are examples with `libName.so`, `libExName.so`, `libTutorialNumber.so`, `libNameTutorial.so`. Also there is `libFairTestDetector.so`.
Now this PR will add `libExamplesName.so` (why plural?). 

Full list (is that all?):
```
libMbsTutorial.so
libPropagatorExample.so
libFairTestDetector.so
libPassive.so
libMCStack.so
libGConfig.so
libExHistogramServer.so
libExLmd.so
libExParameters.so
libPixelAlt.so
libPixel.so
libSimMQ.so
libExSerialization.so
libFairRutherford.so
libTutorial1.so
libTutorial2.so
libTutorial4.so
```



How about just `libExName.so` for all of those? This seems to be the most common pattern (if one can find a pattern).

karabowi:
The examples were growing and the names are inconsistent... There are examples, tutorials, and so on... That would be like total rebranding. As for the Example_s_Passive. I chose the 's' because the parent dir name is 'examples' and because this library is used by multiple examples.

rbx:
I'm only talking about the library names (but i also don't mind a total rebranding, if it helps make thing more clear).

Just an idea:
```
libExMbs.so
libExPropagator.so
libExTestDetector.so
libExPassive.so
libExMCStack.so
libExGConfig.so
libExHistogramServer.so
libExLmd.so
libExParameters.so
libExPixelAlt.so
libExPixel.so
libExSimMQ.so
libExSerialization.so
libExRutherford.so
libExSimulation1.so
libExSimulation2.so
libExSimulation4.so
```
If these are installed maybe also add a FairRoot prefix, e.g. `libFairRootExPassive.so` to avoid conflicts in the future.

dennisklein:
> If these are installed maybe also add a FairRoot prefix, e.g. libFairRootExPassive.so to avoid conflicts in the future.

If stuff gets renamed now in bulk, pls consider renaming to `libfairroot-<targetnamelowercase>.so`. I will do this in my cmake branch https://github.com/dennisklein/FairRoot/blob/6f2d4fddbfe0cd85dbeae7d60b2211ac3fd79c92/cmake/FairRootLib.cmake#L306-L313 anyways.

```cmake
function(fairroot_add_standard_library_properties target)
  string(TOLOWER ${target} __libname__)
  set_target_properties(${target} PROPERTIES
    VERSION "${PROJECT_VERSION}"
    SOVERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}"
    LIBRARY_OUTPUT_NAME ${PROJECT_NAME_LOWER}-${__libname__}
    LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/${PROJECT_INSTALL_LIBDIR}")
endfunction()
```

karabowi:
Dennis, do you suggest to rename them to `libfairroot-<targetnamelowercase>.so` by hand?
Or to use your function? If second, then where do you propose to have the function in?

dennisklein:
> Dennis, do you suggest to rename them to `libfairroot-<targetnamelowercase>.so` by hand?
> Or to use your function? If second, then where do you propose to have the function in?

The function is just meant as an example. Rename by setting the target property `LIBRARY_OUTPUT_NAME` so you don't have to change the target names all over the place. But thats just in case you want to add the fairroot prefix really now in this PR.

karabowi:
I do not. I prefer to go with the examples libraries only. Leave the adding of the fairroot prefix to your big PR.

PR #993, opened 2020-07-30
trying without -inlineInputHeader in rootcint macro

klenze:
rootcint's -inlineInputHeader has at least two drawbacks:
* any errors in the headers will not be attributed to $file:$line, but instead to input_line_12:861:13 or something. 
* #pragma once will not work, leaving us stuck with manually handling include guards like it's 2005 or something. 

So let us try running without that option. 

(It compiled locally here, now let's see if travis likes it as well.)
---

Checklist:

* [ x ] Rebased against `dev` branch
* [ X ] My name is in the resp. CONTRIBUTORS/AUTHORS file 
* [ ? ] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)


klenze:
Seems like ex_sim_tutorial1_pythia8_TGeant3 timeouted once. It was also completed several times. 

https://alfa-ci.gsi.de/blue/rest/organizations/jenkins/pipelines/FairRootGroup/pipelines/FairRoot/branches/PR-993/runs/1/log/?start=0 is not really readable for me. It seems like the 54 tests are started repeatedly (e.g. "43/54" is found 8 times, and so is "ex_sim_tutorial1_pythia8_TGeant3". Furthermore, it seems like the outputs of the eight test blocks are interlaced, so it is rather difficult to say exactly what crashed.)

It would be great if the line
`Drop site:https://cdash.gsi.de/submit.php?project=FairRoot`
would be followed by another line which tells me URL where I can access the relevant files, instead of having to crawl through the builds manually (which I won't). Alternatively, I would feel dumping failed tests directly to the log would also be ok for me. (For timeout cases, I would prefer running a gdb script which does a backtrace, core dump and kill.)

dennisklein:
> It would be great if the line `Drop site:https://cdash.gsi.de/submit.php?project=FairRoot` would be followed by another line which tells me URL where I can access the relevant files

I am also looking for a solution like this since forever. So far our conclusion is that at cdash submission time the cdash server has not yet decided on the id which makes it impossible with current cdash versions to provide a proper link at submission time. Ideally I would want to put this link behind the "Details" hyperlink in the Github Checks UI section in the bottom... You can also find discussion about this in the relevant cdash/cmake forums/lists which come to the same conclusion.

dennisklein:
> [ ? ] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

Since you put a question mark, let me hint how to improve your commit message:

```
Remove -inlineInputHeader from rootcint macro

The option has at least two drawbacks:
* Any errors in the headers will not be attributed to $file:$line,
  but instead to input_line_12:861:13 or something, and
* #pragma once will not work, leaving us stuck with manually
  handling include guards like it's 2005 or something.
```

dennisklein:
For reference, the cdash link for the latest build is: https://cdash.gsi.de/buildSummary.php?buildid=242488. The output for the timed out test is missing unfortunately.

Issue #992, opened 2020-07-28
FairMixedSource::BGWindowWidthNo description is wrong


TobiasStockmanns:
**Describe the bug**
The description of the first parameter of `void BGWindowWidthNo(UInt_t background, UInt_t Signalid);`says that: `*@param background :  Number of background Events for one signal`.
If one sets the value to 1 one should expect that for each signal event a background event is taken.

But in the method `Double_t value = 1.0 / background`is calculated. Which generates only signal events (100%). If one wants to have the same number of signal and background events one has to set background to 2.

**Expected behavior**
Either the description or the calculation of the rate should be fixed.

Issue #991, opened 2020-07-28
FairMixedSource never takes the first event


TobiasStockmanns:
**Describe the bug**
In `FairRunAna::Init()``fRootManager->ReadEvent(0)`is called. If the source is a `FairMixedSource` the first event of the background or the signal event is taken and the the index for the next event is increased by one.
Thus when reading-in the data the first event of the background or the signal file is event 1 and not 0.

**Expected behavior**
After the `Init()`the event counter should be reset.

Issue #990, opened 2020-07-28
FairMixedSource takes runId from background or signal file


TobiasStockmanns:
**Description**
`FairRunAna::Init()` calls `FairSource::ReadEvent(0)` to get the runId to init the runtimeDB.
If the `FairMixedSource` is used it is not defined from which file the runId is taken. Either from the background file or one of the signal files. If the runId does not match with the given par file it causes a crash.

**Expected behavior**
It should be choosable from which file the runId is taken to have the proper parameters taken from the par file.

PR #994, opened 2020-08-12
Fix Issues 990, 991, 992

MohammadAlTurany:
Solving issue #990, #991, #992 

1. User can choose where to take the parameters from in macro by calling UseRunIdFromBG() or UseRunIdfromSG(UInt_t identifier) for the mixed source 
2. New method (SpecifyRunId) is implemented in FairSource to avoid using ReadEvent(0) to get the RunId, this solves also the problem in #991 
3. Calculate the SB Ratio properly (Bug #992)
4. Remove obsolete code from FairRunAna


dennisklein:
@MohammadAlTurany I squashed and reworded your commits a bit: https://github.com/dennisklein/FairRoot/tree/issue_990. If you want, I can force-push this to your branch.

MohammadAlTurany:
Yes, looks much better! please push it. Thanks.

PR #997, opened 2020-08-19
CMake: Improve Pythia8 data dir detection

dennisklein:


dennisklein:
failing tests unrelated

PR #1000, opened 2020-08-27
Refactoring event display

DanielWielanek:
Refactoring event display.

- Now three separate class are needed to draw object:
- - one represents FairTask
- - second represents object
- - third represents object editor
- New class for graphical cuts added
- Class for draw GeoTracks added
- Examples how to draw MC tracks (with propagator)
- Example of usage of xml file to configuration of Event Display
- Examples how to draw reconstructed tracks (with matched MC tracks and hits)
- Helpers for drawing matched tracks with hits
- Track cuts removed from event manager to proper editors
- New options added to FairEventManager
-- global transparency
-- switching dark/light background
-- option of drawing animations (by event/ by time) in form of screenshots - supported currently by geotrack (TODO: adding support of time animations of hits/points)

---

Checklist:


dennisklein:
Congrats to PR #1000 :)

Find the logs for the failed format check here: https://alfa-ci.gsi.de/blue/rest/organizations/jenkins/pipelines/FairRootGroup/pipelines/FairRoot/branches/PR-1000/runs/1/nodes/16/steps/60/log/?start=0

You can format your code by running `git clang-format` on your branch.

eventdisplay/tracks/FairEveTrack.h
```diff
@@ -0,0 +1,52 @@
...
+#ifndef FAIREVETRACK_H_
+#define FAIREVETRACK_H_
+
+#include <TEveTrack.h>
+#include <TEveTrackPropagator.h>
+#include <TVector3.h>
+
+class FairEveTrack : public TEveTrack
+{
+    Int_t fNPoints;
```
dennisklein, line 18: Doesn't `Size()` already count the same?

eventdisplay/tracks/FairEveTrack.h
```diff
@@ -0,0 +1,52 @@
...
+    static Float_t fgWorldSize;
```
dennisklein, line 19: I'd rather avoid new global state, can we find a better place to store the world size? Would a specialized `FairEveTrackPropagator : TEveTrackPropagator` be a suitable place to hold the world size?

eventdisplay/tracks/FairEveTracks.h
```diff
@@ -0,0 +1,116 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

eventdisplay/tracks/FairGeoTracksDraw.cxx
```diff
@@ -0,0 +1,33 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

eventdisplay/tracks/FairGeoTracksDraw.h
```diff
@@ -0,0 +1,34 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/advanced/propagator/macros/eventDisplay.C
```diff
@@ -0,0 +1,59 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/advanced/propagator/src/CMakeLists.txt
```diff
@@ -32,6 +35,11 @@ target_compile_definitions(${target} PUBLIC ${defs})
 target_include_directories(${target} PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/trackbase
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/eventdisplay
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/eventdisplay/tracks
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/trackbase/
+
```
dennisklein, line 42:
```suggestion
```

This is no longer needed (and allowed ;) ). Include directories are only set locally on the CMake library targets they belong to. By linking against such defined targets CMake will construct the proper include directory search path.

examples/advanced/propagator/src/FairEveRecoTracksExample.cxx
```diff
@@ -0,0 +1,255 @@
...
+                el->SetRnrSelfChildren(kTRUE, kTRUE);
+            }
+        }
+    }
+    gEve->Redraw3D(kFALSE);
+}
+
+FairEveRecoTracksExample::~FairEveRecoTracksExample()
+{
+    // TODO Auto-generated destructor stub
```
dennisklein, line 254: ?

examples/advanced/propagator/src/FairEveRecoTracksExample.cxx
```diff
@@ -0,0 +1,255 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/advanced/propagator/src/FairEveRecoTracksExample.h
```diff
@@ -0,0 +1,58 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/advanced/propagator/src/FairEveRecoTracksExampleEditor.cxx
```diff
@@ -0,0 +1,75 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/advanced/propagator/src/FairEveRecoTracksExampleEditor.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <TGedFrame.h>
+
+class FairEveRecoTracksExampleEditor : public TGedFrame
+{
+    FairEveMinMaxCut *fPtCut;
+    FairEveMinMaxCut *fEtaCut;
+    FairEveMinMaxCut *fEnergyCut;
+    FairEveRecoTracksExample *fRecoTracks;
+    FairEveBoolCut *fDrawMC;
```
dennisklein, line 23:
```suggestion
    std::unique_ptr<FairEveMinMaxCut> fPtCut;
    std::unique_ptr<FairEveMinMaxCut> fEtaCut;
    std::unique_ptr<FairEveMinMaxCut> fEnergyCut;
    FairEveRecoTracksExample *fRecoTracks;
    std::unique_ptr<FairEveBoolCut> fDrawMC;
```

Avoid owning raw pointers if possible. See [R.smart](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rsmart-smart-pointers) and [R20](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r20-use-unique_ptr-or-shared_ptr-to-represent-ownership) in particular.

examples/advanced/propagator/src/FairEveRecoTracksExampleEditor.h
```diff
@@ -0,0 +1,37 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/advanced/propagator/src/FairRecoTracksDraw.cxx
```diff
@@ -0,0 +1,33 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/advanced/propagator/src/FairRecoTracksDraw.h
```diff
@@ -0,0 +1,30 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/common/eventdisplay/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+################################################################################
+# Copyright (C) 2014-2019 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  #
```
dennisklein, line 2:
```suggestion
# Copyright (C) 2014-2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  #
```

examples/common/eventdisplay/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+#                                                                              #
+#              This software is distributed under the terms of the             #
+#              GNU Lesser General Public Licence (LGPL) version 3,             #
+#                  copied verbatim in the file "LICENSE"                       #
+################################################################################
+
+set(target ExEvendDisplay)
```
dennisklein, line 9:
```suggestion
set(target ExEventDisplay)
```

examples/common/eventdisplay/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+set_target_properties(${target} PROPERTIES ${PROJECT_LIBRARY_PROPERTIES})
+
+target_include_directories(${target} PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/trackbase
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/eventdisplay
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/eventdisplay/tracks
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/trackbase/
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/examples/common/mcstack
```
dennisklein, line 30:
```suggestion
```

examples/common/eventdisplay/EventDisplayExampleLinkDef.h
```diff
@@ -0,0 +1,22 @@
+/********************************************************************************
+ *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
```
dennisklein, line 2:
```suggestion
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
```

examples/common/eventdisplay/EventDisplayExampleLinkDef.h
```diff
@@ -0,0 +1,22 @@
...
+ *                                                                              *
+ *              This software is distributed under the terms of the             * 
+ *              GNU Lesser General Public Licence (LGPL) version 3,             *  
+ *                  copied verbatim in the file "LICENSE"                       *
+ ********************************************************************************/
+// $Id: PassiveLinkDef.h,v 1.1.1.1 2005/06/23 07:14:26 dbertini Exp $
```
dennisklein, line 8:
```suggestion
```

examples/common/eventdisplay/FairEveMCTracks.cxx
```diff
@@ -0,0 +1,170 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/common/eventdisplay/FairEveMCTracks.cxx
```diff
@@ -0,0 +1,170 @@
...
+        field = new FairField();
+    }
+    fRK = new FairRKPropagator(field);
+    fPDG = TDatabasePDG::Instance();
+    return FairEveTracks::Init();
+}
+
+FairEveMCTracks::~FairEveMCTracks()
+{
+    // TODO Auto-generated destructor stub
```
dennisklein, line 169: ?

examples/common/eventdisplay/FairEveMCTracks.h
```diff
@@ -0,0 +1,51 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/common/eventdisplay/FairEveMCTracksEditor.cxx
```diff
@@ -0,0 +1,77 @@
...
+    use = fPdgCut->GetValue(pdg_code);
+    fMCTracks->SetPdgCut(pdg_code, use);
+    fMCTracks->SetShowPrimSec(fPrimary->GetValue(), fSecondary->GetValue());
+    TGedFrame::Update();
+    fMCTracks->Repaint();
+}
+
+FairEveMCTracksEditor::~FairEveMCTracksEditor()
+{
+    // TODO Auto-generated destructor stub
```
dennisklein, line 76: ?

examples/common/eventdisplay/FairEveMCTracksEditor.cxx
```diff
@@ -0,0 +1,77 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/common/eventdisplay/FairEveMCTracksEditor.h
```diff
@@ -0,0 +1,40 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/common/eventdisplay/FairEveMCTracksEditor.h
```diff
@@ -0,0 +1,40 @@
...
+
+class FairEveMCTracksEditor : public TGedFrame
+{
+    FairEveMinMaxCut *fPtCut;
+    FairEveMinMaxCut *fEtaCut;
+    FairEveMinMaxCut *fEnergyCut;
+    FairEveBoolCut *fPrimary;
+    FairEveBoolCut *fSecondary;
+    FairEveIntCut *fPdgCut;
+    FairEveMCTracks *fMCTracks;
```
dennisklein, line 25: From the looks most of these members are leaking, are they not?

examples/common/eventdisplay/FairMCTracksDraw.cxx
```diff
@@ -0,0 +1,33 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/common/eventdisplay/FairMCTracksDraw.h
```diff
@@ -0,0 +1,34 @@
+/*
```
dennisklein, line 1:
```suggestion
/********************************************************************************
 *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/
/*
```

examples/common/eventdisplay/FairEveMCTracksEditor.h
DanielWielanek: I make them unique pointers, so I hope this solve the issue.
Btw. is the order of declaration access specifiers (public/protected/private) must be keep?
I noticed that I always start with private and ends with public, should I change order?
And one more question what an order I should use at the same level of access - members and later methods?

dennisklein: > Btw. is the order of declaration access specifiers (public/protected/private) must be keep?
> I noticed that I always start with private and ends with public, should I change order?
> And one more question Btw. is the order of declaration access specifiers (public/protected/private) must be keep?

The order of class members determines the order of initialization (and potentially destruction in reverse-order of initialization) regardless of any access specifier. The order also determines at which offset in the type this member is stored in memory. However, the order of the access specifiers does not matter, it is just convenient to group many members with the same access specifier to not have to write it again many times - but you can have an access specifier as often as you want and also multiple times.

dennisklein: Please add `#include <memory>` where you use [`std::unique_ptr<>`](https://en.cppreference.com/w/cpp/memory/unique_ptr), thx!

DanielWielanek: One more question.
I disscused those updated with Florian and Muhammad, we decided that we have to break backward compatibility - I mean FairMCTracks will not be longer added to event display as a task.
I applied patched to proper macros but I cannot test some of them - e.g. examples/simulation/rutherford/macros/eventDisplay.C does't run because the function (eventdisplay) doesnt have any type to return.

dennisklein: > but I cannot test some of them - e.g. examples/simulation/rutherford/macros/eventDisplay.C does't run because the function (eventdisplay) doesnt have any type to return.

This is a bug, thx for noticing! Feel free to apply the necessary fix, e.g.

```diff
--- a/examples/simulation/rutherford/macros/eventDisplay.C
+++ b/examples/simulation/rutherford/macros/eventDisplay.C
@@ -5,7 +5,7 @@
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
  *                  copied verbatim in the file "LICENSE"                       *
  ********************************************************************************/
-eventDisplay(TString mcEngine = "TGeant3")
+void eventDisplay(TString mcEngine = "TGeant3")
 {

     TString inFile = "data/test_" + mcEngine + ".mc.root";
```

eventdisplay/tracks/FairEveTrack.h
dennisklein: e.g.
```
root [0] TEveTrack* t(new TEveTrack())
root [1] t->Size()
(int) 0
root [2] t->SetNextPoint(0,0,1)
(int) 0
root [3] t->Size()
(int) 1
```
?

examples/common/eventdisplay/FairEveMCTracksEditor.h
DanielWielanek: I think I fixed all of them.

eventdisplay/tracks/FairEveTrack.h
DanielWielanek: I'm not sure if this is solve issue. We still have to hold one global variable - FairEveTrackPropagator. Otherwise settings might be different for different tasks.

eventdisplay/tracks/FairEveTracks.h
```diff
@@ -0,0 +1,123 @@
...
+     */
+    void ResetGroup();
+    FairEventManager *GetEventManager() const { return fEventManager; };
+    TObjArray *GetTracksList() const { return fEveTrList; };
+    /**
+     *
+     * @param tr
+     * @return track group based on parameter
+     */
+    virtual TEveTrackList *GetTrackGroup(void *tr);
```
dennisklein, line 70: Now you added a base class implementation which does not depend on the `void*` argument at all? Again, what is the use case? Who is calling this function?

I was thinking more in a direction like this:
```diff
diff --git a/eventdisplay/tracks/FairEveGeoTracks.cxx b/eventdisplay/tracks/FairEveGeoTracks.cxx
index f32c89f57..2520afcd5 100644
--- a/eventdisplay/tracks/FairEveGeoTracks.cxx
+++ b/eventdisplay/tracks/FairEveGeoTracks.cxx
@@ -132,9 +132,8 @@ void FairEveGeoTracks::Repaint()
     gEve->Redraw3D(kFALSE);
 }

-TEveTrackList *FairEveGeoTracks::GetTrackGroup(void *tr)
+TEveTrackList *FairEveGeoTracks::GetTrackGroup(TGeoTrack *track)
 {
-    TGeoTrack *track = static_cast<TGeoTrack *>(tr);
     TParticle *P = static_cast<TParticle *>(track->GetParticle());
     fTrackGroup = nullptr;
     for (Int_t i = 0; i < GetTracksList()->GetEntriesFast(); i++) {
diff --git a/eventdisplay/tracks/FairEveGeoTracks.h b/eventdisplay/tracks/FairEveGeoTracks.h
index 9af902fd5..a127cf511 100644
--- a/eventdisplay/tracks/FairEveGeoTracks.h
+++ b/eventdisplay/tracks/FairEveGeoTracks.h
@@ -46,7 +46,7 @@ class FairEveGeoTracks : public FairEveTracks
     Bool_t CheckCuts(TGeoTrack *tr);
     void DrawTrack(Int_t id);
     void DrawAnimatedTrack(Int_t id);
-    TEveTrackList *GetTrackGroup(void *tr);
+    TEveTrackList *GetTrackGroup(TGeoTrack *tr);

   public:
     FairEveGeoTracks();
diff --git a/eventdisplay/tracks/FairEveTracks.cxx b/eventdisplay/tracks/FairEveTracks.cxx
index ecb5fd037..9f5762216 100644
--- a/eventdisplay/tracks/FairEveTracks.cxx
+++ b/eventdisplay/tracks/FairEveTracks.cxx
@@ -73,16 +73,3 @@ InitStatus FairEveTracks::Init()
 }

 FairEveTracks::~FairEveTracks() {}
-
-TEveTrackList *FairEveTracks::GetTrackGroup(void *tr)
-{
-    if (fTrackGroup == nullptr) {
-        fTrPropagator = new TEveTrackPropagator();
-        fTrackGroup = new TEveTrackList("tracks", fTrPropagator);
-        fTrackGroup->SetMainColor(kRed);
-        GetTracksList()->Add(fTrackGroup);
-        gEve->AddElement(fTrackGroup, this);
-        fTrackGroup->SetRnrLine(kTRUE);
-    }
-    return fTrackGroup;
-}
diff --git a/eventdisplay/tracks/FairEveTracks.h b/eventdisplay/tracks/FairEveTracks.h
index 60c53d294..167bacd06 100644
--- a/eventdisplay/tracks/FairEveTracks.h
+++ b/eventdisplay/tracks/FairEveTracks.h
@@ -62,12 +62,6 @@ class FairEveTracks
     void ResetGroup();
     FairEventManager *GetEventManager() const { return fEventManager; };
     TObjArray *GetTracksList() const { return fEveTrList; };
-    /**
-     *
-     * @param tr
-     * @return track group based on parameter
-     */
-    virtual TEveTrackList *GetTrackGroup(void *tr);

   public:
     FairEveTracks();
diff --git a/examples/advanced/propagator/src/FairEveRecoTracksExample.cxx b/examples/advanced/propagator/src/FairEveRecoTracksExample.cxx
index 642fe09d5..752835d36 100644
--- a/examples/advanced/propagator/src/FairEveRecoTracksExample.cxx
+++ b/examples/advanced/propagator/src/FairEveRecoTracksExample.cxx
@@ -160,9 +160,8 @@ void FairEveRecoTracksExample::DrawTrack(Int_t id)
     trList->AddElement(track);
 }

-TEveTrackList *FairEveRecoTracksExample::GetTrackGroup(void *tr)
+TEveTrackList *FairEveRecoTracksExample::GetTrackGroup(FairTutPropTrack *track)
 {
-    FairTutPropTrack *track = static_cast<FairTutPropTrack *>(tr);
     Color_t col = kYellow;
     TString pid = "neutral";
     Int_t q = track->GetParamFirst().GetQ();
diff --git a/examples/advanced/propagator/src/FairEveRecoTracksExample.h b/examples/advanced/propagator/src/FairEveRecoTracksExample.h
index 2d847e436..e031ce0a3 100644
--- a/examples/advanced/propagator/src/FairEveRecoTracksExample.h
+++ b/examples/advanced/propagator/src/FairEveRecoTracksExample.h
@@ -40,7 +40,7 @@ class FairEveRecoTracksExample : public FairEveTracks
   protected:
     Bool_t CheckCuts(FairTutPropTrack *tr);
     void DrawTrack(Int_t id);
-    TEveTrackList *GetTrackGroup(void *tr);
+    TEveTrackList *GetTrackGroup(FairTutPropTrack *tr);

   public:
     FairEveRecoTracksExample();
diff --git a/examples/common/eventdisplay/FairEveMCTracks.cxx b/examples/common/eventdisplay/FairEveMCTracks.cxx
index e2c67942e..a392a0c45 100644
--- a/examples/common/eventdisplay/FairEveMCTracks.cxx
+++ b/examples/common/eventdisplay/FairEveMCTracks.cxx
@@ -120,9 +120,8 @@ void FairEveMCTracks::DrawTrack(Int_t id)
     trList->AddElement(track);
 }

-TEveTrackList *FairEveMCTracks::GetTrackGroup(void *tr)
+TEveTrackList *FairEveMCTracks::GetTrackGroup(FairMCTrack *track)
 {
-    FairMCTrack *track = static_cast<FairMCTrack *>(tr);
     TString pid = Form("%i", track->GetPdgCode());
     fTrackGroup = nullptr;
     for (Int_t i = 0; i < GetTracksList()->GetEntriesFast(); i++) {
diff --git a/examples/common/eventdisplay/FairEveMCTracks.h b/examples/common/eventdisplay/FairEveMCTracks.h
index 34de1cd0d..439e5d1a6 100644
--- a/examples/common/eventdisplay/FairEveMCTracks.h
+++ b/examples/common/eventdisplay/FairEveMCTracks.h
@@ -35,7 +35,7 @@ class FairEveMCTracks : public FairEveTracks
   protected:
     Bool_t CheckCuts(FairMCTrack *tr);
     void DrawTrack(Int_t id);
-    TEveTrackList *GetTrackGroup(void *tr);
+    TEveTrackList *GetTrackGroup(FairMCTrack *tr);

   public:
     FairEveMCTracks();
diff --git a/examples/common/eventdisplay/FairMCTracksDraw.h b/examples/common/eventdisplay/FairMCTracksDraw.h
index 9b7db7e61..2b9666f88 100644
--- a/examples/common/eventdisplay/FairMCTracksDraw.h
+++ b/examples/common/eventdisplay/FairMCTracksDraw.h
@@ -28,7 +28,6 @@ class FairMCTracksDraw : public FairTask
   protected:
     Bool_t CheckCuts(FairMCTrack *tr);
     void DrawTrack(Int_t id);
-    TEveTrackList *GetTrackGroup(void *tr);

   public:
     FairMCTracksDraw();
```

If you cannot find a common type for the argument other than `void*`, why making it a virtual function? All your specializations static_cast it to a specific type unconditionally anyways - that means those are all just different specific function with nothing in common, right? What am I missing?

eventdisplay/tracks/FairEveGeoTracksEditor.h
```diff
@@ -0,0 +1,60 @@
...
+#include <GuiTypes.h>
+#include <Rtypes.h>
+#include <RtypesCore.h>
+#include <TGButton.h>
+#include <TGDoubleSlider.h>
+#include <TGFrame.h>
+#include <TGLabel.h>
+#include <TGNumberEntry.h>
+#include <TGedFrame.h>   // for TGedFrame
+#include <memory.h>
```
dennisklein, line 31:
```suggestion
#include <memory>
```

STL headers have no `.h`, see https://en.cppreference.com/w/cpp/header/memory

examples/advanced/propagator/src/FairEveRecoTracksExampleEditor.h
```diff
@@ -0,0 +1,45 @@
...
+ *		Warsaw University of Technology, Faculty of Physics
+ */
+#ifndef FAIREVERECOTRACKSEXAMPLEEDITOR_H_
+#define FAIREVERECOTRACKSEXAMPLEEDITOR_H_
+
+#include "FairEveCut.h"
+#include "FairEveRecoTracksExample.h"
+
+#include <TGedFrame.h>
+#include <memory.h>
```
dennisklein, line 23:
```suggestion
#include <memory>
```

STL headers have no `.h`, see https://en.cppreference.com/w/cpp/header/memory

examples/common/eventdisplay/FairEveMCTracksEditor.h
```diff
@@ -0,0 +1,48 @@
...
+ *		Warsaw University of Technology, Faculty of Physics
+ */
+#ifndef FAIREVEMCTRACKSEDITOR_H_
+#define FAIREVEMCTRACKSEDITOR_H_
+
+#include "FairEveCut.h"
+#include "FairEveMCTracks.h"
+
+#include <TGedFrame.h>
+#include <memory.h>
```
dennisklein, line 23:
```suggestion
#include <memory>
```

STL headers have no `.h`, see https://en.cppreference.com/w/cpp/header/memory

dennisklein (changes requested):

eventdisplay/tracks/FairEveTrack.h
dennisklein: > have to hold one global variable - FairEveTrackPropagator.

We should definitely avoid adding any C++ globals. Even global model parameters should be stored in C++ locals (e.g. as it is done with `TEveTrackPropagator`).

side notes:
* I realize FairRoot and ROOT are still full with unnecessary C++ globals, but if we never start avoiding them - at least with new code written in 2020 - then we will never get away from the problems they cause.
* If you wonder what's the problem with C++ globals, in a nutshell: they deny multithreading, limit composability with other libraries, and they may contribute to degraded performance scalability.
* As a rule of thumb, if possible, use C++ globals only in very rare cases, where you need to represent real global (computing) system resources. Physics model parameters, OTOH, are almost always avoidable as C++ globals.

---

Adding @MohammadAlTurany and @TobiasStockmanns to this discussion as they might have much deeper insight than me here on how to implement such a world size filter.

dennisklein: Also refer to [I.2: Avoid non-const global variables](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i2-avoid-non-const-global-variables)

eventdisplay/tracks/FairEveTracks.h
dennisklein: Also refer to [I.4: Make interfaces precisely and strongly typed](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i4-make-interfaces-precisely-and-strongly-typed)

DanielWielanek: I make this only to mark that by default track list is cleaned up and user should check. That's why there is allocation of track group without type checking in FairEveTracks (that in previous version was pure virtual).
But if you think that this version is better - ok.

eventdisplay/tracks/FairEveTrack.h
DanielWielanek: Actually I had another Idea - what about storing this variable in FairEventManager?
Theoretically this store information about geometry - therefore we could load this value from the size of the simulation - however I'm not sure how to get the size of the simulation. It seems that:
 Double_t Z = ((TGeoBBox*)gGeoManager->GetTopNode()->GetVolume()->GetShape())->GetDZ();
allows to get information about size directly from geoManager, however I'm not sure if TGeoBBox based class always represents the cave.

eventdisplay/tracks/FairEveTracks.h
DanielWielanek: Update: I think I found better solution - now this function is not virtual. User only in constructor of base class (FairEveTracks) define if he use TEveTrackList or FairEveTrackList (that accept any compound object).
Thanks to this he doesn't have to think about creation of track groups at all.

eventdisplay/tracks/FairEveTracks.cxx
```diff
@@ -74,12 +76,24 @@ InitStatus FairEveTracks::Init()
...
+    fTrackGroup = nullptr;
+    for (Int_t i = 0; i < GetTracksList()->GetEntriesFast(); i++) {
+        TEveTrackList *TrListIn = static_cast<TEveTrackList *>(GetTracksList()->At(i));
+        if (strcmp(TrListIn->GetName(), groupName) == 0) {
+            fTrackGroup = TrListIn;
+            break;
+        }
+    }
     if (fTrackGroup == nullptr) {
         fTrPropagator = new TEveTrackPropagator();
```
dennisklein, line 90: Can we make this a local variable and remove the member variable?

eventdisplay/tracks/FairEveTracks.h
```diff
@@ -64,13 +65,18 @@ class FairEveTracks
     TObjArray *GetTracksList() const { return fEveTrList; };
     /**
      *
-     * @param tr
+     * @param groupName - name of the track group
+     * @param color - group color
      * @return track group based on parameter
      */
-    virtual TEveTrackList *GetTrackGroup(void *tr);
+    TEveTrackList *GetTrackGroup(TString groupName, Color_t color);
```
dennisklein, line 72:
```suggestion
    TEveTrackList *GetTrackGroup(TString groupName, Color_t color) const;
```

*constness*: Depends on if we can get rid of the `fTrPropagator` and `fTrackGroup` member variables, see my comments in the cxx file.

```suggestion
    std::unique_ptr<TEveTrackList> GetTrackGroup(TString groupName, Color_t color);
```
Ownership:
* If we remove `fTrackGroup`, you are returning an owning pointer, see 
* If we do not remove `fTrackGroup` (and you intend `FairEveTracks` to own it), you need to
   * either make it a `std::unique_ptr<TEveTrackList>`, or
   * call delete on it in the dtor, otherwise it leaks.
* The same story with `fTrPropagator`, it is leaking.

eventdisplay/tracks/FairEveTracks.h
```diff
@@ -42,6 +42,7 @@ class FairEveTracks
     Bool_t fUsePt;
     Bool_t fUseEta;
     Bool_t fUseEnergy;
+    Bool_t fAcceptCompound;
```
dennisklein, line 45:
```suggestion
    Bool_t const fAcceptCompound;
```

Let's make it `const` since we only set it in the ctor.

dennisklein (changes requested):

eventdisplay/tracks/FairEveTracks.h
dennisklein: > Update: I think I found better solution - now this function is not virtual. User only in constructor of base class (FairEveTracks) define if he use TEveTrackList or FairEveTrackList (that accept any compound object).
Thanks to this he doesn't have to think about creation of track groups at all.

I like it! Made some technical comments though.

eventdisplay/tracks/FairEveTracks.h
dennisklein: > If we remove fTrackGroup, you are returning an owning pointer

I actually missed the `GetTracksList()->Add(TrackGroup);` so nevermind my comment about the return value, raw pointer should be fine.

eventdisplay/tracks/FairEveTracks.h
dennisklein: You decided against making it `const`?

dennisklein: See also https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-const. Basically, C++ has the wrong default on constness. I have not checked the rest of your PR for constness opportunities, but I strongly recommend to double check, if you have everything `const` where possible and meaningful.

eventdisplay/tracks/FairEveTracks.cxx
```diff
@@ -0,0 +1,99 @@
...
+
+InitStatus FairEveTracks::Init()
+{
+    fEventManager = FairEventManager::Instance();
+    return kSUCCESS;
+}
+
+FairEveTracks::~FairEveTracks() {}
+
+TEveTrackList *FairEveTracks::GetTrackGroup(TString groupName, Color_t color)
```
dennisklein, line 76: You are right, this cannot be a const member function. Let's maybe rename it then to `AddTrackGroup` or `CreateTrackGroup` to not confuse it with the Getter pattern. What do you think?

eventdisplay/tracks/FairEveTracks.cxx
```diff
@@ -0,0 +1,99 @@
...
+    if (TrackGroup == nullptr) {
+        TEveTrackPropagator *TrPropagator = new TEveTrackPropagator();
+        if (fAcceptCompound) {
+            TrackGroup = new FairEveRecoTrackList(groupName, TrPropagator);
+        } else {
+            TrackGroup = new TEveTrackList(groupName, TrPropagator);
+        }
+        TrackGroup->SetMainColor(color);
+        GetTracksList()->Add(TrackGroup);
+        gEve->AddElement(TrackGroup, this);
```
dennisklein, line 95:
```suggestion
        AddElement(TrackGroup);
```

see https://root.cern.ch/doc/master/TEveManager_8cxx_source.html#l00527

eventdisplay/tracks/FairEveTracks.cxx
dennisklein: Btw, what happens, if we call `GetTrackGroup(TString groupName, Color_t color)` twice with the same arguments? Do you want it to create the same track group twice or do you want it to return the first?

eventdisplay/tracks/FairEveTracks.h
DanielWielanek: One more remark - i was trying to add TEveTrackPropagator to list to properly remove them.
However if I try to remove them by hand I get "free()" error, it seems that they are somehow removed by gEve.
I write my own propagator to check if they are removed - and they are somehow removed by gEve.

eventdisplay/tracks/FairEveTracks.cxx
DanielWielanek: When I call it second time then it should return the old group. All "similar" particles should belong to the same group (e.g. pions)
Bez nazwy
).

DanielWielanek: For no I called this "FindTrackGroup".

dennisklein: I have more questions:
1. Does a `FairEveTracks` object only ever store "track groups" (`TEveTrackList` (derived) objects) on its first level of children?
   a. If yes, have you considered implementing `FindTrackGroup()` with [`TEveElement::FindChild()`](https://root.cern/doc/master/classTEveElement.html#aee28f791af457d9397cb739b3355fbb1)?
   b. If no, and we really have to distinguish "track groups" from other kinds of children, then the follow-up question would be: Why do you not use a map container for caching the access to the "track groups" by name?
2. The `SwapTracks()` naming confused me, usually it means that objects are replaced which is not the case here.
   a. I suggest renaming it to something like `ToggleTracks()`, `ToggleTracksVisibility()` or even `ToggleTrackGroups()` if it is meant to operate on the "track group" level only.
   b. Your current implementation of `SwapTracks()` seems to toggle visibility on each individual track within a "track group". Is this intended? If not, would setting `SetRnrSelfChildren()` on the "track group" level work, too?

DanielWielanek: 1. Fixed
2. Replaced by ToogleTracks
This is designed to work at track level - toggling group can be done by clicking, in case of tracks - it's not so easy because you have few hundreds of them.
However toggle groups is not a bad idea - I added it to FairGeoTracks. I also removed virtual label from FairEveTracks::ToggleTracks - if we operate only on tracks there is no point to make this method virtual.

eventdisplay/tracks/FairEveTracks.h
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * base evelement that represents tracks
+ */
+class FairEveTracks
+    : public TEveElement
+    , public TNamed
+{
+    FairEventManager *fEventManager;
+    std::unique_ptr<TObjArray> fEveTrList;
```
dennisklein, line 39:
```suggestion
```

Do we still need this list at all?

See for example:

```diff
diff --git a/eventdisplay/tracks/FairEveTracks.cxx b/eventdisplay/tracks/FairEveTracks.cxx
index c51dd7f41..2902014c1 100644
--- a/eventdisplay/tracks/FairEveTracks.cxx
+++ b/eventdisplay/tracks/FairEveTracks.cxx
@@ -18,6 +18,7 @@
 #include "FairEveRecoTrackList.h"
 #include "FairEveTrack.h"

+#include <algorithm>
 #include <TEveElement.h>
 #include <TEveManager.h>
 #include <TEveSelection.h>
@@ -25,7 +26,6 @@

 FairEveTracks::FairEveTracks(Bool_t acceptCompound)
     : fEventManager(nullptr)
-    , fEveTrList(nullptr)
     , fPt{0, 10}
     , fEta{-10, 10}
     , fEnergy{0, 10}
@@ -33,37 +33,24 @@ FairEveTracks::FairEveTracks(Bool_t acceptCompound)
     , fUseEta(kFALSE)
     , fUseEnergy(kFALSE)
     , fAcceptCompound(acceptCompound)
-{
-    fEveTrList = std::unique_ptr<TObjArray>(new TObjArray());
-}
+{}

 void FairEveTracks::ResetGroup()
 {
-    TEveElement *el = dynamic_cast<TEveElement *>(this);
-    if (el == nullptr)
-        return;
-    for (Int_t iTrackGroup = 0; iTrackGroup < fEveTrList->GetEntriesFast(); iTrackGroup++) {
-        TEveTrackList *ele = static_cast<TEveTrackList *>(fEveTrList->At(iTrackGroup));
-        gEve->RemoveElement(ele, el);
-    }
-    fEveTrList->Clear();
+    RemoveElements();
 }

 void FairEveTracks::ToggleTracks()
 {
-    for (int iTrackGroup = 0; iTrackGroup < fEveTrList->GetEntriesFast(); iTrackGroup++) {
-        TEveTrackList *trackList = (TEveTrackList *)fEveTrList->UncheckedAt(iTrackGroup);
-        for (TEveElement::List_i eveTrackIterator = trackList->BeginChildren();
-             eveTrackIterator != trackList->EndChildren();
-             ++eveTrackIterator) {
-            FairEveTrack *track = dynamic_cast<FairEveTrack *>((*eveTrackIterator));
+    std::for_each(BeginChildren(), EndChildren(), [](TEveElement* trackGroup) {
+        std::for_each(trackGroup->BeginChildren(), trackGroup->EndChildren(), [](TEveElement* track) {
             if (track->GetRnrSelf()) {
                 track->SetRnrSelfChildren(kFALSE, kFALSE);
             } else {
                 track->SetRnrSelfChildren(kTRUE, kTRUE);
             }
-        }
-    }
+        });
+    });
     gEve->Redraw3D(kFALSE);
 }

@@ -77,32 +64,28 @@ FairEveTracks::~FairEveTracks() {}

 TEveTrackList *FairEveTracks::FindTrackGroup(TString groupName, Color_t color)
 {
-    TEveTrackList *TrackGroup = nullptr;
-    TrackGroup = static_cast<TEveTrackList *>(FindChild(groupName));
-    if (TrackGroup == nullptr) {
-        TEveTrackPropagator *TrPropagator = new TEveTrackPropagator();
+    auto trackGroup(static_cast<TEveTrackList *>(FindChild(groupName)));
+    if (trackGroup == nullptr) {
         if (fAcceptCompound) {
-            TrackGroup = new FairEveRecoTrackList(groupName, TrPropagator);
+            trackGroup = new FairEveRecoTrackList(groupName, new TEveTrackPropagator());
         } else {
-            TrackGroup = new TEveTrackList(groupName, TrPropagator);
+            trackGroup = new TEveTrackList(groupName, new TEveTrackPropagator());
         }
-        TrackGroup->SetMainColor(color);
-        fEveTrList->Add(TrackGroup);
-        AddElement(TrackGroup);
-        TrackGroup->SetRnrLine(kTRUE);
+        AddElement(trackGroup); // takes ownership!
+        trackGroup->SetMainColor(color);
+        trackGroup->SetRnrLine(kTRUE);
     }
-    return TrackGroup;
+    return trackGroup;
 }

 void FairEveTracks::ToggleGroups()
 {
-    for (int iTrackGroup = 0; iTrackGroup < fEveTrList->GetEntriesFast(); iTrackGroup++) {
-        TEveTrackList *trackGroup = (TEveTrackList *)fEveTrList->UncheckedAt(iTrackGroup);
+    std::for_each(BeginChildren(), EndChildren(), [](TEveElement* trackGroup) {
         if (trackGroup->GetRnrSelf()) {
             trackGroup->SetRnrSelfChildren(kFALSE, kFALSE);
         } else {
             trackGroup->SetRnrSelfChildren(kTRUE, kTRUE);
         }
-    }
+    });
     gEve->Redraw3D(kFALSE);
 }
diff --git a/eventdisplay/tracks/FairEveTracks.h b/eventdisplay/tracks/FairEveTracks.h
index 4385b370f..19c921ae1 100644
--- a/eventdisplay/tracks/FairEveTracks.h
+++ b/eventdisplay/tracks/FairEveTracks.h
@@ -36,7 +36,6 @@ class FairEveTracks
     , public TNamed
 {
     FairEventManager *fEventManager;
-    std::unique_ptr<TObjArray> fEveTrList;
     Double_t fPt[2];
     Double_t fEta[2];
     Double_t fEnergy[2];
@@ -52,7 +51,6 @@ class FairEveTracks
     Double_t GetEtaMax() const { return fEta[1]; }
     Double_t GetEnergyMin() const { return fEnergy[0]; };
     Double_t GetEnergyMax() const { return fEnergy[1]; };
-    ;
     Bool_t UsePtCut() const { return fUsePt; };
     Bool_t UseEtaCut() const { return fUseEta; };
     Bool_t UseEnergyCut() const { return fUseEnergy; };
@@ -61,7 +59,6 @@ class FairEveTracks
      */
     void ResetGroup();
     FairEventManager *GetEventManager() const { return fEventManager; };
-    TObjArray *GetTracksList() const { return fEveTrList.get(); };
     /**
      *
      * @param groupName - name of the track group
diff --git a/examples/advanced/propagator/src/FairEveRecoTracksExample.cxx b/examples/advanced/propagator/src/FairEveRecoTracksExample.cxx
index 3f61c4b52..12962b984 100644
--- a/examples/advanced/propagator/src/FairEveRecoTracksExample.cxx
+++ b/examples/advanced/propagator/src/FairEveRecoTracksExample.cxx
@@ -210,20 +210,4 @@ void FairEveRecoTracksExample::SetDrawMC(Bool_t draw)
     }
 }

-void FairEveRecoTracksExample::ToggleTracks()
-{
-    for (int i = 0; i < GetTracksList()->GetEntriesFast(); i++) {
-        TEveTrackList *eveList = (TEveTrackList *)GetTracksList()->UncheckedAt(i);
-        for (TEveElement::List_i iter = eveList->BeginChildren(); iter != eveList->EndChildren(); ++iter) {
-            FairEveRecoTrack *el = dynamic_cast<FairEveRecoTrack *>((*iter));
-            if (el->GetRnrSelf()) {
-                el->SetRnrSelfChildren(kFALSE, kFALSE);
-            } else {
-                el->SetRnrSelfChildren(kTRUE, kTRUE);
-            }
-        }
-    }
-    gEve->Redraw3D(kFALSE);
-}
-
 FairEveRecoTracksExample::~FairEveRecoTracksExample() {}
diff --git a/examples/advanced/propagator/src/FairEveRecoTracksExample.h b/examples/advanced/propagator/src/FairEveRecoTracksExample.h
index 376f07dde..b14476350 100644
--- a/examples/advanced/propagator/src/FairEveRecoTracksExample.h
+++ b/examples/advanced/propagator/src/FairEveRecoTracksExample.h
@@ -55,7 +55,6 @@ class FairEveRecoTracksExample : public FairEveTracks
         fShowPrimary = prim;
         fShowSecondary = sec;
     }
-    virtual void ToggleTracks();
     virtual InitStatus Init();
     virtual ~FairEveRecoTracksExample();
     ClassDef(FairEveRecoTracksExample, 0)
```

dennisklein (changes requested):
Please review your #includes and clean them up please (only in your changes).
* There are tools to help with this, e.g.: https://include-what-you-use.org/ or https://clang.llvm.org/extra/clang-include-fixer.html
* Also check for opportunities to use a forward declarations instead of full includes.

eventdisplay/FairEventManager.h
```diff
@@ -48,22 +49,73 @@ class FairEventManager : public TEveEventManager
...
+    virtual Int_t GetCurrentPDG() { return 0; }
+    virtual void SetMaxEnergy(Float_t max) {}
+    virtual void SetMinEnergy(Float_t min) {}
+    virtual void SetEvtMaxEnergy(Float_t max) {}
+    virtual void SetEvtMinEnergy(Float_t min) {}
+    virtual Float_t GetEvtMaxEnergy() { return 1E+9; }
+    virtual Float_t GetEvtMinEnergy() { return 0; }
+    virtual Float_t GetMaxEnergy() { return 1E+9; }
+    virtual Float_t GetMinEnergy() { return 0; }
+    void GetWorldSize(Double_t &x, Double_t &y, Double_t &z) const;
```
dennisklein, line 111: Do not introduce in/out parameters for out-only parameters in 2020 (see [F.20: For “out” output values, prefer return values to output parameters](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f20-for-out-output-values-prefer-return-values-to-output-parameters) and [F.21: To return multiple “out” values, prefer returning a struct or tuple](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f21-to-return-multiple-out-values-prefer-returning-a-struct-or-tuple))

```suggestion
    TVector3 GetWorldSize() const;
```
or if you wanted to support any world shape without changing the API again in the future, consider:
```suggestion
    TGeoShape* GetWorld() const;
```

eventdisplay/tracks/FairEveTracks.h
DanielWielanek: I fixed headers with eclipse-cdt. I hope now it is fine. I also commented out some objects that should not be streamed.

eventdisplay/FairEventManagerEditor.cxx
```diff
@@ -111,77 +124,127 @@ void FairEventManagerEditor::Init()
...
+                                            6,
+                                            -1,
+                                            TGNumberFormat::kNESInteger,
+                                            TGNumberFormat::kNEANonNegative,
+                                            TGNumberFormat::kNELLimitMinMax,
+                                            0,
+                                            100);
+    fGlobalTransparency->SetIntNumber(80);
+    transparency_frame->AddFrame(fGlobalTransparency, new TGLayoutHints(kLHintsRight, 1, 1, 1, 1));
+    fGlobalTransparency->Connect("ValueSet(Long_t)", "FairEventManagerEditor", this, "SwitchTransparency(Bool_t)");
```
dennisklein, line 153: Something with this observer goes wrong. The `Long_t` argument seems to not contain the transparency number, when clicking the increase button.

For a fix I propose something like the following: https://github.com/dennisklein/FairRoot/commit/3f13e209257facf4f678f86eed70d71bd28b11aa

DanielWielanek: This ValueSet is only to be sure that function will be called when user change the value (fCurrentEvent works in similar way). According to documentation it contains information about step (or something like that, but not about the stored value).
Also in ROOT tutorials (https://root.cern/doc/master/numberEntry_8C.html ) GetValue(Long_t) is not used to get value from button. I don't think that creation of new class only for a single button (and basically no extra functionality) is a good idea.
Btw. what is the point of using unique_pointer and call release method few lines after?

dennisklein: > I don't think that creation of new class only for a single button (and basically no extra functionality) is a good idea.

Why?

> (and basically no extra functionality)

My example has quite some of different functionality. It creates a new signal that consolidates the two signal from the check box and from the number entry. This signal only fires if there is a meaningful change and only presents a single int parameter (instead of bool + int). This signal is also wired up directly with the manager and not in-directed throw another function in the editor.

> Btw. what is the point of using unique_pointer and call release method few lines after?

* Exception-safety: If your programs throws an exception in these few lines, the object will not leak. (stack unwinds and dtor of the unique ptr will delete the object)
* Readability: The code shows clearly and unambiguously who owns the objects.

---

Anyways, if I check your box and then click the arrow up button on the number entry widget, it does not behave as expected. It behaves as if I un-check the box.

TobiasStockmanns:
I have tested the eventdisplay for PANDA and it works with minor modifications to our macros. Nevertheless I see some smaller points for improvements:
- For the _animations type_ PANDA would need an entry "time step". I can provide the necessary code for this change.
- It would be nice to have the possibility to set a prefix for the output file of a screenshot or an animation
- I have not tested the new feature FairEveRecoTracks. There is an example file mentioned in the `CMakeLists.txt`of `examples\advanced\propagator\src` but the file is not there. Some documentation would be nice.

eventdisplay/FairEventManagerEditor.cxx
DanielWielanek: Now I think it's fixed - however I slightly modified your class. I also moved all button-like classes to eventdisplay/gui directory (I suppose we will get more of them in the future).
Thank you for noticing this bug.

DanielWielanek:
@TobiasStockmanns  what do you mean by "an entry time step"? Would you like to not paint tracks before T_min?
Concering screeshots - now it's possible to set the name of the screenshot file (however due to limitation to ROOT it's doesn't work perfectly).
Concering prefix of animation - do you mean stoning files as e.g. custom_prefix_1.png for animations?

TobiasStockmanns:
@DanielWielanek With "time step" I mean that the experiment time t0 is changed in e.g. 1 ns steps from 0 to 1000.
The different sub-detectors of PANDA have different time resolutions so that for an assumed t0 time, different amount of events overlap in the different detectors. The `FairBoxSetDraw` has the functionality to draw hits based on their time stamps. You can set a time window before and after the measured time stamp for which hits of one sub-detector are still active.
The only thing you have to do in addition is to give an assumed event time t0 to the `FairRootManager` and to activate `FairRunAna::RunWithTimeStamps()`. This would be the change I would like to have.

For the animation prefix it is what you said: custom_prefix_1.png
I have an additional question. Why do you not store directly an animated gif from root?

Is it possible to do an in-event animation to see the propagation of the MC tracks in the detector?

DanielWielanek:
I don't use gifs because:
- originally people from Dubna wanted the event by event animation where each file is stored as separate file, therefore I created a simple code that use the same function for animations in "event by event" and "time" mode, actually I'm not sure if there is option of saving  gif directly in a simple way (at least with this shape of event display)
- I'm also worry about size of the gif that stores very long animation - for 1 minute video you need 1.5k frames
- I would also use this animation to check some tracking procedures and browsing many files instead of gif is much more simple (at leas to me)
Concerning propagation of MC track - it's possible, FairEveTracks::Repaint in principle is called for each event/animation step. However for calculation of FairMCTrack position I used FairRKPropagator that doesn't have option to calculate the position of particle at given time. There is also problem with unstable particles - there is no information in FairMCTrack where particle decays - and this is needed to correctly draw animation. Actually in current example all particles are drawn up to some Z - even if they decayed.
Concerning timestamp stuff - I will try to implement this.

eventdisplay/FairEventManagerEditor.cxx
```diff
@@ -111,77 +121,114 @@ void FairEventManagerEditor::Init()
...
-}
 
-void FairEventManagerEditor::MaxEnergy() { fManager->SetMaxEnergy(fMaxEnergy->GetValue()); }
+    //=============== graphics =============================
+    TGVerticalFrame* scene_conf = CreateEditorTabSubFrame("Graphics");
+    TGHorizontalFrame* transparency_frame = new TGHorizontalFrame(scene_conf);
 
-void FairEventManagerEditor::MinEnergy() { fManager->SetMinEnergy(fMinEnergy->GetValue()); }
+    auto transparency(
+        std::unique_ptr<FairEveTransparencyControl>(new FairEveTransparencyControl(scene_conf, "Global transparency")));
```
dennisklein, line 137:
```suggestion
    std::unique_ptr<FairEveTransparencyControl> transparency(new FairEveTransparencyControl(scene_conf, "Global transparency"));
```

or

```suggestion
    auto transparency(fair::make_unique<FairEveTransparencyControl>(scene_conf, "Global transparency"));
```

For the latter please cherry-pick https://github.com/dennisklein/FairRoot/commit/d7bfe17446f87483d287857c761af3861cc8a58a and add `#include "FairSmartPointer.h"`. This is just a backport of a helper that was added to the standard library in C++14 ([std::make_unique](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique)). See also [C.150: Use make_unique() to construct objects owned by unique_ptrs](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c150-use-make_unique-to-construct-objects-owned-by-unique_ptrs).

I am sorry for being pedantic about this, but this is framework code and I want to avoid people copy&pasting bad examples (we already have plenty of anti-patterns that get copied around like crazy). We should either use `unique_ptr`, `auto` etc the right way or not at all.

dennisklein:
For creating a video from a folder with pictures, one can use e.g. `ffmpeg` - The following command will create a 15 sec video with 2 events per seconds (output video has 30fps):

`ffmpeg -r 2 -i 'event_%d.png' -start_number 0 -r 30 out.mp4`: [out.mp4](https://sf.gsi.de/f/c46cdf98ccff4ffbb5bd/?dl=1)

If one wants to create a video and control exact timing of each image, one can e.g. do this:

`ffmpeg -f concat -i input.txt -r 30 out2.mp4`: [out2.mp4](https://sf.gsi.de/f/96568981c0264de18b8c/?dl=1)

with `input.txt` containing:
```
file 'event_0.png'
duration 0.1
file 'event_1.png'
duration 0.1
file 'event_2.png'
duration 0.1
file 'event_3.png'
duration 0.3
file 'event_4.png'
duration 0.3
file 'event_5.png'
duration 0.3
file 'event_6.png'
duration 0.1
file 'event_7.png'
duration 0.1
file 'event_8.png'
duration 0.5
file 'event_9.png'
duration 0.5
file 'event_10.png'
duration 0.5
file 'event_11.png'
duration 0.5
file 'event_12.png'
duration 0.1
file 'event_13.png'
duration 0.1
file 'event_14.png'
duration 0.1
file 'event_15.png'
duration 0.1
file 'event_16.png'
duration 0.1
```

So for timebased animations we would need to export some form of timestamps for each image from which one could construct such a scaled input sequence as shown above. For a start we could just put some suitable example commands into some README file?

TobiasStockmanns:
I have made the changes to run the "time step" animation. Shall I merge them into this pull request or how should we do that?

MohammadAlTurany:
@TobiasStockmanns: best would be a PR to @DanielWielanek  branch

TobiasStockmanns:
I created a new merge request for @DanielWielanek to improve the time-slice animation for MC points and hits. This works well for event based data. Time based data is more complicated and should come at a later stage.

dennisklein (approved):

MohammadAlTurany:
Merged Manually

Issue #1010, opened 2020-09-11
[WARN] FairGenericVMCConfig::Setup() Using gConfig.C macro DEPRACATED.


dennisklein:
Our own examples should not use deprecated functionality.

PR #1011, opened 2020-09-15
Remove the depracated warning when using gconfig.C

karabowi:
Use the FairVMCConfig in the simulation macros.

Addresses issue #1010.

---

Checklist:


dennisklein:
Please rebase on current `dev`, thx!

dennisklein (approved):
I took the liberty and removed the commits that added and removed stuff (with the netto result of doing nothing). Also rebased to `dev` and renamed one commit from `Add missing library` to `Tune MQ pixelDetector example output`.

examples/MQ/pixelDetector/macros/run_sim.C
```diff
@@ -49,6 +49,7 @@ void run_sim(Int_t nEvents = 10, TString mcEngine = "TGeant3", Int_t fileId = 0,
     // -----   Create simulation run   ----------------------------------------
     FairRunSim* run = new FairRunSim();
     run->SetName(mcEngine);   // Transport engine
+    run->SetSimulationConfig(new FairVMCConfig());
```
dennisklein, line 52: The `FairVMCConfig` object appears to be leaking (not going to repeat this comment for the other places in this PR).

examples/MQ/pixelDetector/src/CMakeLists.txt
```diff
@@ -71,6 +72,7 @@ target_link_libraries(${target} PUBLIC
   FairRoot::Gen # FairBoxGenerator
   FairRoot::GeoBase
   FairRoot::ParBase # FairParGenericSet
+  FairRoot::ExGConfig
```
dennisklein, line 75: The `FairRoot::ExGConfig` target is built conditionally https://github.com/FairRootGroup/FairRoot/blob/453ab6c945bb7438eccb0061888cd9353841d960/examples/CMakeLists.txt#L15-L17
because `FairMCConfigurator` is built conditionally https://github.com/FairRootGroup/FairRoot/blob/453ab6c945bb7438eccb0061888cd9353841d960/fairtools/CMakeLists.txt#L11-L13
which leads to compile errors if those dependencies are not built. Please revisit the code and add the necessary conditions to the `add_subdirectory()` CMake calls for all the example that have a dependency on `FairRoot::ExGConfig` (or handle it differently if you wish).

dennisklein (commented):
I reverted my changes, because the PR turned out not quite ready for merging.

* See inline comments
* Please clean the commits from changes that are introduced and again removed within this PR.
* Two following commits have the same commit message, but one of them does not match the changes, please reword more appropriately.

(once done, press the button "Ready for review" at the bottom of this page right above the CI checks list)

ChristianTackeGSI:
The centos 7 CI tests fail, because they don't have yaml-cpp installed.

As Dennis already pointed out, we should get clear on how this should be handled?

examples/MQ/pixelDetector/src/CMakeLists.txt
ChristianTackeGSI: Okay, we decided to require yaml-cpp. So this should be resolved now.

ChristianTackeGSI:
Next step: Memory management.

If SetSimulationConfig gets an owning pointer (as all the proposed examples do), FairRunSim should clean up that memory.

My proposal would be to perform the same steps as for Sink/Source:
Use a `unique_ptr` inside FairRunSim, provide a new (additional) `SetSimulationConfig` API that also takes a `unique_ptr`, and rewrite the examples accordingly.
See for examples on how this could look like: #1194, #1200

PR #1013, opened 2020-09-22
Project template error: Geant4 is not found by CMake

MohammadAlTurany:
On case sensitive machines Geant4 is not found when using the project templates, this is because the name in CMakelist is written all capital, which works on Mac OS but not on linux or case sensitive Mac OS.


templates/project_root_containers/CMakeLists.txt
```diff
@@ -100,8 +100,8 @@ find_package2(PUBLIC FairLogger  VERSION 1.2.0 REQUIRED)
 find_package2(PUBLIC Pythia6)
 find_package2(PUBLIC Pythia8)
 find_package2(PUBLIC Geant3)
-find_package2(PUBLIC GEANT4)
-find_package2(PUBLIC GEANT4DATA)
+find_package2(PUBLIC Geant4)
+find_package2(PUBLIC Geant4Data)
```
dennisklein, line 104:
```suggestion
```

This is obsolete.

dennisklein (approved):

PR #1014, opened 2020-09-24
Use IWYU to clean the includes 

MohammadAlTurany:
Clean the includes using IWYU


base/MQ/tasks/FairMQProcessorTask.cxx
```diff
@@ -14,6 +14,9 @@
 
 #include "FairMQProcessorTask.h"
 
+#include <FairMQMessage.h>  // for FairMQMessage
+#include <type_traits>      // for move
```
rbx, line 18: Not sure why IWYU suggests `<type_traits>` here, move is defined in `<utility>` according to cppreference:
https://en.cppreference.com/w/cpp/utility/move

MohammadAlTurany: Corrected, thanks!

PR #1025, opened 2020-11-13
Fix for the missing magnetic field in the event display

karabowi:
When no magnetic field available from FairRunAna, the event
display crashes when drawing events.
Small fix in FairEveMCTracks fixes the issue - it will not exptrapolate tracks when no magnetic field.

Introduce the FairFieldCreator in the Tutorial3 advanced example,
which allows FairRunAna to retrieve magnetic field from the parameter file.

---

Checklist:

### DIFF
--- a/examples/common/eventdisplay/FairEveMCTracks.cxx
+++ b/examples/common/eventdisplay/FairEveMCTracks.cxx
@@ -13,24 +13,26 @@
  *		E-mail: daniel.wielanek@gmail.com
  *		Warsaw University of Technology, Faculty of Physics
  */
- #include "FairEveMCTracks.h"
- #include <TClonesArray.h>       // for TClonesArray
- #include <TDatabasePDG.h>       // for TDatabasePDG
- #include <TEveManager.h>        // for TEveManager, gEve
- #include <TEveTrack.h>          // for TEveTrackList
- #include <TLorentzVector.h>     // for TLorentzVector
- #include <TParticle.h>          // for TParticle
- #include <TParticlePDG.h>       // for TParticlePDG
- #include <TString.h>            // for Form
- #include <TVector3.h>           // for TVector3
- #include <fairlogger/Logger.h>  // for LOG
- #include "FairEveTrack.h"       // for FairEveTrack
- #include "FairEventManager.h"   // for FairEventManager
- #include "FairField.h"          // for FairField
- #include "FairMCTrack.h"        // for FairMCTrack
- #include "FairRKPropagator.h"   // for FairRKPropagator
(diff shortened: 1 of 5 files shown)


dennisklein:
I am probably blind, but I cannot figure out how the new class `FairConstFieldCreator` is being used?

examples/common/eventdisplay/FairEveMCTracks.cxx
```diff
@@ -151,9 +156,10 @@ InitStatus FairEveMCTracks::Init()
     FairField *field = ana->GetField();
     if (field == nullptr) {
         LOG(ERROR) << "Lack of magnetic field map!";
-        field = new FairField();
+        //        field = new FairField();
```
dennisklein, line 159: Is this intended? If yes, let's remove it, otherwise a commented out line without any further comments will hurt readability of the code.

dennisklein (changes requested):

examples/common/eventdisplay/FairEveMCTracks.cxx
DanielWielanek: One comment to @karabowi, are we sure that now magnetic field is present in all examples, fRK is used by  FairEveMCTracks::DrawTrack.

karabowi:
I have a problem with all the FairRoot examples' eventDisplay.C.
To reproduce:
1. go to examples/advanced/Tutorial3/macro/
2. run root -l -q run_sim.C ,
3. run root -l eventDisplay.C ,
4. in the top left white panel, click on "FairEventManager",
5. click the "Info" submenu below,
6. push the top arrow in the "Current event".
This causes my console to explode with 
`
[WARN] FairField::GetBx Should be implemented in User class
[WARN] FairField::GetBy Should be implemented in User class
[WARN] FairField::GetBz Should be implemented in User class
`
Unstoppable by ctrl-c, have to kill the current root session.

**Reason**: FairRunAna does not find magnetic field,
while the FairEveMCTracks needs it to propagate tracks. 
Creating dummy field in the Init() does not help,
as it hasn't got implementations of the above functions.
**Solution in the FairEveMCTracks**: Do not create the field, and do create the 
FairRKPropagator* fRK _only_ if the field is obtained from FairRunAna.

**Question**: How does the FairRunAna not find the magnetic field,
even though FairRunSim had it, and the magnetic field is in the parameter file?
**Answer**: To translate the field parameters into magnetic field the
FairFieldFactory needs a specific instance of FairFieldCreator.
**Problem**: Neither of our examples was showing that.
**Solution**: Implement the FairConstFieldCreator in Tutorial3.

dennisklein (approved):

PR #1028, opened 2020-12-03
Add initial codemeta.json

ChristianTackeGSI:
codemeta.json is becoming a standard format for describing software. Let's start to add it.

This is also starting to get relevant in the context of ESCAPE.

I created the current codemeta.json to match the entry on zenodo as closely as possible. That means, that any issues with the zenodo entry are also issues with the codemeta data.

Helpful:
* There is a codemeta generator/viewer: https://codemeta.github.io/codemeta-generator/
  You can even paste the suggested codemeta.json there and view it.

---

Checklist:

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,7 +1,7 @@
 Al-Turany, Mohammad [https://orcid.org/0000-0002-8071-4497]
 Karabowicz, Radoslaw
 Klein, Dennis [https://orcid.org/0000-0003-3787-1910]
-Kresan, Dmytro
+Kresan, Dmytro [https://orcid.org/0000-0002-7537-2875]
 Rybalchenko, Alexey
 Tacke, Christian [https://orcid.org/0000-0002-5321-8404]
 Uhlig, Florian
--- a/codemeta.json
+++ b/codemeta.json
@@ -1,0 +1,51 @@
+{
+    "@context": "https://doi.org/10.5063/schema/codemeta-2.0",
+    "@type": "SoftwareSourceCode",
+    "name": "FairRoot",
+    "description": "<p>A simulation, reconstruction and analysis framework that is based on the ROOT system. The user can create simulated data and/or perform analysis with the same framework. Geant3 and Geant4 transport engines are supported, however the user code that creates simulated data do not depend on a particular monte carlo engine. The framework delivers base classes which enable the users to construct their detectors and /or analysis tasks in a simple way, it also delivers some general functionality like track visualization. Moreover an interface for reading magnetic field maps is also implemented.</p>",
+    "license": "https://spdx.org/licenses/LGPL-3.0-only",
+    "version": "v18.2.1",
+    "datePublished": "2012-05-25",
+    "keywords": [
+        "geant4",
+        "c-plus-plus",
+        "cmake",
+        "reconstruction",
+        "vmc",
+        "modular",
+        "analysis",
+        "simulation"
+    ],
+    "contributor": [
+        {
+            "@type": "Person",
+            "familyName": "Al-Turany",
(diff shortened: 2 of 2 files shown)


ChristianTackeGSI:
If you need more details, or have any other questions, you might consider waiting until Monday's Group Meeting. I have allocated a short slot there.

dennisklein:
Let's continue here once FairRootGroup/FairMQ#324 is merged and update accordingly.

dennisklein:
I suggest to follow up on this by:
* [ ] Rebase on current `dev` branch
* [ ] Define the authoritative place for the version number
  a. Keep it in `CMakeLists.txt` or
  b. CMake queries version number from `codemeta.json`
     * Add the [CMake code written by @ChristianTackeGSI](https://github.com/FairRootGroup/FairMQ/pull/324#issuecomment-744620316) that queries the version number from `codemeta.json`
 * [ ] Adopt [`codemeta_update.py`](https://github.com/FairRootGroup/FairMQ/blob/master/codemeta_update.py) from FairMQ repo (in case of b) probably can be stripped down to the author logic only)
 * [ ] Add a release checklist for maintainers so we do not forget to press this from time to time
 * [ ] CI consistency check welcome too

Regarding the authoritative place for the version number, I personally have no preference whether to go with options a or b. @fuhlig1 @ChristianTackeGSI What is your preference?

ChristianTackeGSI:
Regarding authoritative storage of the version number, I have no strong opinion.

If we adopt `codemeta_update.py` anyway, using CMakeLists.txt as the authoritative source might seem simpler (and doesn't require a very recent CMake version).

The other way round might have its positive sides as well.
I personally would for example move the author / maintainer info fully into codemeta.json and drop the simple text files. Putting the version there as well would make codemeta.json fully authoritative and wouldn't require it to be updated at all.

dennisklein:
> I personally would for example move the author / maintainer info fully into codemeta.json and drop the simple text files. 

For the authors I dislike that the json file will hide this info much more IMHO. I find the authors deserve the visibility they get by those capitalized-and-easy-to-find, dead-simple-to-read `CONTRIBUTORS`/`AUTHORS` files.

ChristianTackeGSI:
First: The "personally" part was exactly that, highly personal. I can see all good reasons to consider json a machine format and not a format for humans! That's why we have `codemeta_update.py`.

My "problem" with the dead-simple-to-read files is the ad-hoc format. How would you store an [ORCID](https://orcid.org/) in it, just to name an example.

rbx:
I also prefer CONTRIBUTORS and AUTHORS files. It seems much more common and is something I would look for first if I wanted to find the authors.

dennisklein:
> My "problem" with the dead-simple-to-read files is the ad-hoc format. How would you store an [ORCID](https://orcid.org/) in it, just to name an example.

```
Jane Doe <https://orcid.org/0000-0000-0000-0000>
```
?

dennisklein:
In the last group meeting on 20.9. we (@karabowi, @TobiasStockmanns, @rbx, @ChristianTackeGSI, and myself) agreed to keep the files `AUTHORS` and `CONTRIBUTORS` with a simple one line human-readable format as today. Whether those files are generated from the author lists in the `codemeta.json` file or the other way around is up to the implementer of the `codemeta.json` facility (probably @ChristianTackeGSI :wink:).

ChristianTackeGSI:
@kresan

I added your ORCID to AUTHORS (in its own commit).

I suggest, that we merge this PR mostly as is, so that we have an initial codemeta to start out. I have moved the big TODO list into #1351 so that we can incrementally get forward with this.

kresan (approved):
Good as initial version. We can merge it.

PR #1029, opened 2020-12-10
Time based event animation in event display

TobiasStockmanns:
Modification of the event display to handle MC data in a time based animation.

- New class to collect event time information from digi file
- Data handler to simplify the handling of data from FairMCPoints, FairHits (with relative time information) and FairHits with absolute time inforamtion
- Possibility to animate overlapping events
- Animate tracks with a tail of settable length

This commit breaks the backward compatibility for classes derived from FairBoxSetDraw / FairPointSetDraw. The change is the switch to `FairDataSourceI`, which requires the usage of `FairDataSourceI::GetNData()` instead of `TClonesArray::GetEntries(Fast)()` and `TObject* FairDataSourceI::GetData(index)`instead of `TObject* TClonesArray::At(index)`. 

---

Checklist:


base/steer/FairRootManager.cxx
```diff
@@ -455,11 +456,12 @@ Int_t FairRootManager::ReadEvent(Int_t i)
     fCurrentEntryNo = i;
 
     Int_t readEventResult = fSource->ReadEvent(i);
+    Double_t eventTime = ((FairFileSource*)fSource)->GetEventTime();
 
     fSource->FillEventHeader(fEventHeader);
     fCurrentTime = fEventHeader->GetEventTime();
 
-    LOG(debug) << "--Event number --- " << fCurrentEntryNo << " with time ---- " << fCurrentTime;
+    std::cout << "--Event number --- " << fCurrentEntryNo << " with time ---- " << fCurrentTime << " eventTime " << eventTime << std::endl;
```
rbx, line 464: Why not use the logger here? This will always print the lines regardless of the configured log severity.

TobiasStockmanns: I will switch it back to the logger. The only reason this was done was because of debugging reasons.

eventdisplay/datasource/FairTimebasedSource.h
```diff
@@ -0,0 +1,42 @@
...
+    virtual ~FairTimebasedSource();
+
+    virtual void RetrieveData(double time);
+    virtual int GetNData() { return fCArray->GetEntries(); };
+    virtual TObject* GetData(int index) { return fCArray->At(index); }
+    virtual double GetTime(int index) { return static_cast<FairTimeStamp*>(fCArray->At(index))->GetTimeStamp(); }
+    virtual void Reset(){};
+
+  private:
+    BinaryFunctor* fStartFunctor{new StopTime()};   //!
```
rbx, line 34: Who is deleting these objects? Having a brief look at FairRootManager and FairTSBufferFunctional, I don't see how these are deleted. @MohammadAlTurany @fuhlig1

base/steer/FairRootManager.cxx
rbx: ok, thanks. I think eventdisplay/tracks/FairGeoTrackHandler.cxx:19 has the same issue.

eventdisplay/datasource/FairTimebasedSource.h
TobiasStockmanns: I guess you are speaking about the deletion of the `FairTSBufferFunctional` and not of the fStart/StopFunctors. Looking at the code I agree with you, that I have not found a delete for the FairTSBufferFunctionals nor for the containers used inside.
I assume that the root garbage collection takes over this task but it should be done in a more propper way.

base/steer/FairRootManager.cxx
TobiasStockmanns: I changed now all std::cout in folder /eventdisplay to fairlogger output :-)

eventdisplay/datasource/FairTimebasedSource.h
rbx: now i see the delete calls at eventdisplay/datasource/FairTimebasedSource.cxx:20, i guess i overlooked them before.

rbx: But then the ownership is completely in the hands of this class and they can be made non-pointer members, and passed to ptr-requiring functions via `&`.

PR #1038, opened 2021-02-25
Drop check for "-std=*" in CXX_FLAGS

ChristianTackeGSI:
The old way of passing the wanted C++ language level into FairRoot was to set (CMAKE_)CXX_FLAGS with a "-std=c++11" flag.

In a more modern cmake world, one uses CMAKE_CXX_STANDARD and target_compile_features() to let cmake choose the right language level.
The problem starts here: cmake doesn't parse CMAKE_CXX_FLAGS and so flags from both sources end up in the final compile command.
If one wants to configure a project using the modern methods, one does not want to use the old method.

This commit removes a fatal check, that enforces the user to use the old method to set the language level.  That way the user can choose the old or the modern way of doing it.

To be precise: This does not stop any current use. It only removes some (old) "safety guards". And thus allows more modern uses.

Some more notes:

* @dennisklein and I had a long talk about C++ language level choosing and the way forward with this today.
* We're trying to move to more modern cmake style choosing of the C++ language level / features. See for example: https://github.com/FairRootGroup/FairMQ/pull/333
* Especially in FairSoft (legacy and spack) we want to move to `-DCMAKE_CXX_STANDARD=` for configuring the language level

---

Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,23 +106,6 @@ include(CheckCompiler)
 #Check the compiler and set the compile and link flags
 Check_Compiler()
 
-# Check also if FairSoft has been compiled with C++11 or C++14 support
-# If FairSoft is used the compiler flags provided by fairsoft-config
-# are added to the variable CMAKE_CXX_FLAGS.
-# If alibuild is used the compiler flags are passed on the command line
-# and are also added to CMAKE_CXX_FLAGS.
-# So check if CMAKE_CXX_FLAGS has the compiler flags -std=c++11 or -std=c++14
-String(FIND "${CMAKE_CXX_FLAGS}" "-std=c++11" POS_C++11)
-If(${POS_C++11} EQUAL -1)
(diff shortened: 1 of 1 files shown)


ChristianTackeGSI:
HI @fuhlig1 

Dennis suggested to ping you about this in particular.

To repeat: This is not intended to stop any current use! It's only intended to allow other (more modern) use.

We were wondering, whether any of your use cases might require this check in this case. If so, the question is, whether a warning instead of a fatal error would be okay too. (That way allowing the modern use, just with a warning).

dennisklein:
Merging. @fuhlig1 We will implement your suggestion to take the cxxstd from ROOT in v19.0.0. For now, this PR only removes the check, but does not alter the mechanism which sets the std flag based on `fairsoft-config` output.

ChristianTackeGSI:
Are you going to cherry pick this into 18.4.3?
Should I open a new PR for this?

PR #1042, opened 2021-03-12
Use correct flag for post init configuration

### LINKED ISSUE #1041: CbmRoot simulation does not work with FairRoot v18.6.0
**Describe the bug**
When running a transport simulation in CbmRoot using FairRoot v18.6.0 I always get the following crash

 ```
Calculating cross section tables, see gphysi.dat for more information

 Cross section calculation concluded successfully
[INFO] Monte Carlo Engine Initialisation with: TGeant3TGeo
terminate called after throwing an instance of 'std::bad_function_call'
  what():  bad_function_call
```

This are only the last lines showing the error. There is no problem before. The same CbmRoot source code works without any problems when using FairRoot v18.4.3 

**To Reproduce**
Compile CbmRoot with FairRoot v18.6.0 and start a transport simulation. 

**Expected behavior**
The CbmRoot transport simulation should run also when using FairRoot v18.6.0 

**Additional context**
The problem turned out to be related to the changes which were implemented to allow post init configuration of Geant4. In the respective code it is checked if an external function should be executed. This function has to be added by the user. Unfortunately in the check the variable for pre init configuration is checked. In the case of CbmRoot we use the pre init configuration such that the respective variable is true. So the crash happens because the variable is true but there is no pre init function connected which then ends in the seen bad_function_call.

I will provide a pull request fixing the issue in a minute.

### PR DESCRIPTION
**Pull request by fuhlig1**
Also check if there is really a function bound.

Fixes #1041 .

The fix has also to be ported to the v18.6_patches branch. 
---

Checklist:

### DIFF
--- a/base/steer/FairRunSim.cxx
+++ b/base/steer/FairRunSim.cxx
@@ -296,10 +296,12 @@ void FairRunSim::SetMCConfig()
 
     fApp->InitMC("foo", "bar");
 
-    if (fUseSimSetupFunction) {
+    if (fUseSimSetupPostInitFunction) {
         fSimSetupPostInit();
     } else {
-        fSimulationConfig->SetupPostInit(GetName());
+        if (fSimulationConfig != nullptr) {
+            fSimulationConfig->SetupPostInit(GetName());
+        }
     }
 }
 

### CODE REVIEW

File: base/steer/FairRunSim.cxx [hunk 1]
```diff
@@ -296,10 +296,12 @@ void FairRunSim::SetMCConfig()
 
     fApp->InitMC("foo", "bar");
 
-    if (fUseSimSetupFunction) {
+    if (fUseSimSetupPostInitFunction) {
         fSimSetupPostInit();
     } else {
```
dennisklein, line 301:
```suggestion
    if (fSimSetupPostInit) {
        fSimSetupPostInit();
    } else {
```

It looks to me that the bool flag is even redundant, see https://en.cppreference.com/w/cpp/utility/functional/function/operator_bool.

Also, e7dfa6a3a363d745027cfd6570ec51f2ec0af3a8 added new fields to the `FairRunSim` class, but did not bump the `ClassDef(FairRunSim, 2);` revision?!

@fuhlig1 @karabowi How do you want to proceed?

karabowi: I prefer to leave the flags for the user to set.
This is the common scheme throughout the whole code - the users are used to having a flag to set.

karabowi (approved):

base/steer/FairRunSim.cxx
fuhlig1: I agree with Radek but thanks for pointing out the possibility to avoid the extra flags. I think this is something we should keep in mind.

rbx: Leave the flags if you really want to.
But don't use them in new code. It makes no sense. Use what Dennis suggested.
Ideally mark the flags as deprecated.

dennisklein (changes requested):
```diff
ERROR: format check failed. Suggested changes:
diff --git a/base/steer/FairRunSim.cxx b/base/steer/FairRunSim.cxx
index 25ae76218..b73e3ea40 100644
--- a/base/steer/FairRunSim.cxx
+++ b/base/steer/FairRunSim.cxx
@@ -300,7 +300,7 @@ void FairRunSim::SetMCConfig()
         fSimSetupPostInit();
     } else {
         if (fSimulationConfig != nullptr) {
-           fSimulationConfig->SetupPostInit(GetName());
+            fSimulationConfig->SetupPostInit(GetName());
         }
     }
 }
```

Will merge and backport to v18.6 once the format check passes. (macOS checks are broken at the moment, we can ignore them)

Issue #1044, opened 2021-03-22
FairYamlVMCConfig uses FastSim config only


mgoncerz:
The class that's designed to parse Geant configs in .yaml format (FairYamlVMCConfig) only uses FairFastSimRunConfiguration (lines 186-191) for Geant4 setup, which then creates FairFastSimulation and tries to run whole simulation using fastsim.

I have re-implemented this class for my use case and changed lines:

```
    FairFastSimRunConfiguration* runConfiguration =
        new FairFastSimRunConfiguration(fYamlConfig["Geant4_UserGeometry"].as<std::string>(),
                                        fYamlConfig["Geant4_PhysicsList"].as<std::string>(),
                                        fYamlConfig["Geant4_SpecialProcess"].as<std::string>(),
                                        specialStacking,
                                        mtMode);
```

into:

```
    TG4RunConfiguration* run_configuration;

    if(m_yamlConfig["UseFastSim"] && true == m_yamlConfig["UseFastSim"].as<Bool_t>()) 
        run_configuration = new FairFastSimRunConfiguration(m_yamlConfig["Geant4_UserGeometry"].as<std::string>(),
                                            m_yamlConfig["Geant4_PhysicsList"].as<std::string>(),
                                            m_yamlConfig["Geant4_SpecialProcess"].as<std::string>(),
                                            specialStacking,
                                            mtMode);
    else
        run_configuration = new TG4RunConfiguration(m_yamlConfig["Geant4_UserGeometry"].as<std::string>(),
                                            m_yamlConfig["Geant4_PhysicsList"].as<std::string>(),
                                            m_yamlConfig["Geant4_SpecialProcess"].as<std::string>(),
                                            specialStacking,
                                            mtMode);      
```

I think it's worth it to add this also into the FairRoot version.

karabowi:
Addressed in commit #1105 .

PR #1045, opened 2021-03-23
Add FastSim flag plus setters

karabowi:
Addresses issue #1044.

---

Checklist:

### DIFF
--- a/base/sim/fastsim/FairFastSimRunConfiguration.cxx
+++ b/base/sim/fastsim/FairFastSimRunConfiguration.cxx
@@ -45,5 +45,5 @@ TG4VUserFastSimulation* FairFastSimRunConfiguration::CreateUserFastSimulation()
         LOG(info) << "Going to create FairFastSimulation";
         return new FairFastSimulation();
     }
-    return 0;
+    return nullptr;
 }
--- a/base/sim/fastsim/FairFastSimRunConfiguration.cxx
+++ b/base/sim/fastsim/FairFastSimRunConfiguration.cxx
@@ -22,6 +22,7 @@ FairFastSimRunConfiguration::FairFastSimRunConfiguration(const TString& geometry
                                                          const bool specialStacking,
                                                          const bool isMT)
     : TG4RunConfiguration(geometry, physicsList, specialProcess, specialStacking, isMT)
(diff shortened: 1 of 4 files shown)


base/sim/fastsim/FairFastSimRunConfiguration.cxx
```diff
@@ -40,6 +41,9 @@ FairFastSimRunConfiguration::~FairFastSimRunConfiguration()
 
 TG4VUserFastSimulation* FairFastSimRunConfiguration::CreateUserFastSimulation()
 {
-    LOG(info) << "Going to create FairFastSimulation";
-    return new FairFastSimulation();
+    if (fUseFastSim) {
+        LOG(info) << "Going to create FairFastSimulation";
+        return new FairFastSimulation();
+    }
+    return 0;
```
rbx, line 48: Do you have an example of how `CreateUserFastSimulation()` is used?
What happens when `0` is returned (it should really be `nullptr`)?
Does the caller have to check for `nullptr` and if it occurs create `TG4RunConfiguration` instead?

The proposal from the related issue looks more reasonable to me (albeit more verbose on the caller side).
It doesn't seem to make sense to create `FairFastSimRunConfiguration` in the first place, if `FairFastSim_UseFastSim` is false.

karabowi: The test shows that there is no difference between using TG4RunConfiguration
and FairFastRunSimConfiguration without FairFastSimulation.
I agree with you though, that it does make more sense to use the original
TG4RunConfiguration in case of standard simulation.

rbx (approved):

base/sim/fastsim/FairFastSimRunConfiguration.cxx
dennisklein: To be honest, I do not think we should add the proposed `FairFastSim_UseFastSim` option. Also the proposed `void FairFastSimRunConfiguration::UseFastSim(bool flag = true)` C++ API makes no sense to me as a user.

dennisklein:
@karabowi Going to reject this PR. If you want, we can make a short meeting next week or so and discuss together a better API to implement the requested `UseFastSim` option.

PR #1050, opened 2021-04-07
cmake: Switch ROOT to config mode

ChristianTackeGSI:
Instead of using the internal find module, let's use ROOT's own cmake config mode system.

This commit does the basic switch over.
It turns all used library targetd over to their ROOT:: equivalent and does some basic cleanup.

It should compile and hopefully run the test suite.

There is more stuff to cleanup afterwards.

---

Checklist:


base/MQ/CMakeLists.txt
```diff
@@ -52,10 +52,6 @@ target_include_directories(${target} PUBLIC
   ${ROOT_INCLUDE_DIR} # TClonesArray, TMessage, etc
```
dennisklein, line 52: Just a heads-up: The usage of `${ROOT_INCLUDE_DIR}` should also be removed, but this is only possible from ROOT 6.20 upwards.

CMakeLists.txt
```diff
@@ -162,7 +162,22 @@ if(SIMPATH)
...
+find_package2(PUBLIC ROOT REQUIRED
+  VERSION 6.18.00
+  COMPONENTS ${required_root_components}
+  OPTIONAL_COMPONENTS ${optional_root_components})
+Find_Program(ROOT_CINT_EXECUTABLE
+  NAMES rootcint
+  PATHS ${ROOT_BINDIR}
+  NO_DEFAULT_PATH
+  )
+include(ROOTMacros)
```
dennisklein, line 180:
```suggestion
```

Nothing in this module should be in use in FairRoot any more, why include it? (The file may be used by experiment repos however)

dennisklein: Ah, I see, it was included in the deleted `FindROOT.cmake`

CMakeLists.txt
```diff
@@ -162,7 +162,22 @@ if(SIMPATH)
...
+list(APPEND optional_root_components gdml vmc)
+find_package2(PUBLIC ROOT REQUIRED
+  VERSION 6.18.00
+  COMPONENTS ${required_root_components}
+  OPTIONAL_COMPONENTS ${optional_root_components})
+Find_Program(ROOT_CINT_EXECUTABLE
+  NAMES rootcint
+  PATHS ${ROOT_BINDIR}
+  NO_DEFAULT_PATH
+  )
```
dennisklein, line 179:
```suggestion
```

Let's remove this and replace its usage in `cmake/modules/FairRootTargetRootDictionary.cmake` with the imported target, something like:

CMakeLists.txt
ChristianTackeGSI: And without this, some things don't work. Don't ask me what. I forgot about it.

So maybe experiments will need to add this `include` somewhere? Because the `find_package` wont load it any more.

CMakeLists.txt
ChristianTackeGSI: * Changed `FairRootTargetRootDictionary.cmake` as suggested.
* `cmake/scripts/generate_dictionary_root.sh.in` has a `@ROOT_CINT_EXECUTABLE@` in it, and I doubt generator expressions work there. So I think, we need a `find_program`.
* `ROOTMacros.cmake` references this `generate_dictionary_root.sh.in`. So maybe experiments use this stuff? Moved the `find_program` there, so that this works for experiments.

base/MQ/CMakeLists.txt
ChristianTackeGSI: Let's move that into a new PR?

ChristianTackeGSI:
Okay, this should be ready for a first real review.

I think, I have already adressed all the raised issues above.

That said:
* Reviewing the complete diff is probably not very pleasant.
* Each commit should be easier to review.
* If needed, this PR can be split at each commit to reduce the review overhead (it compiles / tests fine after each commit).
* I guess, I missed something, but I do think, this is in a shape to be reviewed and considered.

dennisklein (approved):

Issue #1052, opened 2021-04-10
warning: 'strncpy' specified bound depends on the length of the source argument [-Wstringop-overflow=]


dennisklein:
**OS**: Debian 10
**GCC**: 8.3.0

```
MbsAPI/fLmd.c: In function 'fLmdInitMbs':
MbsAPI/fLmd.c:427:9: warning: 'strncpy' specified bound depends on the length of the source argument [-Wstringop-overflow=]
         strncpy(pLmdControl->cFile, Nodename, len);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
MbsAPI/fLmd.c:425:18: note: length computed here
     size_t len = strlen(Nodename);
                  ^~~~~~~~~~~~~~~~
```
```
MbsAPI/fLmd.c: In function 'fLmdPutOpen':
MbsAPI/fLmd.c:142:9: warning: 'strncpy' specified bound depends on the length of the source argument [-Wstringop-overflow=]
         strncpy(pLmdControl->cFile, Filename, len);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
MbsAPI/fLmd.c:140:18: note: length computed here
     size_t len = strlen(Filename);
                  ^~~~~~~~~~~~~~~~
```
```
MbsAPI/fLmd.c: In function 'fLmdConnectMbs':
MbsAPI/fLmd.c:391:9: warning: 'strncpy' specified bound depends on the length of the source argument [-Wstringop-overflow=]
         strncpy(pLmdControl->cFile, Nodename, len);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
MbsAPI/fLmd.c:389:18: note: length computed here
     size_t len = strlen(Nodename);
                  ^~~~~~~~~~~~~~~~
```
```
MbsAPI/fLmd.c: In function 'fLmdGetOpen':
MbsAPI/fLmd.c:590:9: warning: 'strncpy' specified bound depends on the length of the source argument [-Wstringop-overflow=]
         strncpy(pLmdControl->cFile, Filename, len);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
MbsAPI/fLmd.c:588:18: note: length computed here
     size_t len = strlen(Filename);
                  ^~~~~~~~~~~~~~~~
```
```
[...]

PR #1054, opened 2021-04-15
Calm `strncpy` Warnings in MbsApi

ChristianTackeGSI:
`strncpy()` is a really bad API for copying strings over. For example it always pads the destination with NUL bytes. But does not NUL terminate in case of overflow. See first stackoverflow link.

The nicest alternative to `strncpy` is `strlcpy`.  But that one is not portable.  The second simplest one is using `snprintf()`.  Yes, this feels like overkill. But it's really the by far most readable solution. So unless performance is highly critical (in which case you shouldn't be using `strncpy` anyway, as it pads!), `snprintf` should be fine.

Fixes: #1052
See: https://stackoverflow.com/questions/2114896/why-are-strlcpy-and-strlcat-considered-insecure/2115015#2115015
See: https://stackoverflow.com/questions/41869803/what-is-the-best-alternative-to-strncpy/41885173#41885173

---

Checklist:

### DIFF
--- a/MbsAPI/fLmd.c
+++ b/MbsAPI/fLmd.c
@@ -106,6 +106,22 @@ lmdoff_t fLmdOffsetGet(sLmdControl*, uint32_t);
 void fLmdOffsetElements(sLmdControl*, uint32_t, uint32_t*, uint32_t*);
 #define OFFSET__ENTRIES 250000
 
+/**
+ * Internal Helper functions
+ */
+static void _fLmd_set_cFile(sLmdControl* pLmdControl, const char* Filename)
+{
+    int snp_res;
+
+    if ((pLmdControl == NULL) || (Filename == NULL)) {
+        return;
+    }
+    snp_res = snprintf(pLmdControl->cFile, sizeof(pLmdControl->cFile), "%s", Filename);
+    if ((snp_res < 0) || (snp_res >= sizeof(pLmdControl->cFile))) {
+        printf("Filename %s too long!", Filename);
+    }
+}
+
 //===============================================================
 uint32_t fLmdPutOpen(sLmdControl* pLmdControl,
                      char* Filename,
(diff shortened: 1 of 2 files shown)


ChristianTackeGSI:
Requesting review by @fuhlig1 as he touched this strncpy business in the past..

MbsAPI/fLmd.c
```diff
@@ -137,12 +137,7 @@ uint32_t fLmdPutOpen(sLmdControl* pLmdControl,
 
     pLmdControl->pMbsFileHeader->iType = LMD__TYPE_FILE_HEADER_101_1;
     pLmdControl->pMbsFileHeader->iEndian = 1;
-    size_t len = strlen(Filename);
-    if (len < sizeof(pLmdControl->cFile)) {
-        strncpy(pLmdControl->cFile, Filename, len);
-    } else {
-        strncpy(pLmdControl->cFile, Filename, sizeof(pLmdControl->cFile) - 1);
-    }
+    snprintf(pLmdControl->cFile, sizeof(pLmdControl->cFile), "%s", Filename);
```
fuhlig1, line 140: Maybe one could add a warning when the length of the source is larger than the destination. Otherwise the filename is truncated without any information. This should have done already in the past.

The same holds for all the other places were this pattern appears.

fuhlig1: Actually I touched the code in the past to avoid strcpy. Obviously I didn't understood strncpy and did not improve the code but only got rid of the warning.

ChristianTackeGSI: Well, strncpy is already an improvement over strcpy, as it doesn't write past the end of the buffer.

My original intention was to have a minimal patch that doesn't change the intended semantics (truncate without warning).
I have added some printing of warnings in this update. Is it now better?
(I haven't done it for all cases that I touched).

fuhlig1: I think this is fine. Probably @kresan should also comment on this issue. R3B is meanwhile the main user of the MBSApi.

ChristianTackeGSI:
So should we merge this one anyway?

Should I add the "disable MBS by default" to this PR, or to a new one?

MohammadAlTurany:
Yes, we should merge it

MbsAPI/fLmd.c
fuhlig1: Since @kresan confirmed in the meeting today that R3B doesn't use it any longer please merge it

fuhlig1 (approved):

Issue #935, opened 2020-03-03
In FairRootManager::RegisterAny, no check whether a Sink exist before using it when persistence activated


PALoizeau:
**Describe the bug**
If one enables the persistence of a branch while registering it with RegisterAny without adding a sink to the run, it leads to a segmentation fault.

**Known affected versions:**
- v18.2.1
- master

**To Reproduce**
In a macro or class, register an output vector using RegisterAny with persistence enabled but without adding before a Sink such as a FairRootFileSink to the run.
The bug was found with a FairRunOnline, FairRootFileSink and std::vector combination.

**Expected behavior**
One of the following two:
- assert or "fatal" error message indicating that no Sink is defined => Fail the run
- "error" error message indicating that the persistence flag could not be applied due missing Sink => Allow run to go on but message could be missed

**Corresponding line in master:**
https://github.com/FairRootGroup/FairRoot/blob/e6cd7da7d60d0983cd0c1014a611376a1e8fe0c4/base/steer/FairRootManager.h#L432-L436

PR #1055, opened 2021-04-16
Avoid mandatory Fatal in FairRunOnline if no Sink defined

PALoizeau:
This was an unintended consequence/oversight following #940 (fixing issue #935).

The newly introduced FATAL was triggered by the hard-coded `write output flag = kTRUE` (persistence) when the header object was registered in the FairRoot manager.
The change keeps the original behavior in case a Sink is existing and skip the writing otherwise, avoiding the FATAL for pure monitoring tasks without ROOT file output .

Affected branches and releases:
- master and dev
- v18.4.x
- v18.6.x

---

Checklist:

### DIFF
--- a/base/steer/FairRunOnline.cxx
+++ b/base/steer/FairRunOnline.cxx
@@ -182,7 +182,7 @@ void FairRunOnline::Init()
         return;
     }
     LOG(info) << "FairRunOnline::InitContainers: event header at " << fEvtHeader;
-    fRootManager->Register("EventHeader.", "Event", fEvtHeader, kTRUE);
+    fRootManager->Register("EventHeader.", "Event", fEvtHeader, (nullptr != fRootManager->GetSink()));
     fEvtHeader->SetRunId(fRunId);
 
     fRootManager->GetSource()->InitUnpackers();


fuhlig1 (dismissed):

fuhlig1:
@dennisklein,

is there a chance to restart the failing test or is it only possible to rerun the complete matrix?

dennisklein:
> @dennisklein,
> 
> is there a chance to restart the failing test or is it only possible to rerun the complete matrix?

sequential pipeline steps can be re-run separately, but not a parallel step like we have in this case. I was searching for such a functionality a while back, but only found open issues for this topic back then.

dennisklein:
> > @dennisklein,
> > is there a chance to restart the failing test or is it only possible to rerun the complete matrix?
> 
> sequential pipeline steps can be re-run separately, but not a parallel step like we have in this case. I was searching for such a functionality a while back, but only found open issues for this topic back then.

but in order to reduce test failure noise, I have enabled the ctest option `--rerun until-pass:5`, unfortunately this only works from CMake 3.17+ and the ubuntu 20.04 image has still CMake 3.16.

fuhlig1:
@dennisklein,

do you by chance if there is a possibility to allow the option  --rerun until-pass:5 only for some tests. I am asking since in CBM mostly the Geant3 simulations crash due to the well known memory problem of Geant3. If I could allow a rerun only for those would help me a lot.

dennisklein:
> @dennisklein, do you by chance if there is a possibility to allow the option --rerun until-pass:5 only for some tests. I am asking since in CBM mostly the Geant3 simulations crash due to the well known memory problem of Geant3. If I could allow a rerun only for those would help me a lot.
> […](#)
> Am 19.04.2021 um 21:57 schrieb Dennis Klein ***@***.***>: @dennisklein, is there a chance to restart the failing test or is it only possible to rerun the complete matrix? sequential pipeline steps can be re-run separately, but not a parallel step like we have in this case. I was searching for such a functionality a while back, but only found open issues for this topic back then. but in order to reduce test failure noise, I have enabled the ctest option --rerun until-pass:5, unfortunately this only works from CMake 3.17+ and the ubuntu 20.04 image has still CMake 3.16. — You are receiving this because your review was requested. Reply to this email directly, view it on GitHub, or unsubscribe.

AFAIK it is not possible from the command line arguments of `ctest`, but you could do this with multiple [`ctest_test`](https://cmake.org/cmake/help/latest/command/ctest_test.html) in the cmake script with `ctest -S...` script mode:

```
ctest_test(EXCLUDE geant3test)
ctest_test(INCLUDE geant3test REPEAT UNTIL_PASS:5)
```

Unfortunately there have not added the REPEAT parameter as a [test property](https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#properties-on-tests).

dennisklein:
Actually, `ctest` also has `-E` and `-R`, so with multiple invocations of `ctest` it would also work from the command line.

And if the test names are too complicated for regular expressions, you could add test labels to select them with `-L` and `-LE`.

PR #1056, opened 2021-04-19
Prepare for Disabling MbsAPI

ChristianTackeGSI:
We want to remove MbsAPI.
So let's deprecate it first.

* A `BUILD_MBS` option already exists.
* Make sure, that setting it to OFF actually works.
* Do not yet set it to OFF
* Add `libdl` to `MCStepLogger`
* Move myself from CONTRIBUTORS to AUTHORS

---

Checklist:


ChristianTackeGSI:
I have left out the "set BUILD_MBS default to OFF" for now.
I can add that commit soon here, or create a new PR after this one is merged.
What do you prefer?

CMakeLists.txt
```diff
@@ -308,7 +308,9 @@ else()
   message(STATUS "eventdisplay will not be built, because ROOT has no opengl support.")
 endif()
 
-add_subdirectory(MbsAPI)
+if(BUILD_MBS)
```
dennisklein, line 311: Let's also declare this option as official build [option](https://cmake.org/cmake/help/latest/command/option.html) somewhere before this line.

dennisklein (changes requested):
See comment, rest looks good!

dennisklein:
> I have left out the "set BUILD_MBS default to OFF" for now.
> I can add that commit soon here, or create a new PR after this one is merged.
> What do you prefer?
> 
> Oh, and please add anyone in, who might be interested.

ah, yes, please set the default to `OFF` already.

CMakeLists.txt
dennisklein: hm, seeing this, let's also rename it to `BUILD_MBSAPI` to match the component name.

ChristianTackeGSI: It's already an official `option`:
https://github.com/FairRootGroup/FairRoot/blob/dev/CMakeLists.txt#L295

So should we rename it, despite it already being "established"?
* Con: The name is already there, in theory people could already know about it and use `-DBUILD_MBS=ON` so be sure, they get this.
* Pro: It was broken anyway. (at least the OFF setting)

I don't have an opinion on this.

dennisklein: > So should we rename it, despite it already being "established"?

No, in that case, let's leave it. Sry, didn't remember this.

dennisklein:
> An option BUILD_MBS is already available.
> Setting it to OFF breaks the build currently.

What breaks?

ChristianTackeGSI:
> What breaks?

Phrased badly.
Better: "Without this PR".

So without this PR doing a `-DBUILD_MBS=OFF` did not work. Basically, because the examples were building, but Mbs did not.

That's why I added a bunch of more `if(BUILD_MBS)` around the place. To stop building any depenendencies of MBS.

dennisklein (approved):

PR #1058, opened 2021-04-21
cmake: Drop some variables, little cleanup

ChristianTackeGSI:
* Drop AlFa_DIR variable 
* Drop GCC_TOOLCHAIN_ROOT variable
* Clean up Pythia8 env usage
* Move example installation into examples/
* Drop global include directories

---

Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -354,13 +354,6 @@ if(BUILD_UNITTESTS)
   add_subdirectory(test)
 endif()
 
-if(BUILD_EXAMPLES)
-  install(DIRECTORY
-    examples/common/gconfig
-    examples/common/geometry
-    DESTINATION ${PROJECT_INSTALL_DATADIR}/examples/common
-  )
-endif(BUILD_EXAMPLES)
 
 install(FILES ${CMAKE_BINARY_DIR}/config.sh_install
   DESTINATION bin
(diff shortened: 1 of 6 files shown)


dennisklein (approved):
> Note to ourselves: The changes in `FindPythia8.cmake` will most likely require changes in https://github.com/alisw/alidist/blob/master/pythia.sh and/or https://github.com/alisw/alidist/blob/master/fairroot.sh.

1. Can we already make those "forward compatible" by already now adding something useful?
2. Should we go a step back and not drop too much from our FindPythia8?

dennisklein:
> > Note to ourselves: The changes in `FindPythia8.cmake` will most likely require changes in https://github.com/alisw/alidist/blob/master/pythia.sh and/or https://github.com/alisw/alidist/blob/master/fairroot.sh.
> 
> 1. Can we already make those "forward compatible" by already now adding something useful?
> 2. Should we go a step back and not drop too much from our FindPythia8?

I think it is fine, once we bump to FairRoot to `19` in alidist, we should just clean up the recipe(s).

ChristianTackeGSI:
> I think it is fine, once we bump FairRoot to `19` in alidist, we should just clean up the recipe(s).

Should we file some issue on FairRoot to keep track?

PR #1059, opened 2021-04-22
Use FairCMakeModules for find_package2

ChristianTackeGSI:
* Make cmake module installation more granular (only install specific subdirs of `cmake/`)
* Optionally use FairCMakeModules for some modules. Fallback is in `cmake/fallback/`
* Extract `find_package` into `FairFindPackage2.cmake` (added to fallback directory)

---

Checklist:


CMakeLists.txt
```diff
@@ -98,6 +98,12 @@ If(NOT _HAS_CXX11_FLAG)
 EndIf()
 
 # Load some basic macros which are needed later on
+find_package(FairCMakeModules 0.1 CONFIG)
```
dennisklein, line 101: Why is the `CONFIG` needed?

CMakeLists.txt
```diff
@@ -98,6 +98,12 @@ If(NOT _HAS_CXX11_FLAG)
...
+if(NOT FairCMakeModules_FOUND)
+  message(WARNING "Please highly consider installing FairCMakeModules!")
```
dennisklein, line 103:
```suggestion
  message(WARNING "Could not find FairCMakeModules, falling back to local copy. It is recommended to install https://github.com/FairRootGroup/FairCMakeModules!")
```

dennisklein: Maybe also consider making it a `STATUS` message, `WARNING` is quite "loud".

CMakeLists.txt
ChristianTackeGSI: I found the verbosity about find-modules annoying. As we now have a dedicated warning (see below), a QUIET is probbaly better?

CMakeLists.txt
ChristianTackeGSI: Will take the better message!

Well, I would like to remove the fallback in 19.2. So I want to be loud in the 19.0 release.

dennisklein (approved):

PR #1060, opened 2021-04-27
Start to Refactor Summary Output

ChristianTackeGSI:
Refactor the summary output into some functions.
This really only moves the functionality over into functions.

There is one commit with some slight improvments.

Making these things more generic/etc is left for another PR.

---

Checklist:


cmake/private/FairRootSummary.cmake
```diff
@@ -0,0 +1,218 @@
...
+# Copyright (C) 2014-2021 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  #
+#                                                                              #
+#              This software is distributed under the terms of the             #
+#              GNU Lesser General Public Licence (LGPL) version 3,             #
+#                  copied verbatim in the file "LICENSE"                       #
+################################################################################
+
+function(fairroot_summary_header)
+  message(STATUS "  ")
+  message(STATUS "               C O N F I G U R A T I O N   S U M M A R Y")
```
dennisklein, line 11:
```suggestion
```

Too verbose IMHO.

cmake/private/FairRootSummary.cmake
```diff
@@ -0,0 +1,218 @@
...
+  message(STATUS "  ")
+  message(STATUS "  ${Cyan}PROJECT NAME${CR}         ${BGreen}${PROJECT_NAME}${CR}")
+  message(STATUS "  ${Cyan}VERSION${CR}              ${BGreen}${PROJECT_VERSION}${CR}")
```
dennisklein, line 14: Would be cool, if we print also the git version in brackets, if available and different from `PROJECT_VERSION`. But doesn't need to happen in this PR.

cmake/private/FairRootSummary.cmake
```diff
@@ -0,0 +1,218 @@
...
+#              This software is distributed under the terms of the             #
+#              GNU Lesser General Public Licence (LGPL) version 3,             #
+#                  copied verbatim in the file "LICENSE"                       #
+################################################################################
+
+function(fairroot_summary_header)
+  message(STATUS "  ")
+  message(STATUS "               C O N F I G U R A T I O N   S U M M A R Y")
+  message(STATUS "  ")
+  message(STATUS "  ${Cyan}PROJECT NAME${CR}         ${BGreen}${PROJECT_NAME}${CR}")
```
dennisklein, line 13: Could you please rename the project to `FairRoot`? Unless something depends on it, some `FAIRROOT_*` variables possibly. If it is too much, then let's do it in a separate PR.

dennisklein (changes requested):
Looks fine to me, pls remove the heading. The other two remarks are optional, we can delay them to later.

cmake/private/FairRootSummary.cmake
ChristianTackeGSI: I tried. And some variables depend on it. It took me a bit of time to track some of them down. And I still fear I missed some. So let's move that into another PR.

cmake/private/FairRootSummary.cmake
ChristianTackeGSI: Commented out for now. Maybe we will find a not "too verbose" heading?

cmake/private/FairRootSummary.cmake
ChristianTackeGSI: Added FAIRROOT_GIT_VERSION in brackets, if it is set.

ChristianTackeGSI:
Also added CXX standard checking:
Check that a given `CMAKE_CXX_STANDARD` maches up with the one from ROOT.

dennisklein:
fairroot_summary

Hmm, looking at the added lines again and thinking about the added value, I am not sure, if I like it. Main goal was to put the typical info which is relevant for the user and in a way that it is easy to learn what are the options one would most likely change from time to time. A secondary goal was to put information which is useful in logs and not shown by CMake by default or not in a compact way easy for the eye.

We are now adding 7 questionable lines regarding the goals from above:

* `PROJECT NAME` will basically never change and is of no importance in the summary at all. For logging just the directory itself and such already make clear which project it is. IMHO, remove it or see the example with FairMQ later.
* `VERSION` is somewhat interesting, definitely for logging, also see FairMQ example later.
* `INSTALL PREFIX` is absolutely fine, probably the most changed option.
* the top-level install tree dirs are questionable IMHO. What do you think about not show them if they are the default ones, and only show the ones that were changed (for logging)?

This shows how one could save a lot of space for project name and version:

fmq_summary

Looking at version now, it shows `18.6.0 (18.4.1-94-g8711c788a)` - I realize both are not of any use (partially because of the way I managed the branches). But here we are neither at version `18.6.0`, nor are we at `18.4.1-94-g8711c788a` (technically yes, but not for the average viewer). So, this needs more logic now. I will have a look into this.

CMakeLists.txt
```diff
@@ -181,6 +183,9 @@ find_package2(PUBLIC ROOT REQUIRED
   COMPONENTS ${required_root_components}
   OPTIONAL_COMPONENTS ${optional_root_components})
 include(ROOTMacros)
+
+fairroot_check_root_cxxstd()
```
dennisklein, line 187:
```suggestion
fairroot_check_root_cxxstd_compatibility()
```

Somehow, I was missing what the check is about. What do you think?

ChristianTackeGSI: Yeah, maybe that name is better! Next push will have the new name.

Do we need more docs on this?

dennisklein: Some inline comments above each function with a brief description would be good style, but we can add it, when we migrate it to FairCMakeModules

dennisklein:
> Okay, added a lot of STATUS -> VERBOSE in there.
> So now you need to do
> 
> `$ cmake … --log-level=VERBOSE`
> to see this stuff.
> 
> Maybe that's better now?

Yes, awesome.

> I am currently highly considering to add such a flag in our spack recipe

Agreed. It is a good idea.

PR #1061, opened 2021-04-30
cmake: Introduce fairroot_library_settings()

ChristianTackeGSI:
This is an internal function only.  It encapsulates things that already happened and one new.

It is meant for all those libraries that get installed and are expected to be used by other projects.
It is explicitly not meant for the examples. Because they're not meant to be consumed by other projects.

Already happened:
* Add the FairRoot::${target} alias
* Set PROJECT_LIBRARY_PROPERTIES properties

New: fairroot_compile_features()

This sets cxx_std_* on the target.  This isn't very granular, but hopefully fits the current development model better.
The value being used depends on CMAKE_CXX_STANDARD or PROJECT_MIN_CXX_STANDARD.
So when exported targets come, consumers will at least use the CMAKE_CXX_STANDARD.

---

Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,12 +92,6 @@ Set(CheckSrcDir "${CMAKE_SOURCE_DIR}/cmake/checks")
 include(CheckCXX11Features)
 include(CheckSymbolExists)
 
-# FairRoot only uses C++11 or later, so we need an compiler which supports C++11
-# Check if the used compiler support C++11. If not stop with an error message
-If(NOT _HAS_CXX11_FLAG)
-  Message(FATAL_ERROR "The used C++ compiler (${CMAKE_CXX_COMPILER}) does not support C++11. CbmRoot can only be compiled with compilers supporting C++11. Please install such a compiler.")
-EndIf()
-
 # Load some basic macros which are needed later on
 find_package(FairCMakeModules 0.1 QUIET)
 if(NOT FairCMakeModules_FOUND)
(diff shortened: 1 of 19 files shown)


rbx:
Just my 5 cents regarding adding custom stuff like `fairroot_library_settings`:
unless it's super useful and hides lots of complexity, like `find_package2`, I would stick to a few standard cmake commands. That way (1) people familiar with cmake immediately understand what is going on when reading it and (2) people new to cmake have an easier time learning standard useful cmake from our code and reusing it for their own code.

ChristianTackeGSI:
> At least for the cmake files in examples directories I would use standard cmake.

Okay, what about: I'll unroll the things in examples only and we leave the rest as "internal code"?

Second: What about `fairroot_compile_features`? Should we put a plain
```cmake
target_compile_features("${target}" PUBLIC cxx_std_11)
```
in the examples instead? Is everybody aware of maintaining the CXX-level on a per target basis (at least for the examples)?

rbx:
> > At least for the cmake files in examples directories I would use standard cmake.
> 
> Okay, what about: I'll unroll the things in examples only and we leave the rest as "internal code"?

Sounds good to me. Ultimately you and/or FairRoot should decide where it's useful. Just wanted to add some thoughts for consideration.

dennisklein:
> Is everybody aware of maintaining the CXX-level on a per target basis (at least for the examples)?

Obviously not! Stuff will just be copied around like before with only guessing what stuff does, why would that change^^?

ChristianTackeGSI:
Dropped all the changes in `examples/`.
Rephrased commit and PR.
Rebased.

PR #1062, opened 2021-05-06
Start Extremely Minimal Config Mode Package

ChristianTackeGSI:
In the long run, we want to move to a clean config mode package for FairRoot. So let's go small steps in that direction.

* Add the smallest skeleton
* Add exported targets

---

Checklist:


cmake/private/FairRootTargets.cmake
```diff
@@ -36,3 +36,13 @@ function(fairroot_library_settings target)
...
+
+
+set(PROJECT_EXPORT_SET FairRootTargets)
+
+
+function(fairroot_install_exported_library target)
+  install(TARGETS ${target}
+    EXPORT ${PROJECT_EXPORT_SET}
+    LIBRARY DESTINATION ${PROJECT_INSTALL_LIBDIR})
+endfunction()
```
dennisklein, line 48:
```suggestion
function(fairroot_install)
  cmake_parse_arguments(PARSE_ARGV 0 ARGS "" "" "TARGETS")

  install(TARGETS ${ARGS_TARGETS}
    EXPORT ${PROJECT_EXPORT_SET}
    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
    ARCHIVE ${PROJECT_INSTALL_LIBDIR}
    LIBRARY DESTINATION ${PROJECT_INSTALL_LIBDIR})
endfunction()
```

* doesn't need to be library-specific

ChristianTackeGSI: Updated, force pushed.
I named it still `fairroot_install_exported`, because it does the EXPORT unconditionally for now.
As it is a private function/command, we can rename it as needed.

dennisklein (approved):

ChristianTackeGSI:
Thanks!

PR #1078, opened 2021-07-12
creating an actual core dump.

klenze:
Previously, the code did not work as intended: It did not create a core dump.

Now it does (unless ulimit -H -c  0 is explicitly set).

---

Checklist:

* [?] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/fairtools/FairLogger.cxx
+++ b/fairtools/FairLogger.cxx
@@ -19,10 +19,19 @@
 #include <iostream>    // for cout, cerr
 #include <stdio.h>     // for fclose, freopen, remove, etc
 
+#include <stdio.h>        // perror
+#include <sys/time.h>  
+#include <sys/resource.h> // setrlimit
+#include <signal.h>       // signal
+#include <stdlib.h>       // abort
+
+
 FairLogger* FairLogger::instance = nullptr;
 
 const char FairLogger::endl = ' ';
 const char FairLogger::flush = ' ';
+
+
 
 FairLogger::FairLogger()
     : fLogFileName()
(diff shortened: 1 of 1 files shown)


dennisklein:
I am confused what was the original intention here to be honest. I checked in older versions of FairLogger and it looks to me that the `core_dump_<pid>` file always just contained a stack trace and never a core dump, so the name may be just misleading? @fuhlig1 ?

fuhlig1:
The code never create a core dump file, it simply creates a stack trace, so the message as well as the file name is misleading.

klenze:
Yes, the message/filename was misleading. A compromise might be to remove the setrlimit line in my PR. 
This way, if the user explicitly set ulimit -c unlimited, they will get their core dump, and otherwise, they get a quick crash, which I would prefer as a default for LOG(fatal): unlike e.g. a SIGSV, LOG(fatal) points directly to a file, line and what went wrong. 

(In R3BRoot, most of the LOG(fatal) appear when a FairTask does not find resources required, such as a Branch, some Parameters or FairSingletons. In these cases, the list of calls leading to e.g. MyFairTask::Exec() is very unlikely to reveal useful information about what went wrong.)

The Root stack trace is strictly less useful than a core dump and takes a longer time to create, in my experience. 

(Generally, I wish ROOT would stop installing these sighandlers per default and instead relied on the OS creating core file, which have served programmers well since before the dawn of unix.)

dennisklein:
> (Generally, I wish ROOT would stop installing these sighandlers per default and instead relied on the OS creating core file, which have served programmers well since before the dawn of unix.)

I personally fully agree. Actually there seems to be a root setting `Root.Stacktrace: no`. If this does switch off ROOT's sighandler, I think FairLogger shall also honor it and only generate the stacktrace file if it is set to `yes`. @fuhlig1 @klenze What do you think?

fuhlig1:
For me this sounds like a good proposal.

dennisklein:
* We clarified that the name of the created file `core_dump` is misleading and never meant to contain an actual core dump.
* We verified that the rootrc setting `Root.Stacktrace: no` can be used to opt-out of any stacktrace generation after a fatal exception.
   * It turned out, the `core_dump` file written by FairRoot is even empty in this case. So, #1096 now handles this case a bit better and does not create the empty file any more.

> Generally, I wish (...) and instead relied on the OS creating core file, which have served programmers well since before the dawn of unix.

I fully agree that FairRoot or Root should not interfere with our OS settings about creating core dumps. In combination with the rootrc setting `Root.Stacktrace: no` I believe you can achieve the desired behaviour which renders this PR obsolete, closing it therefore. Thank you for pointing out this issue!

Issue #1079, opened 2021-07-15
Feature request: allow for external RTDB backends


patternoia:
**Is your feature request related to a problem? Please describe.**
Hello,

as discussed on Monday's at FairRoot group meeting I am putting up this issue to track the proposed change to FairRoot.

In CbmRoot we implemented a TGenBase RTDB backend, but in order for it to work as expected, FairRoot shall be patched, additional clauses in the switch statement shall be added, etc. (see the presentation) Because of this patch TGenBase becomes a hard dependency of FairRoot which should not be the case.

**Describe the solution you'd like**
I propose to retouch the RTDB code to clearly define the RTDB's `ParIo` API so that any class implementing this API shall be recognized by the FairRoot properly.

**Additional context**
See [my solution](https://git.cbm.gsi.de/TGenBase/FairRoot/-/commit/251d0734e46420f621a07b023bddff27d7d56ceb?view=inline) in our gitlab.

Thanks for consideration.
Evgeny.

dennisklein:
After having reviewed the affected code, I have to reclassify this as a bug report. 08609c153e4aeaedce4d3367506721d2c3db1f6b introduced a regression by closing down the set of `FairParIo`s. What you are asking here has already been a feature 9 years ago AFAICS ;)

PR #1081, opened 2021-07-21
Make rtdb less restrictive regarding backends

patternoia:
I went ahead and implemented the small changes requested in #1079 against 18.2.1 used in CbmRoot.

This way I was able to create an RTDB backend in an external project and properly use it.

If possible I'd like to request a new 18.2.2 release with this code, so we can use it in current CbmRoot.

The same changes are also available for current master and can be proposed in a separate PR.

---

Checklist:

### DIFF
--- a/parbase/FairRuntimeDb.cxx
+++ b/parbase/FairRuntimeDb.cxx
@@ -18,10 +18,8 @@
 #include "FairContFact.h"            // for FairContFact
 #include "FairDetParAsciiFileIo.h"   // for FairDetParAsciiFileIo
 #include "FairDetParRootFileIo.h"    // for FairDetParRootFileIo
-//#include "FairDetParTSQLIo.h"           // for FairDetParTSQLIo
 #include "FairGenericParAsciiFileIo.h"   // for FairGenericParAsciiFileIo
 #include "FairGenericParRootFileIo.h"    // for FairGenericParRootFileIo
-//#include "FairGenericParTSQLIo.h"       // for FairGenericParTSQLIo
 #include "FairLogger.h"           // for FairLogger, MESSAGE_ORIGIN
 #include "FairParAsciiFileIo.h"   // for FairParAsciiFileIo
 #include "FairParIo.h"            // for FairParIo
@@ -43,11 +41,6 @@ using std::cout;
 using std::endl;
 using std::ios;
 using std::setw;
-
-/// DEBUG DEBUG Temporary define
-// 0 = use original code; 1 = use new code
-#define USE_DB_METHOD 1
-/////////////////////////////////
 
 ClassImp(FairRuntimeDb);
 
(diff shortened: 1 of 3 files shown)


parbase/FairRuntimeDb.cxx
```diff
@@ -461,8 +461,12 @@ Bool_t FairRuntimeDb::writeContainer(FairParSet* cont, FairRtdbRun* run, FairRtd
...
       break;// End of Ascii IO
     default: // Unknown IO
-      Error("writeContainer()","Unknown output file type.");
-      break;
+      if (cont->hasChanged()) {
+        cv = findOutputVersion(cont);
+        if(cv == 0) {
+          cont->write(output);
+        }
+      }
```
dennisklein, line 469: * I think all those case statements should be moved to `FairParSet::write(FairParIo*)` overrides in the corresponding child classes and your proposed addition should go into the base class.
   * Then replace the whole `switch(ioType)` statement with just something like `cont->write(output)`
* Then we should get rid of the `FairRuntimeDb::ioType` member (`FairRuntimeDb::isRootFileOutput` can stay for the scope of this PR)
* Please also remove the `USE_DB_METHOD` macro from this file entirely (including comments) and also all code which is no longer enabled, because it is anyways hard-coded to the value `1`

dennisklein (changes requested):

dennisklein:
> If possible I'd like to request a new 18.2.2 release with this code, so we can use it in current CbmRoot.

Please rebase this PR against the `dev` branch. Once it is merged there, I will backport it to all minor releases 18.2, 18.4, and 18.6.

parbase/FairRuntimeDb.cxx
dennisklein: Please make 3 separate commits, one for each of the above main bullet points.

patternoia:
Does this mean, that reverting the code shall be enough?

parbase/FairRuntimeDb.cxx
patternoia: I do not think that this code should be moved to the inheritors of `FairParSet` but to inheritors `FairDetParIo`, like `FairDetParRootFileIo`. This does not seem to be trivial though. I can attempt to, but I am not quite sure the implementation will be proper

mainnet-pat: Also dev is incompatible with my current fairsoft version, will take time to do the compilation.
Can you please go ahead and implement the proposed changes?

patternoia: Also dev is incompatible with my fairsoft version and compiling the compatible version will take its long time.
Can you please go ahead and implement the suggested changes? Thanks.

dennisklein: > I do not think that this code should be moved to the inheritors of `FairParSet` but to inheritors `FairDetParIo`, like `FairDetParRootFileIo`.

You are absolutely right, I confused these two `write`s:
* `Int_t FairParSet::write(FairParIo*)`
* `Int_t FairDetParIo::write(FairParSet*)`

Whose purpose I do not yet understand tbh. Will have to have another deeper look at this.

> This does not seem to be trivial though. I can attempt to, but I am not quite sure the implementation will be proper
> Can you please go ahead and implement the suggested changes? Thanks.

Sure, but unfortunately not immediately, I will try to squeeze it in next week or so. Can you comment on what exactly might be non-trivial (so I do not miss it)?

dennisklein:
> Does this mean, that reverting the code shall be enough?

Probably not for the cases of the built-in `FairDetParIo`s, but it would be interesting to know, if reverting 08609c153e4aeaedce4d3367506721d2c3db1f6b would make it work for your external `FairDetParIo` child class.

parbase/FairRuntimeDb.cxx
patternoia: I do, similarly to you, struggle to understand all the logic and purpose of these `ParIo`s. Writing logic is scattered across many classes and not very transparent.

patternoia:
@dennisklein I've followed your suggestions and implemented the changes. Since there are no unit and integration tests for rtdb, I have ran some of my own tests and it seems to work fine.
Disregard the source branch name, I have force pushed into it and it is now equivalent to the upstream dev.

dennisklein:
The `ex_sim_tutorial2_read_digis` test segfaults now consistently: https://cdash.gsi.de/testSummary.php?project=3&name=ex_sim_tutorial2_read_digis&date=2021-07-27. 

Looks like some of the code in the case statements was important. To untangle this, one has to understand this `Int_t FairParSet::write(FairParIo*)` <-> `Int_t FairDetParIo::write(FairParSet*)` convolution to be able to move the case code to the appropriate `FairDetParIo` subclasses. I will have a look at it, but will take a bit.

patternoia:
I was incorrect regarding tests. They were just disabled for me due to my custom `-DBBUILD_EXAMPLES=OFF`
Which is quite interesting.

parbase/FairRuntimeDb.cxx
```diff
@@ -363,144 +355,42 @@ Int_t FairRuntimeDb::findOutputVersion(FairParSet* cont)
...
-    const Text_t* c = cont->GetName();
-    LOG(debug) << "RuntimeDb: write container: " << cont->GetName();
-    FairParVersion* vers = run->getParVersion(c);
-    Bool_t rc = kTRUE;
-    Int_t cv = 0;
-    if (getOutput() && output->check() && output->isAutoWritable()) {
-        switch (ioType) {
-            case RootFileOutput:   // RootFile
-                if (cont->hasChanged()) {
-                    cv = findOutputVersion(cont);
```
dennisklein, line 443: If we remove this, we should make sure, it is no longer needed.

patternoia: I did remove it since (from my point of view) requiring `cont->hasChanged()` and `findOutputVersion(cont) == 0` is the same check.

dennisklein: > I did remove it since (from my point of view) requiring `cont->hasChanged()` and `findOutputVersion(cont) == 0` is the same check.

I agree that this is the case, if the `FairParSet` input versions at indices `1` and `2` are `-1`, but what about these cases:
https://github.com/FairRootGroup/FairRoot/blob/0d7a49edf0c4a2fe7cbc247d4e867abb110c4fa1/parbase/FairRuntimeDb.cxx#L350-L362

@fuhlig1 @MohammadAlTurany Do you have more insight here? What are these input versions used for? Why have these cases above been handled for root file and TSQL outputs und not for ASCII file output. Shouldn't they be handled always?

MohammadAlTurany: As far as I remember ASCII does not have versions, it is only for debugging or overwriting parameters.

dennisklein:
@patternoia After some group-internal discussions we have decided to accept and merge your original change proposal (see #1093). The refactoring I have requested from you during the course of the review in this PR turns out to be more complicated than I hoped for. However, your progress is not lost, but recorded in the [feature/parbase_refactor](https://github.com/FairRootGroup/FairRoot/tree/feature/parbase_refactor) branch for now (so you may delete this branch in your own clones of FairRoot if you wish). We have decided to address some issues with the parbase module more generally and your work will gladly be incorporated!

I will close this PR for technical reason, it is based against the v18.2, but our workflow foresees to develop against the `dev` branch only and backports to past releases will be created by cherry-picking those commits (which I will do without explicit PRs in this case).

Thank you very much for your contribution! (You should get notified when this lands within the next days through the github release notifications, if you are subscribed to the FairRoot repo)

dennisklein:
Note: There is also a backport of this in the `v18.2_patches` branch, but I won't make another point release for it.

Issue #1002, opened 2020-08-28
False positives on load_all_libs test


dennisklein:
```
test 3
      Start  3: load_all_libs

3: Test command: /tmp/fairsoft_ctest.1X6EpE/spSF4V/FairRoot/build/examples/simulation/Tutorial1/macros/load_all_libs.sh
3: Test timeout computed to be: 60
3:    ------------------------------------------------------------
3:   | Welcome to ROOT 6.18/04                  https://root.cern |
3:   |                               (c) 1995-2019, The ROOT Team |
3:   | Built for linuxx8664gcc on Sep 11 2019, 15:38:23           |
3:   | From tags/v6-18-04@v6-18-04                                |
3:   | Try '.help', '.demo', '.license', '.credits', '.quit'/'.q' |
3:    ------------------------------------------------------------
3: 
3: /bin/bash: /tmp/fairsoft_ctest.1X6EpE/spSF4V/fairsoft/lib/libtinfo.so.6: no version information available (required by /bin/bash)
3: 
3: Processing /tmp/fairsoft_ctest.1X6EpE/spSF4V/FairRoot/examples/simulation/Tutorial1/macros/load_all_libs.C()...
[...]

PR #1082, opened 2021-07-21
Be more Pedantic During Linking

ChristianTackeGSI:
On Linux, when creating a shared library, it is usually allowed to have undefined symbols.
One reason, why this might be useful: A "main" program that links to the library provides that symbol.

None of the shared libraries in FairRoot use this feature.
So let the linker enforce this.
Undefined symbols will now result in an error.

See also: #1002

---

Checklist:


ChristianTackeGSI:
Marked as draft until I have feedback on possibly removing the `load_all_libs` test (#1002). If the test is to be removed, I would do that in this PR as well.

dennisklein (approved):

Issue #1005, opened 2020-09-02
warning: '%i' directive writing between 1 and 11 bytes into a region of size 10 [-Wformat-overflow=]


dennisklein:
**OS**: Ubuntu 20.04
**Compiler**: GCC 9.3.0
**FairRoot**: dev @ 71dedec0fbeee1ff0a093709a6916c11663951f5

```
/tmp/jenkins-FairRootGroup-FairSoft-PR-314-1.zup/env_post_workdir.PmH/FairRoot/examples/simulation/rutherford/src/FairRutherfordGeo.cxx: In member function 'virtual const char* FairRutherfordGeo::getModuleName(Int_t)':
/tmp/jenkins-FairRootGroup-FairSoft-PR-314-1.zup/env_post_workdir.PmH/FairRoot/examples/simulation/rutherford/src/FairRutherfordGeo.cxx:32:33: warning: '%i' directive writing between 1 and 11 bytes into a region of size 10 [-Wformat-overflow=]
   32 |     sprintf(modName, "rutherford%i", m + 1);
      |                                 ^~
/tmp/jenkins-FairRootGroup-FairSoft-PR-314-1.zup/env_post_workdir.PmH/FairRoot/examples/simulation/rutherford/src/FairRutherfordGeo.cxx:32:22: note: directive argument in the range [-2147483647, 2147483647]
   32 |     sprintf(modName, "rutherford%i", m + 1);
      |                      ^~~~~~~~~~~~~~
In file included from /usr/include/stdio.h:867,
                 from /tmp/jenkins-FairRootGroup-FairSoft-PR-314-1.zup/install-tree/linux-ubuntu20.04-nehalem/gcc-9.3.0/root-6.18.04-bubxyy25pi6mqrxu63r5t44qmyggg4h3/include/snprintf.h:14,
                 from /tmp/jenkins-FairRootGroup-FairSoft-PR-314-1.zup/install-tree/linux-ubuntu20.04-nehalem/gcc-9.3.0/root-6.18.04-bubxyy25pi6mqrxu63r5t44qmyggg4h3/include/Rtypes.h:26,
                 from /tmp/jenkins-FairRootGroup-FairSoft-PR-314-1.zup/env_post_workdir.PmH/FairRoot/geobase/FairGeoBasicShape.h:11,
[...]

PR #1083, opened 2021-07-23
Fix Warnings: sprintf and missing override

ChristianTackeGSI:
* Use `override` in methods that actually override base functions.

* Increase string buffers to have enough space for 32bit integeers, and use snprintf so to not overwrite any buffer boundaries.

Should fix: #1005

---

Checklist:


ChristianTackeGSI:
@dennisklein BTW should this be backported to 18.x?

dennisklein (approved):
The C-style string manipulation is not very pretty, but I guess changing that is for another day..

dennisklein:
> @dennisklein BTW should this be backported to 18.x?

As you wish.

Issue #1085, opened 2021-07-24
FairRootConfig.sh minor issues


klenze:
I just did a fresh make install on the dev version of FairRoot, and the configuration generated by FairRootConfig.sh leaves some room for improvement, in my opinion. 

**To Reproduce**
```bash
git clone ...
mkdir fairroot_build && cd fairroot_build
cmake -DCMAKE_TOOLCHAIN_FILE=$PWD/../toolchain.cmake -DCMAKE_INSTALL_PREFIX=$PWD/../fairroot_inst/ \
  ../FairRoot -DCMAKE_FIND_USE_PACKAGE_REGISTRY=0
make -j install
cat ../fairroot_inst/bin/FairRootConfig.sh
```

Results (reordered, dubious only, #comments are mine)
```bash
#empty env variables:
export ROOTSYS="" #empty
export FAIRROOTPATH="" #also empty
export G3SYS="" # but Geant3_... stuff is set?
# set twice:
export Geant3_INCLUDE_DIRS="/path/to/fairsoft_inst/include/TGeant3"
#other (won't argue about it):
export VMCWORKDIR="/path/to/fairroot_inst/share/fairbase/examples" 
#overwritten general env:
export CLASSPATH="" #never liked java anyhow
export PYTHONPATH="/path/to/git/repo/of/FairRoot/python:[...]" # does not actually exist
case $1 in
#[...]
* ) #$1 is neither -a nor -p
   export DYLD_LIBRARY_PATH="" #some apple thing?
   LD_LIBRARY_PATH="/path/to/fairroot_inst/lib:/path/to/fairsoft_inst/lib" # hope there was nothing important in before
   export PATH="/bin" # at least it is not empty
;;
```

Some remarks. 
None of this is very serious, figuring out that ROOTSYS should be /path/to/fairsoft_inst/ and FAIRROOTPATH should be /path/to/fairroot_inst/ is rather obvious.
In R3BRoot, I have always used VMCWORKDIR as "that variable helpfully pointing to my git repository", but that is probably not what it is for?
Overwriting CLASSPATH and PYTHONPATH unconditionally could actually annoy users of java or python who rely on it. Any variables not invented by FairRoot or FairSoft packages (e.g. ROOT) should go into the case block, in my opinion.

Speaking of that case block, I think it nice to give the users a prepend and append option. However, I think "replace" is a terrible default.
[...]

PR #1086, opened 2021-07-26
Drop USE_PATH_INFO CMake option

ChristianTackeGSI:
Baking the PATH and LD_LIBRARY_PATH values at configure time into the install tree is no good idea, really.

See also: #1085

---

Checklist:


fuhlig1 (approved):
Hi @ChristianTackeGSI,

the proposed changes are fine with me. 

This is old legacy code. Since we had in ancient times huge problems with messed up environments we implemented the config.sh script in a way that we removed anything from the PATH and LD_LIBRRAY_PATH. This was brute force but helped us a lot. In some environments this was to strict such that we allowed to overwrite the default setting. Meanwhile this isn't needed any longer since the config.sh script has the append and prepend options which do exactly the same in a much better way.  

Please go ahead.

ChristianTackeGSI:
Thanks @fuhlig1 ! :-)

PR #1089, opened 2021-08-08
Drop Unrelated Variables from FairRootConfig.sh

ChristianTackeGSI:
FairRootConfig.sh used to unconditionally override a few variables that are completely unrelated to FairRoot.
They were just set to values passed to cmake at configure time.

If you really need to set those in your use case, you might create YourUseCase.sh, set the variables there directly, and call FairRootConfig.sh.

Removed:
- `CLASSPATH` (`$NEW_CLASSPATH`)
- `PLUTO_INCLUDE_DIR` (`-DPLUTO_INCLUDE_DIR=foo`)
- `PLUTO_LIBRARY_DIR` (`-DPLUTO_LIBRARY_DIR=foo`)

Should this be documented in the ChangeLog?

Fixes part of #1085 by removing `CLASSPATH` setting.

---

Checklist:


dennisklein (approved):
Looks good from my perspective, but better to also get approval by @MohammadAlTurany or @fuhlig1.

dennisklein:
Note to self:
* [ ] Backport to v18.6

fuhlig1 (commented):
To remove the pluto related variables is fine with me. This is not used in CbmRoot and I think no other of the experiments use it.

About CLASSPATH I am not 100% sure. It is not used in CbmRoot but I think the variable was added for one of the other experiments. 

Since each experiment can add his one script which then calls config.sh this should also be no problem.

PR #1090, opened 2021-08-12
FairParRootFileIo merging: race condition fixed.

klenze:
(Disclaimer: Philosophically, I think that using custom container classes serialized to *.root files to store calibration/slow control parameters is a terrible idea for a host of reasons, chief among them the lack of interoperability, limitations of version control for binary files and the fact that modern operating systems such as Bell Labs' UNIX(R) feature so-called " hierarchical file systems", which happen to duplicate innovations from TFile such as key-value storage of information, directory hierarchies etc while also being vastly superior to TFiles in terms of accessibility and features. Thus, the following PR should not be mistaken for an endorsement of that design choice >:->  )

FairParRootFileIo::open(const TList*, ...) is broken. Calling it will generate a file named allParams_$DATE_$TIME.root in the directory of the first input file, mostly for internal purposes.

This is bad for the following reasons:
* Implicitly creating junk files in directories with useful data is considered bad (Citation needed)
* Write permissions might not be present for that directory
* the suffix $DATE_$TIME creates horrible race conditions as soon as two or more processes are started at the same time

If one needs to create a temporary file, the generally accepted procedure in the *nix  world seems to be to create it in tmp (or /var/tmp), and to use a race-avoiding function of the mkstemp family. (Of course, ROOT being ROOT, passing that file on with the "recreate" option will generate another race between the unlink and the subsequent open call, but at least the result from mkstemp contains enough entropy to make an actual collision unlikely.)

As there is no requirement to wait for the first list item to copy the directory name, this also allows to create the temporary file unconditionally, thus simplifying the process flow somewhat. 

Another option would be to create the file in memory (per TFile default constructor, perhaps) and pass on the TFile* instead of the name, forcing people to explicitly write out the file if they want the merged parameter set for some other reason. 

(From my limited understanding of how TFiles operate, it seems likely that the code in that function will probably not handle TFiles containing directories correctly. Also, there is no check for duplicate object names being created.)

### DIFF
--- a/parbase/FairParRootFileIo.cxx
+++ b/parbase/FairParRootFileIo.cxx
@@ -37,6 +37,10 @@
 #include <TString.h>       // for TString, Form
 #include <iostream>        // for operator<<, basic_ostream, etc
 
+#include <stdlib.h> // free, mkstemps
+#include <unistd.h> // close
+#include <string.h> // strncpy
+
 using std::cerr;
 using std::cout;
 using std::endl;
@@ -147,31 +151,29 @@ Bool_t FairParRootFileIo::open(const Text_t* fname, Option_t* option, const Text
 
 Bool_t FairParRootFileIo::open(const TList* fnamelist, Option_t* option, const Text_t* ftitle, Int_t compress)
 {
-    TDatime currentDate;
-    TString newParFileName = "";
-    TFile* newParFile = 0;
-
+    char fname[100];
+    strncpy(fname, "/tmp/allParams_XXXXXX.root", 100);
+    int fd=mkstemps(fname, 5);
+    if (fd==-1)
+      throw std::runtime_error("mkstemps failed!");
+    ::close(fd);
+    TFile* newParFile = TFile::Open(fname, "RECREATE");
+    // if the above call is subject to race conditions, there is little we can do about it.
+    if (!newParFile)
+      throw std::runtime_error(std::string("TFile::Open(\"")
+                               +fname
+                               +"\", \"RECREATE\") failed.");
     TObjString* string;
+    // string is a horrible name for a variable :-/
     TListIter myIter(fnamelist);
-
     TKey* inpKey;
-
     TFile* inFile;
 
-    Int_t nofFiles = 0;
     while ((string = static_cast<TObjString*>(myIter.Next()))) {
         inFile = TFile::Open(string->GetString().Data());
         if (!inFile) {
             cout << "-W- File \"" << string->GetString().Data() << "\" does not exist" << endl;
             continue;
-        }
-
-        if (nofFiles == 0) {
-            newParFileName = string->GetString();
-            newParFileName.Replace(newParFileName.Last('/') + 1, newParFileName.Length(), "");
-            newParFileName =
-                Form("%sallParams_%d_%06d.root", newParFileName.Data(), currentDate.GetDate(), currentDate.GetTime());
-            newParFile = TFile::Open(newParFileName.Data(), "RECREATE");
         }
 
         TList* inputKeys = static_cast<TList*>(inFile->GetListOfKeys());
@@ -184,17 +186,9 @@ Bool_t FairParRootFileIo::open(const TList* fnamelist, Option_t* option, const T
             tempObj->Write();
         }
         inFile->Close();
-
-        nofFiles++;
     }
-    if (newParFile != 0) {
-        newParFile->Close();
-    } else {
-        std::cout << "****NO file to close file = \"" << std::endl;
-    }
-    std::cout << "**** merged file = \"" << newParFileName.Data() << "\"" << std::endl;
-
-    return this->open(newParFileName, option, ftitle, compress);
+    std::cout << "**** merged file = \"" << fname << "\"" << std::endl;    
+    return this->open(fname, option, ftitle, compress);
 }
 
 Bool_t FairParRootFileIo::open(TFile* f)


ChristianTackeGSI:
Hi @klenze !

Thanks for your patch!

And many thanks for your nice writing! We loved to read it!

Let's get on topic.

First: You're fully right!

Some notes:
* Can you take a look at the output of the format checks? [here](https://alfa-ci.gsi.de/blue/organizations/jenkins/FairRootGroup%2FFairRoot/detail/PR-1090/1/pipeline)
* Did you try ordering the `::close()` after the `TFile::Open()`? This at least gives a better feeling that the file should still be held by the process. Although it's not really required, as the file is first created by `mkstemps()` and not deleted on close.
* > mkstemps(): unstandardized, but appears on several other systems.

  (from my local manpage)
  I hope it also works on macos.
  I guess, we do not have any unit tests for this anyway.

That said, the whole thing looks good to me!

parbase/FairParRootFileIo.cxx
```diff
@@ -147,33 +151,31 @@ Bool_t FairParRootFileIo::open(const Text_t* fname, Option_t* option, const Text
 
 Bool_t FairParRootFileIo::open(const TList* fnamelist, Option_t* option, const Text_t* ftitle, Int_t compress)
 {
-    TDatime currentDate;
-    TString newParFileName = "";
-    TFile* newParFile = 0;
-
+    char fname[100];
```
dennisklein, line 154: Have you considered using `std::array<char, 100> fname`? see [ES.27](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es27-use-stdarray-or-stack_array-for-arrays-on-the-stack) and [SL.con.1](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slcon1-prefer-using-stl-array-or-vector-instead-of-a-c-array)

parbase/FairParRootFileIo.cxx
```diff
@@ -147,33 +151,31 @@ Bool_t FairParRootFileIo::open(const Text_t* fname, Option_t* option, const Text
...
+    strncpy(fname, "/tmp/allParams_XXXXXX.root", 100);
+    int fd=mkstemps(fname, 5);
+    if (fd==-1)
+      throw std::runtime_error("mkstemps failed!");
+    ::close(fd);
+    TFile* newParFile = TFile::Open(fname, "RECREATE");
```
dennisklein, line 160: Thinking about the race condition you point out, is there actually any need to have a filesystem handle for this `allParams...` file at all (also who is cleaning up such files?)? What about using a [`TMemFile`](https://root.cern/doc/master/classTMemFile.html) and avoid using the filesystem alltogether (I am not too familiar with the `FairParRootFileIo`)?

dennisklein (commented):
See my inline comments. Regarding the proposed C usage, I would prefer if we stick to C++ where feasible.

parbase/FairParRootFileIo.cxx
```diff
@@ -37,6 +37,10 @@
 #include <TString.h>       // for TString, Form
 #include <iostream>        // for operator<<, basic_ostream, etc
 
+#include <stdlib.h> // free, mkstemps
+#include <unistd.h> // close
+#include <string.h> // strncpy
```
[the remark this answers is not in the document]
ChristianTackeGSI: `std::tmpfile` seems to only return a `FILE *`. Which raises the question, whether `TFile` can be instantiated from a `FILE *`. And this open file does not have a name. So if anything later on needs a filename, that needs fixing as well.

parbase/FairParRootFileIo.cxx
ChristianTackeGSI: It looks like `FairRunAnaProof::RunOnProof` actually uses the filename (via `::getFilename()`) for something. And from a quick looks, this *seems* to make its way to `FairAnaSelector::Init`, where it looks like it is used for actually opening the file (again).

Might all be bogus and not used by anybody / not happen in the real world. But please take this into account.

dennisklein: Then I get the impression this generated *allParams* root file is not really something that should be dumped to `/tmp` tbh

ChristianTackeGSI:
Yet another note:

* `/tmp` is usually a local filesystem. The directory of the first given file might be on a shared filesystem.
* Just learned, that the (generated) filename *might* make its way from `FairRunAnaProof::RunOnProof` to `FairAnaSelector::Init`
* So if the (generated) filename travels across machines, then the file will not be accessible on the "target" machine any longer.
* I have no clue, whether this is a real problem or just academic.

Suggestion:
* Let's deprecate the current function.
* Let's create a utility function that merges ROOT files into one.
  * The target directory could be given. If it's not given, then default to `${TMPDIR:-/tmp}`.
  * Returns the generated filename.
  * Yes, this is not a nice public API, but parbase dictetes some of this.
* That way people have the choice:
  * Use the old (deprecated,) racy, junk creating API
  * Use the new utility before opening, to create a file in `/tmp`
    * Letting the OS clean it up later on, or being a nice citizen and cleaning it up after the run / reading the parameters from it
  * Create it in a known location, taking care of cleaning it up.

dennisklein:
Update: As it is now, I do not think we can merge this. But we have started a group internal effort to address this and other issues within the parbase module. So, we can leave the PR open for reference for now.

ChristianTackeGSI:
Note that @klenze said [here](https://github.com/FairRootGroup/FairRoot/pull/1156#issuecomment-1072906171) that we can close this PR (#1090) when the other one (#1156) gets merged.

PR #1096, opened 2021-09-01
fairtools: Create stacktrace file only if 'Root.Stacktrace: yes'

dennisklein:
resolve #1078

### DIFF
--- a/fairtools/FairLogger.cxx
+++ b/fairtools/FairLogger.cxx
@@ -14,6 +14,7 @@
 
 #include "FairLogger.h"
 
+#include <TEnv.h>      // for gEnv
 #include <TString.h>   // for TString, operator==, etc
 #include <TSystem.h>   // for gSystem, TSystem
 #include <iostream>    // for cout, cerr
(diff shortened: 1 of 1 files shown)


fuhlig1 (commented):
Should be okay. Somehow we need to inform users about the change.

dennisklein:
> Somehow we need to inform users about the change.

I think no notification is need, the behaviour is

**before**
*  `Root.Stacktrace: yes` (default): FairRoot creates the `core_dump_<pid>` file containing a stacktrace
*  `Root.Stacktrace: no`: FairRoot creates an empty `core_dump_<pid>` file

**after**
* `Root.Stacktrace: yes` (default): FairRoot creates the `core_dump_<pid>` file containing a stacktrace
* `Root.Stacktrace: no`: FairRoot does not create a file

The call to `gSystem->StackTrace()` is already honoring the `Root.Stacktrace` setting. So this PR just additionally does not create empty files.

dennisklein:
but true, I will add a changelog entry. and if we do a backport it will show up in the changelog of the patch release.

rbx (approved):

PR #1097, opened 2021-09-06
base: Remove 'requested branch name' instrumentation

### LINKED ISSUE #1075: Problems with links (FairRootManager, FairRunAnaProof)
While browsing our MpdRoot gitlab repository I noticed a complaint of memory leaks from years ago. I have personally had this issue, so I began testing in order to solve it. 
The relatively old versions of FairSoft/Root (v18.2_patches)  is default for now for us but that is supposedly soon to change . For about a week of diagnosis we simplified our reconstruction chain to only include a single HitProducer by a suggestion from our tracking expert.  
This HitProducer ... produces hits and then links them. 
  It uses the TClonesArray new placement and Delete to free them. 
 (I tested ConstructedAt with a custom Clear for the hits - It did not help for memory, somewhat improved time)
   
I experimented with 500 pions/event, 1000 events produced 1.7 GiB with fRun->SetUseFairLinks(true);   
Setting fRun->SetUseFairLinks(false) results in 321 MiB. Commenting the links in the producer - the same.  
If we make this a bit more extreme 10k events costed 14 GiB with UseFairLinks and 321 without.   
  
Then I tried to run in proof-mode (using FairRunAnaProof), that was running, however it did not produce any links. Memory was low, and since there are no Links there are no leaks.   
  
We suspected first our code. Tried all types of Delete(), Clear("C") with a MpdTpcHit::Clear which resets links, then  we moved to virtual destructors on everything FairMultiLinkedData /_interface, FairHit - Nothing...

So then I saw there was a bug there, which was supposedly fixed in FairRoot v18.4 - no dice.  
Installed FairSoft apr21 and one  of latest FairRoot devs/master (not sure) - same issue.
  
I forgot to mention during this process I ran various profiling tools.
 Valgrind's memcheck, callgrind, but massif was the most useful in this case.  
image
  
Finally, after a week of poking at this, commenting this ```    fReqBrNames.emplace_back(BrName);``` at line 523 in FairRootManager::GetObject  seems to fix it.  This is getting called from FairMultiLinkedData::InsertHistory. 
In this context I am not sure what it is doing, I mean why do we save the branch name? Why doesn't this list get cleared?   
Anyway. Recompile. Code is purring. Does not pass 300 MiB. When I include our FillDstTask rises to 577 MiB due to loading other branches, but that is it.

Still, FairRunAnaProof does not have links, and since our code depends on them it is not possible to use it. GNU/parallel is the alternative of my choice.
Also, FairRunAnaProof does not merge the resulting worker files at the end. It produces a single empty file.

Cheers!

Edit:
Forgot to mention that these are the ways we do linking:
```pHit->AddLink(FairLink(MpdTpcHit::PointIndex, pointIndx));  ```
```pHit->AddLink(FairLink(MpdTpcHit::MCTrackIndex, trackIndx));```
Trying the long-style (with branch name) did not seem to fix it for me, maybe I did not try properly.

### PR DESCRIPTION
**Pull request by dennisklein**
fixes #1075

### DIFF
--- a/base/steer/FairRootManager.h
+++ b/base/steer/FairRootManager.h
@@ -90,8 +90,6 @@ class FairRootManager : public TObject
 
     /**Return a TList of TObjString of branch names available in this session*/
     TList* GetBranchNameList() { return fBranchNameList; }
-    /**Return the vector of branch names that were requested by tasks as input*/
-    const std::vector<std::string>& GetReqBranchNames() const { return fReqBrNames; }
 
     /**  Get the Object (container) for the given branch name,
          this method can be used to access the data of
@@ -363,8 +361,6 @@ class FairRootManager : public TObject
     Int_t fBranchSeqId;
     /**List of branch names as TObjString*/
     TList* fBranchNameList;   //!
-    /**Vector of (not necessarily unique) branch names requested per GetObject / InitObjectAs */
-    std::vector<std::string> fReqBrNames;   //!
 
     /**The branch ID for the special (required) MCTrack branch**/
     Int_t fMCTrackBranchId;   //!
@@ -421,7 +417,7 @@ class FairRootManager : public TObject
     // data members
     Int_t fId;   // This manager ID
 
-    ClassDef(FairRootManager, 12);
+    ClassDef(FairRootManager, 13);
 };
 
 // FIXME: move to source since we can make it non-template dependent
@@ -506,9 +502,6 @@ TPtr FairRootManager::InitObjectAs(const char* brname)
     }
     // add into branch list
     AddMemoryBranchAny<T>(brname, addr);
-    // register as a **requested** branch
-    // (duplications are explicitely allowed)
-    fReqBrNames.emplace_back(brname);
 
     // NOTE: ideally we would do proper resource management for addr and *addr
     // since the FairRootManager becomes owner of these pointers/instances; Unfortunately this
--- a/base/steer/FairRootManager.cxx
+++ b/base/steer/FairRootManager.cxx
@@ -520,7 +520,6 @@ Bool_t FairRootManager::ReadNextEvent(Double_t)
 
 TObject* FairRootManager::GetObject(const char* BrName)
 {
-    fReqBrNames.emplace_back(BrName);
     /**Get Data object by name*/
     TObject* Obj = nullptr;
     LOG(debug2) << " Try to find if the object " << BrName << " is already activated by another task or call";

### CODE REVIEW

File: base/steer/FairRootManager.h [hunk 1]
```diff
@@ -421,7 +417,7 @@ class FairRootManager : public TObject
     // data members
     Int_t fId;   // This manager ID
 
-    ClassDef(FairRootManager, 12);
+    ClassDef(FairRootManager, 13);
```
ChristianTackeGSI, line 420: This inherits from `TObject`. Can you please use `ClassDefOverride` here, if you touch this line anyway?

fuhlig1: Why this should be changed? We don't use the override keyword in the class. According to the ROOT team this macro is needed when this is used

> ClassDefOverride: Use this one if your class definition use the c++ `override` keyword for virtual functions (If you use the regular ClassDef, the compiler might warn that the function declared in the ClassDef are missing the override keyword).

dennisklein: After [looking it up](https://github.com/root-project/root/blob/0e0154b4960c92ae62df652b5b21f7d24a1fab1d/core/base/inc/Rtypes.h#L309), I agree with @fuhlig1. The difference between `ClassDef` and `ClassDefOverride` seems to be that the generated code literally uses `virtual` or `override` keywords (mutually exclusive).

> ClassDefOverride: Use this one if your class definition use the c++ override keyword for virtual functions (If you use the regular ClassDef, the compiler might warn that the function declared in the ClassDef are missing the override keyword).

So the compiler warning this description refers to, must be https://clang.llvm.org/docs/DiagnosticsReference.html#winconsistent-missing-override ?!

dennisklein: Do we have consensus on using `ClassDefOverride` once we modernize the `FairRootManager` class to use the `override` keyword? @ChristianTackeGSI Or do you have a different interpretation of when this macro shall be used?

ChristianTackeGSI: That sounds fine!

If we switch to "the override" world (make things `-Wsuggest-override`-clean), then we should also switch to `ClassDefOverride`.

As far, as I can tell, there is only one other place in this class that would get `override`:
```diff
-    Int_t Write(const char* name = 0, Int_t option = 0, Int_t bufsize = 0);
+    Int_t Write(const char* name = 0, Int_t option = 0, Int_t bufsize = 0) override;
```

All that said: This is only a suggestion. Do whatever fits the FairRoot world best!

PR #1104, opened 2021-09-14
parbase: Do not Leak TFile Objects

ChristianTackeGSI:
`TFile::Open()` returns a `TFile*` and expects the caller to `delete` the object after using it.
Use `std::unique_ptr` to handle this.

---

Checklist:

### DIFF
--- a/parbase/FairParRootFileIo.cxx
+++ b/parbase/FairParRootFileIo.cxx
@@ -36,6 +36,7 @@
 #include <TObject.h>       // for TObject
 #include <TString.h>       // for TString, Form
 #include <iostream>        // for operator<<, basic_ostream, etc
+#include <memory>          // for std::unique_ptr
 
 using std::cerr;
 using std::cout;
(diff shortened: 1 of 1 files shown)


dennisklein (changes requested):
Looks good, please add the missing `#include <memory>` and this is a go

parbase/FairParRootFileIo.cxx
```diff
@@ -149,49 +149,41 @@ Bool_t FairParRootFileIo::open(const TList* fnamelist, Option_t* option, const T
...
 
         TListIter keyIter(inputKeys);
+        TKey* inpKey;
         while ((inpKey = static_cast<TKey*>(keyIter.Next()))) {
             TObject* tempObj = inFile->Get(inpKey->GetName());
 
             newParFile->cd();
             tempObj->Write();
         }
-        inFile->Close();
```
fuhlig1, line 186: Why the files are not closed any longer? Currently this is done in the destructor of TFile but it this something one should rely on?

ChristianTackeGSI: I'd say, that this is the whole purpose of a destructor?

The only real problem I can see: `Close` is a virtual method. Calling a virtual method from a destructor is highly not reommended / possibly doesn't do what you would expect it to do.

But then again: Why do we have a destructor, when it is not cleaning up properly?

@dennisklein any thoughts on this?

dennisklein: My first thought (before @fuhlig1's question) was to absolutely rely on TFile object destruction to also close the file. But now I am puzzled

@fuhlig1 Can you concretize your concern?

dennisklein (approved):
thx for adding the header

parbase/FairParRootFileIo.cxx
fuhlig1: Since the file is properly closed in the destructor I am fine with the commit.

PR #1105, opened 2021-09-14
fairtools: Add yaml vmc config option "UseFastSim"

dennisklein:
Open questions:
* Is fast simulation Geant4 specific? Or should we call the option rather generically just `UseFastSim`?
* Is the default value of `true` a good default?

### DIFF
--- a/fairtools/MCConfigurator/FairYamlVMCConfig.cxx
+++ b/fairtools/MCConfigurator/FairYamlVMCConfig.cxx
@@ -27,6 +27,7 @@
 #include <TVirtualMC.h>   // for TVirtualMC
 #include <cstdlib>        // for getenv
 #include <cstring>        // for strcmp, strncmp
+#include <memory>         // for unique_ptr
 #include <ostream>        // for operator<<, ostringstream
 #include <string>         // for string, basic_string, cha...
 #include <vector>         // for vector
@@ -165,12 +166,18 @@ void FairYamlVMCConfig::SetupGeant4()
     if (!fYamlConfig["Geant4_UserGeometry"]) {
         LOG(fatal) << "User geometry not provided";
     }
+    auto const g4UserGeometry(fYamlConfig["Geant4_UserGeometry"].as<std::string>());
+
     if (!fYamlConfig["Geant4_PhysicsList"]) {
         LOG(fatal) << "Physics list not provided";
     }
+    auto const g4PhysicsList(fYamlConfig["Geant4_PhysicsList"].as<std::string>());
+
     if (!fYamlConfig["Geant4_SpecialProcess"]) {
         LOG(fatal) << "Special processy not provided";
     }
+    auto const g4SpecialProcess(fYamlConfig["Geant4_SpecialProcess"].as<std::string>());
+
     bool specialStacking = false;
     if (fYamlConfig["Geant4_SpecialStacking"]) {
         //        LOG(info) << "Special stacking used";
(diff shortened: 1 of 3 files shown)


ChristianTackeGSI (commented):
Looks good to me. I can't judge, whether this is actually doing "The Right Thing(TM)".

karabowi:
The FastSim option is only available in Geant4.

karabowi (approved):

dennisklein:
In the group meeting on 20.9. we (@karabowi, @TobiasStockmanns, and myself) agreed to name the option just `UseFastSim` in foresight to future fastsim implementations (currently FairRoot only implements this option for Geant4). The default value shall be `false`.

fairtools/MCConfigurator/FairYamlVMCConfig.cxx
```diff
@@ -183,22 +190,25 @@ void FairYamlVMCConfig::SetupGeant4()
...
-                                        mtMode);
+    auto const useFastSim(fYamlConfig["UseFastSim"] ? fYamlConfig["UseFastSim"].as<bool>() : fUseFastSimDefault);
+    std::unique_ptr<TG4RunConfiguration> runConfiguration;
+    if (useFastSim) {
+        runConfiguration = std::make_unique<FairFastSimRunConfiguration>(
+            g4UserGeometry, g4PhysicsList, g4SpecialProcess, specialStacking, mtMode);
+    } else {
+        runConfiguration = std::make_unique<TG4RunConfiguration>(
+            g4UserGeometry, g4PhysicsList, g4SpecialProcess, specialStacking, mtMode);
+    }
```
ChristianTackeGSI, line 201: `std::make_unique` requires C++14. FairRoot currently is saying, that it needs C++11.

Please either bump to C++14 (fine with me, I would like to add some `[[deprecated]]`).

Or rewrite using `.reset(new …)`.

fairtools/MCConfigurator/FairYamlVMCConfig.cxx
```diff
@@ -183,22 +190,25 @@ void FairYamlVMCConfig::SetupGeant4()
...
 
-    TGeant4* geant4 = new TGeant4("TGeant4", "The Geant4 Monte Carlo", runConfiguration);
+    TGeant4* geant4 = new TGeant4("TGeant4", "The Geant4 Monte Carlo", runConfiguration.release());
```
ChristianTackeGSI, line 203: Just wondering: Where is this pointer (ownership) going after the method ends?

(Not the target of this PR, I know!)

fairtools/MCConfigurator/FairYamlVMCConfig.cxx
dennisklein: Thx, I will address this!

fairtools/MCConfigurator/FairYamlVMCConfig.cxx
dennisklein: https://github.com/vmc-project/geant4_vmc/blob/f6ced3fa4a5253848186c466b4865f37e75d223e/source/run/src/TGeant4.cxx#L165-L169

I guess, one would call this a "conditional singleton"? TBH, since a few years, when I got aware of this *pattern*, which is btw also used throughout the FairRoot steering classes, I could not find a good idea yet how to fix this situation.

ChristianTackeGSI: Thanks for the explanation!

Fixing this without a breaking change is really, really non-trivial. I agree.

dennisklein: The user interface of this singleton implementation is also always confusing me. Because it sort of allows you to construct the object, but implicitly requires you to instantiate as a heap object. And it also requires you to first call a constructor before you get what you want from calling [`TGeant4* TGeant4::MasterInstance()`](https://github.com/vmc-project/geant4_vmc/blob/f6ced3fa4a5253848186c466b4865f37e75d223e/source/run/include/TGeant4.icc#L26). I have also seen macros with just a single line `new FairRootManager;` which completely blew my mind.

ChristianTackeGSI: I know, this is not the target of this PR!

Could you consider adding a comment above this line? Something like
```c++
// Instantiate a singleton like object, "leaking" it is the current API
```
Rephrase / Modify to your liking!

Just something, so that people like me now that this is not a leak.

P.S.: I am not going to ask, whether FairRoot should call `TGeant4::MasterInstance()` to make sure that the singleton like thing isn't already initialized.

dennisklein: done

fairtools/MCConfigurator/FairYamlVMCConfig.cxx
dennisklein: done

dennisklein:
@karabowi Note, that this now changes the default back to false. That means that our examples/unit tests run again **without fast sim**. Is this ok? Or should I enable fast sim in a particular example/test. (I have not checked if this even makes any difference)

ChristianTackeGSI (approved):
All my points have been addressed. Things look good. :)

karabowi (approved):
Looks great, found a small bug in my code, but will commit a separate fix.

PR #1106, opened 2021-09-15
Only Support Prepend in FairRootConfig.sh

ChristianTackeGSI:
(As discussed in last group meeting - see https://indico.gsi.de/event/13175/)

Overriding PATH-like environment variables with information at configure time requires users to re-add their local directories after sourcing FairRootConfig.sh.

Appending has no good use case at all. It only adds a fallback. This is normally not, what you want.

Which leaves us with prepend logic.

If your use case (say in your ExperimentRoot) requires hardwiring PATH/etc, you can create ExperimentRootConfig.sh, preset PATH/etc and inside that script call FairRootConfig.sh.

It doesn't look like `set_env_macos.sh.in` is used inside FairRoot at all. So drop it.

See-Also: #1085
See-Also: ef405a5c5c6514db05c552134a71267f6f8177ab

---

Checklist:


fuhlig1 (approved):

dennisklein:
We discussed this topic once more last group meeting (20.9., developers present: @TobiasStockmanns, @karabowi, @ChristianTackeGSI, and myself - @fuhlig1 approved this PR even before) and concluded that not having an `--override` option as proposed in this PR (and contrary to the decision in the group meeting on 13.9.) will be fine, therefore merging.

PR #1112, opened 2021-09-29
Tutorial3: Use Fixtures Instead of Depends

ChristianTackeGSI:
In CMake tests `DEPENDS` is only asking for proper ordering. That is "A depends on B" only means that B runs before A, but nothing more.

if A should only start if B was successful, one needs fixtures.  B will declare a "setup fixture" and A then lists this fixture as required.

---

Checklist:


fuhlig1 (dismissed):
At the time when the tests were implemented the fixture feature wasn't available and the best we could do at that time was to run the tests sequentiell were this was needed.
Unfortunately the tests weren't reworked after the feature was available.

dennisklein (approved):
Just a headsup: This kind of improvement would also apply for some of the other examples.

PR #1123, opened 2021-12-06
Move to VMC standalone as default

benedikt-voelkel:
* ROOT's builtin VMC will be deprecated as of version v6.26
  ==> use standalone as default and fallback to ROOT's builtin if
      possible

* all new VMC tags
  * G3 3-9
  * G4_VMC v5-4
  * FLUKA_VMC 4-1.1-vmc2
  moved already to standalone as default

Describe your proposal.

Mention any issue this PR is resolves or is related to.

---

Checklist:


benedikt-voelkel:
@MohammadAlTurany 
I built FairRoot also against VMC standalone already, also with all VMC interfaces linked against standalone VMC.

MohammadAlTurany (approved):

dennisklein (commented):
If the found ROOT was compiled with builtin VMC, VMC headers will bleed in via ROOT's include directories. So, the only correct logic is as it is already: "Take standalone VMC only, if built-in VMC is not found."

benedikt-voelkel:
Dear @dennisklein 
I reckon you are right. Indeed, that is really unfortunate. So in that case, no change is necessary and I think we can close this PR.

ihrivnac:
> If the found ROOT was compiled with builtin VMC, VMC headers will bleed in via ROOT's include directories. So, the only correct logic is as it is already: "Take standalone VMC only, if built-in VMC is not found."

@dennisklein , you are corect, but we wanted to pass a clear message when vmc-in-root is used that this will be removed in future. See below the code put in the last updates in geant3 and geant4_vmc:

```
#-- VMC (required) ------------------------------------------------------------
find_package(VMC CONFIG)
# first try VMC standalone (default)
if(VMC_FOUND)
  if(NOT VMC_FIND_QUIETLY)
    message(STATUS "Found VMC ${VMC_VERSION} in ${VMC_DIR}")
  endif()
else()
  # otherwise fallback to ROOT's internal if possible (deprecated)
  if(ROOT_vmc_FOUND)
    message(WARNING "Using VMC built with ROOT - Deprecated")
    set(VMC_LIBRARIES ROOT::VMC)
    add_definitions(-DUSE_ROOT_VMC)
  else()
    message(FATAL_ERROR
            "Could not find VMC package. "
            "VMC package is not provided with ROOT since v6.26/00. "
            "Please, install it from https://github.com/vmc-project/vmc. ")
  endif()
endif()

```

I think that instead of going back to the previous version when VMC config was using only if  vmc-in-root is not found, we can add an error message just after if(VMC_FOUND) to prevent from mixing two versions:

```
if(VMC_FOUND)
   # cannot mix VMC standalone and vmc in ROOT(deprecated)
  if(ROOT_vmc_FOUND)
     message(FATAL_ERROR
           "Cannot use VMC standalone with ROOT built with vmc.")
   endif()
   if(NOT VMC_FIND_QUIETLY)
      message(STATUS "Found VMC ${VMC_VERSION} in ${VMC_DIR}")
   endif()
else()
  ...
```

dennisklein:
> I think that instead of going back to the previous version when VMC config was using only if vmc-in-root is not found, we can > add an error message just after if(VMC_FOUND) to prevent from mixing two versions:
>
> ```
> if(VMC_FOUND)
>   # cannot mix VMC standalone and vmc in ROOT(deprecated)
>  if(ROOT_vmc_FOUND)
>     message(FATAL_ERROR
>           "Cannot use VMC standalone with ROOT built with vmc.")
>   endif()
> ```

Hmm, FairRoot never sees the condition `VMC_FOUND AND ROOT_vmc_FOUND` to become true. To be able to check this, FairRoot would always have to call `find_package(VMC)` which - I am afraid - I am not convinced (yet) is what it should do. (The general idea is also to move away from optional `find_package` invocations and fail fast by using the `REQUIRED` option as it is already done in this case and which we would have to remove to check the proposed condition, right?)

@ChristianTackeGSI What do you think?

dennisklein:
> To be able to check this, FairRoot would always have to call find_package(VMC) which - I am afraid - I am not convinced (yet) is what it should do.

However, in the future, if FairRoot requires a minimum ROOT version that does not have built-in VMC any more, we would of course apply the following change, e.g.:

```diff
find_package2(PUBLIC ROOT VERSION 6.26.00 REQUIRED)
- if(ROOT_vmc_FOUND)
-   set(VMCLIB ROOT::VMC)
- else()
-   find_package2(PUBLIC VMC REQUIRED)
+ find_package2(PUBLIC VMC REQUIRED) # and then use VMCLibrary target name everywhere directly instead of variable VMCLIB
-   set(VMCLIB VMCLibrary)
- endif()
```

dennisklein:
@ihrivnac tbh, I'd rather see this kind of a sanity check located in VMC itself instead of all its dependees. VMC can easily and meaningfully fail its configure phase if `ROOT_vmc_FOUND` is true, right?

ihrivnac:
@dennisklein , my comment above was a suggestion how to update geant3 && geant4_vmc cmake, to fix a problem that you pointed above.

In FairRoot, the code can stay as it is now, and when you move to ROOT v6.26/00 you add the required to VMC find as you suggest.

benedikt-voelkel:
Thank you @ihrivnac for taking care of that!

ihrivnac:
The update in vmc is proposed.  Please, check. If everything works fine, I will put a patch tag 1.1.p1 including this fix. Then, no update will be needed in geant3, geant4_vmc and FairRoot. Please, confirm, and if you agree, this PR can be closed.

dennisklein:
The proposed sanity check has been implemented in VMC directly (https://github.com/vmc-project/vmc/pull/13), making this PR obsolete, therefor closing. Thanks to all!

PR #1128, opened 2022-01-11
Add missing example dependency

dennisklein:
To be backported to `v18.4_patches` and `v18.6_patches`.


ChristianTackeGSI (approved):
Looks good to me, assuming the TARGET exists on fairmmq-with-dds.

PR #1132, opened 2022-01-18
Use WriteTObject instead of cd + Write

ChristianTackeGSI:
Instead of
```c++
tfile->cd();
tobj->Write();
```
one can also do
```c++
tfile->WriteTObject(tobj);
```

This has the advantage that it does not modify the global gDirectory/gFile (which the calling code might want to keep).

See-Also: https://root.cern/root/htmldoc/guides/users-guide/InputOutput.html#saving-objects-to-disk

---

Checklist:

### DIFF
--- a/base/steer/FairRootManager.cxx
+++ b/base/steer/FairRootManager.cxx
@@ -30,6 +30,7 @@
 #include <TBranch.h>        // for TBranch
 #include <TClonesArray.h>   // for TClonesArray
 #include <TCollection.h>    // for TCollection, TIter
+#include <TDirectory.h>     // for TDirectory::TContext
 #include <TFile.h>          // for TFile
 #include <TFolder.h>        // for TFolder
 #include <TGeoManager.h>    // for TGeoManager, gGeoManager
@@ -412,11 +413,10 @@ void FairRootManager::CreateGeometryFile(const char* geofile)
      *  framework. The geomanager used by the framework is still
      *  stored in the parameter file or database
      */
-    TFile* oldfile = gFile;
+    TDirectory::TContext restorecwd{};
     std::unique_ptr<TFile> file{TFile::Open(geofile, "RECREATE")};
     file->WriteTObject(gGeoManager);
     file->Close();
-    gFile = oldfile;
 }
 
 void FairRootManager::WriteFolder()
--- a/base/steer/FairRootManager.cxx
+++ b/base/steer/FairRootManager.cxx
@@ -48,6 +48,7 @@
 #include <iostream>   // for operator<<, basic_ostream, etc
 #include <list>       // for _List_iterator, list, etc
 #include <map>        // for map, _Rb_tree_iterator, etc
+#include <memory>     // for unique_ptr
 #include <set>        // for set, set<>::iterator
 #include <stdlib.h>   // for exit
 #include <utility>    // for pair
@@ -412,11 +413,9 @@ void FairRootManager::CreateGeometryFile(const char* geofile)
      *  stored in the parameter file or database
      */
     TFile* oldfile = gFile;
-    TFile* file = TFile::Open(geofile, "RECREATE");
-    file->cd();
-    gGeoManager->Write();
+    std::unique_ptr<TFile> file{TFile::Open(geofile, "RECREATE")};
+    file->WriteTObject(gGeoManager);
     file->Close();
-    file->Delete();
     gFile = oldfile;
 }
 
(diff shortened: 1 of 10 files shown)


base/steer/FairRootManager.cxx
```diff
@@ -413,8 +413,7 @@ void FairRootManager::CreateGeometryFile(const char* geofile)
      */
     TFile* oldfile = gFile;
     TFile* file = TFile::Open(geofile, "RECREATE");
```
ChristianTackeGSI, line 415: 1. I wonder, whether the `oldfile` thing can go away?
2. I guess, this should be `unique_ptr`?

Any comments appreciated!

```suggestion
    std::unique_ptr<TFile> file{TFile::Open(geofile, "RECREATE")};
```

dennisklein: > I wonder, whether the oldfile thing can go away?

Yes. Also the [other one in `FairRunInfo::WriteHistosToFile`](https://github.com/FairRootGroup/FairRoot/pull/1132/files#diff-a6317d587accc4f34577889f5ea102681ae20c0a0a3bcf352c1ccdb1e3e39249R156).

dennisklein: > I guess, this should be `unique_ptr`?

Tbh, it should just be a local stack object, e.g.:

```cpp
void FairRootManager::CreateGeometryFile(const char* geofile)
{
    TFile file(geofile, "RECREATE");
    if(file.IsZombie()) {
        LOG(error) << "(Re-)Creating geometry file '" << geofile << "' has failed. Skipping.";
    } else {
        auto const nbytes = file.WriteTObject(gGeoManager);
        LOG(debug) << "Wrote " << nbytes << " bytes to the geometry file '" << geofile << "'.";
    }
}
```

Looks like there is no way to retrieve the recreation error, seems to be just logged by ROOT with log level `kError`. Whether this is treated as a fatal error appears to depend on if the log level is greater than [`gErrorAbortLevel`](https://root.cern.ch/doc/master/TError_8h.html#a856492c7910a90c339d96338211b76fb) - by default it is not fatal.

dennisklein: > > I wonder, whether the oldfile thing can go away?
> 
> Yes. Also the [other one in `FairRunInfo::WriteHistosToFile`](https://github.com/FairRootGroup/FairRoot/pull/1132/files#diff-a6317d587accc4f34577889f5ea102681ae20c0a0a3bcf352c1ccdb1e3e39249R156).

Actually, I am not sure, it seems even a "local ROOT file" (as the docs call it) sets all kinds of global things...

dennisklein: Wonder, if [`RFile`](https://root.cern.ch/doc/master/classROOT_1_1Experimental_1_1RFile.html) fixes this..

dennisklein: > > > I wonder, whether the oldfile thing can go away?
> > 
> > 
> > Yes. Also the [other one in `FairRunInfo::WriteHistosToFile`](https://github.com/FairRootGroup/FairRoot/pull/1132/files#diff-a6317d587accc4f34577889f5ea102681ae20c0a0a3bcf352c1ccdb1e3e39249R156).
> 
> Actually, I am not sure, it seems even a "local ROOT file" (as the docs call it) sets all kinds of global things...

```
root [0] gFile
(TFile *) nullptr
root [1] TFile file("/tmp/asdf.root", "RECREATE");
root [2] gFile
(TFile *) 0x10d45f1b0
```

MohammadAlTurany: This old file trick is used to keep the gFile pointer unchanged, the use case is simple: if a user create a new file to just save some histograms which are created in Macro and should not go in the OutputFile then ROOT set the gFile to this new file and we do not went to change it for the user. I think internally in FairRoot gFile is not used at all (I hope!) but ROOT and User code use it so with this trick we at least do not touch it!

ChristianTackeGSI: Okay, having looked again a bit:

* I can't find any (non experimental) API in ROOT to open a TFile without changing gFile (IMHO this is a missing feature!). So we still need to backup/restore gFile in relevant code.
* TFile::Open has features (for example it can open TWebFiles) that the TFile() constructor does not have, at least the code and docs suggest that to me. So I would only want to switch away from TFile::Open to TFile() if we actually know that the passed name is always a local file.
* I think, we have some places that use gFile… I'll take a look sometime.

I will close this thread and force push an update with `unique_ptr`. Using local stack variables can be done, if we know that these are really local files all the time.

ChristianTackeGSI: ROOT provides an exception safe method for saving and restoring `gFile`/`gDirectory`:

https://root.cern/doc/master/classTDirectory_1_1TContext.html

I have used TContext all around in a dedicated commit in this PR.

dennisklein (approved):

PR #1138, opened 2022-01-31
Add tgeo test

fuhlig1:
Add a test which checks if the correct unit system is used in the TGeoManager.
The test should fail for the moment for some versions of FairSoft. A bug was introduced in the contained ROOT versions which changed the default unit system used in FairRoot.

Checklist:

* [x ] Rebased against `dev` branch
* [ x] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/base/steer/FairRun.cxx
+++ b/base/steer/FairRun.cxx
@@ -21,7 +21,12 @@
 #include "FairRuntimeDb.h"      // for FairRuntimeDb
 #include "FairTask.h"           // for FairTask
 
-#include <TFile.h>     // for TFile
+#include <RVersion.h>      // for ROOT_VERSION_CODE and ROOT_VERSION
+#include <TFile.h>         // for TFile
+#include <TGeoManager.h>   // for gGeoManager, TGeoManager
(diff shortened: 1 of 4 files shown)


fuhlig1:
All test for FairSoft apr21 failed which is the expected behaviour. 
I will add the fix to the PR after I have tested it locally with CbmRoot.

Don' t merge yet.

fuhlig1:
This PR needs to be ported at least to the v18.6_patches branch to create a new patch release of v18.6. This new version is urgently needed by CBM.

base/steer/FairRun.cxx
```diff
@@ -21,7 +21,11 @@
 #include "FairRuntimeDb.h"      // for FairRuntimeDb
 #include "FairTask.h"           // for FairTask
 
-#include <TFile.h>     // for TFile
+#include <TFile.h>         // for TFile
+#include <TGeoManager.h>   // for gGeoManager, TGeoManager
+#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 18, 2) && ROOT_VERSION_CODE <= ROOT_VERSION(6, 22, 6)
```
dennisklein, line 26: Can you please add `#include <RVersion.h>`, thx!

fuhlig1: Done.

dennisklein (approved):

Issue #1111, opened 2021-09-29
Setting name of TTree that is different from cbmsim


ThomasRuf:
Our experiment SND@LHC is starting recording real data. I would like to give the name of the TTree a name different from cbmsim, more related to real data. The only way to change the default name at the moment is to create a directory config, adding a file with the content, "treename=name_you_chose". BUT this is not a good  solution, since I want to be able to read real and MC data in parallel. Don't want to have different projects. Ideally, there should be a set method of the FairRootManager to set the default name. My current solution, create this file on the fly before calling FairFileSource and delete it after. This is not a perfect solution, since it will fail if two jobs, one for MC and one for data are starting at the same time. 
Any suggestions are welcome. Thomas

dennisklein:
Thank you very much for the suggestion. We are looking into it.

karabowi:
Dear Thomas,

we have discussed the issue on our weekly meeting
and we propose you to create your own SndFileSource and SndRootFileSink.
There, the created/read tree is set its name.

In the corresponding FairFileSource and FairRootFileSink, the names
are taken from FairRootManager via the GetTreeName() function.
In your implementations, you will be able to set the tree name 
dynamically to whatever you like.

Should you have any problem, contact us!

Best regards,
Radek

hnatics:
Is it possible to redesign current implementation at some point? 

if I must create a folder /config in main with a one-liner file in it, this is quite a bad design...

ChristianTackeGSI:
Can we close this, now that #1140 has landed?

@ThomasRuf can you take a look, whether #1140 looks useful to you?

ThomasRuf:
What I coincidence! I just started to modify FairRootManager to have this functionality. I first thought also about a Set method, but was then thinking about gROOT->GetListOfGlobals()->FindObject("TreeName") instead of reading from a file.  But if a Set method is now available, will immediately use it. Thank you very much.

ThomasRuf:
Was a bit too optimistic with testing. Just changing to the FairRoot dev version, and building in our environment with alibuild immediately fails, see below. We are currently using  version v18.4.2. What would be the fastest / efficient way to implement #1140 ? 

CMake Error at CMakeLists.txt:95 (find_package):
  By not providing "FindFairCMakeModules.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "FairCMakeModules", but CMake did not find one.

[...]

PR #1140, opened 2022-02-03
Treename fix

karabowi:
Add SetTreeName and SetFolderName setters

Reworked the setting of the tree and folder names.
Added the possibility to set these names in the macro
by calling either FairRootManager::Instance()->SetTreeName()
or FairRootManager::Instance()->SetFolderName().
These setters have to be called before calling run->Init().
    
Addresses issue #1111.

---

Checklist:

[ x ] Rebased against `dev` branch
[ x ] My name is in the resp. CONTRIBUTORS/AUTHORS file
[ x ] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/base/steer/FairRootManager.cxx
+++ b/base/steer/FairRootManager.cxx
@@ -67,6 +67,8 @@ TMCMutex deleteMutex = TMCMUTEX_INITIALIZER;
 }   // namespace
 
 Int_t FairRootManager::fgCounter = 0;
+std::string FairRootManager::fFolderName = "";
+std::string FairRootManager::fTreeName = "";
 
 FairRootManager* FairRootManager::Instance()
 {
@@ -103,7 +105,6 @@ FairRootManager::FairRootManager()
     , fEntryNr(0)
     , fListFolder(0)
     , fSource(0)
-    , fSourceChain(new TChain(GetTreeName(), Form("/%s", GetFolderName())))
     , fSignalChainList()
     , fEventHeader(new FairEventHeader())
     , fSink(nullptr)
@@ -161,6 +162,8 @@ FairRootManager::~FairRootManager()
 
 Bool_t FairRootManager::InitSource()
 {
+    // creating fSourceChain now, to give the user chance to change the tree and folder name
+    fSourceChain = new TChain(GetTreeName(), Form("/%s", GetFolderName()));
 
     LOG(debug) << "Call the initialiazer for the FairSource in FairRootManager ";
     if (fSource) {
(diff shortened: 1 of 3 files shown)


base/steer/FairRootManager.cxx
```diff
@@ -103,7 +105,7 @@ FairRootManager::FairRootManager()
     , fEntryNr(0)
     , fListFolder(0)
     , fSource(0)
-    , fSourceChain(new TChain(GetTreeName(), Form("/%s", GetFolderName())))
+    , fSourceChain(0)   // 2022.01.11: moved the creation of fSourceChain to InitSource
```
ChristianTackeGSI, line 108: Is the comment really needed?

What about moving this into the header like this?
```c++
TChain* fSourceChain = nullptr;
```

base/steer/FairRootManager.cxx
```diff
@@ -974,70 +978,81 @@ void FairRootManager::DeleteOldWriteoutBufferData()
...
+        strcpy(treeName, fTreeName.c_str());
+        return treeName;
     }
-    char* workdir = getenv("VMCWORKDIR");
-    if (nullptr == workdir) {
-        return default_name;
+
+    std::string nameFromFile = GetNameFromFile("treename=");
+    if (nameFromFile.length() > 0) {
+        char* treeName = new char[nameFromFile.length()];
```
ChristianTackeGSI, line 1025: You have to allocate `fTreeName.length() + 1`. The final NUL byte needs to have space as well.

base/steer/FairRootManager.cxx
```diff
@@ -974,70 +978,81 @@ void FairRootManager::DeleteOldWriteoutBufferData()
...
+        return treeName;
     }
 
-    // Key "foldername" was not found in file: return default
-    fclose(file);
+    char* default_name = new char[7];
+    strcpy(default_name, "cbmout");
+    if (!FairRun::Instance()->IsAna()) {
+        delete[] default_name;
+        strcpy(default_name, "cbmroot");
```
ChristianTackeGSI, line 1054: `delete` and then `strcpy`? Doesn't sound like this works?

base/steer/FairRootManager.cxx
```diff
@@ -974,70 +978,81 @@ void FairRootManager::DeleteOldWriteoutBufferData()
...
-                fclose(file);
-                return foldername;
-            }
-            delete[] foldername;
-        }
+    std::string nameFromFile = GetNameFromFile("foldername=");
+    if (nameFromFile.length() > 0) {
+        char* treeName = new char[nameFromFile.length()];
+        strcpy(treeName, nameFromFile.c_str());
+        return treeName;
```
ChristianTackeGSI, line 1047: All of that said, I am still not happy with `FairRootManager::Get{Tree,Folder}Name` returning allocated space that probably nobody is going to free (see our own usage a few lines up).

One option would be to use [`Form`](https://root.cern/doc/master/TString_8cxx.html#acb421e430108807d6d57ee3398f84716)'s circular buffer and let the methods return a `const char *`?
```suggestion
        return Form("%s", nameFromFile.c_str());
```
I am quite unsure, really.

base/steer/FairRootManager.h
```diff
@@ -343,6 +345,12 @@ class FairRootManager : public TObject
      */
     std::map<TString, TObject*> fMap;   //!
 
+    /**folder name variable*/
+    static std::string fFolderName;
+    /**tree name variable*/
+    static std::string fTreeName;
```
ChristianTackeGSI, line 351: Please append a `   //!`.
I learned, that this will remove the fields from root's serializer.

OR if they should be serialized, consider bumping the version.

ChristianTackeGSI (changes requested):
Please at least fix the memory issues (allocate more space, reallocate after free).

base/steer/FairRootManager.cxx
ChristianTackeGSI: Looks soo much better!

As the functions now return memory that doesn't need to be freed nor changed: What about returning `const char*`?

ChristianTackeGSI: Thanks!

ChristianTackeGSI (approved):
I added a bit of changelog and used the string literals directly, now that we return `const char*`.

ChristianTackeGSI (approved):
Looks very good!

I just learned
> These setters have to be called before calling run->Init().

from the commit message. Maybe we should add some doxygen style comments to both setters and note that there?

PR #1142, opened 2022-02-04
Cleanup FairRadGridManager and FairMCApplication

ChristianTackeGSI:
* FairRadGridManager: Drop unused private member variables, better handling of singleton variable in destructor
* FairMCApplication: Drop unused private Operator
* FairMCApplication: Use unique_ptr for fRadGridMan

---

Checklist:


fuhlig1 (dismissed):

ChristianTackeGSI:
rebased, ready for review.

karabowi (approved):
I have tried your change with Geant4 MT, I wanted to check if the deletion of FairMCApplication's operator= causes any problem. It does not.
Tested with PandaRoot as well.

base/sim/FairMCApplication.h
```diff
@@ -302,13 +303,15 @@ class FairMCApplication : public TVirtualMCApplication
     /** Current state */
     FairMCApplicationState fState;   //!
 
-    ClassDef(FairMCApplication, 4);
+    ClassDefOverride(FairMCApplication, 4);
```
fuhlig1, line 306: Since you changed a data member, you should increase the version counter.

fuhlig1 (changes requested):
See my change request.

base/sim/FairMCApplication.h
ChristianTackeGSI: I think, I only changed `fRadGridMan`.
This one was and still is marked with `//!`.

From looking at [Chappter 13.3.2 Transient Data Members](https://root.cern/root/htmldoc/guides/users-guide/InputOutput.html#transient-data-members) it seems that this marks a member variable as not taking part in marshalling.

I had assumed, that the version marking is resposible for on-disk versions?

fuhlig1: You are right. Normally I increase the version when I change something but in this case it shouldn't be necessary.

ChristianTackeGSI:
So, the change request has been resolved.

What's missing so that this can be merged?

fuhlig1 (approved):

PR #1150, opened 2022-02-28
Fix segfault when closing event display.

DanielWielanek:
This should fix the issue 

---

Checklist:

### DIFF
--- a/eventdisplay/FairEventManager.cxx
+++ b/eventdisplay/FairEventManager.cxx
@@ -157,20 +157,10 @@ void FairEventManager::Init(Int_t visopt, Int_t vislvl, Int_t maxvisnds)
     fAxesPhi = new TEveProjectionAxes(fRPhiProjManager);
     fAxesRho = new TEveProjectionAxes(fRhoZProjManager);
 
-    TEveWindowSlot *RPhiSlot = TEveWindow::CreateWindowInTab(gEve->GetBrowser()->GetTabRight());
-    TEveWindowPack *RPhiPack = RPhiSlot->MakePack();
-    RPhiPack->SetElementName("RPhi View");
-    RPhiPack->SetShowTitleBar(kFALSE);
-    RPhiPack->NewSlot()->MakeCurrent();
     fRPhiView = gEve->SpawnNewViewer("RPhi View", "");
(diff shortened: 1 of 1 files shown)


ChristianTackeGSI:
Thanks for looking into this!

I have no experience with this part of the code at all, but I just wonder, what the removed code was actually supposed to perform? Isn't that needed any longer or was it not working anyhow?

DanielWielanek:
Actually this code is responsible for "the tabs" that hold detector's views (3D, two projections, and 3D+ projections).
In principle I copied them from some ROOT tutorials. But now I found out, that (from uknown reason) I can create only one "tab" without segfaut. So now only one "tab" is created - for this 3 views. It seems that rest of the views doesn't require "tab" creation - it's done automatically. Maybe that's why code crash.
Unfortunatelly the ROOT documentation is very, very poor - there is even no description what some classes really do. Actually event debugger did not help mee because gdb stucks in some infinite loops.

kresan:
Did not fix the crash in my case, still segfaulting when closing the window. I was using Rutherford example to test your modifications.
macOS 12.2.1, FairSoft apr21p2.

DanielWielanek:
Do you mean examples/simulation/rutherford? I looked and seems to work for me (FairSoft apr21_patches from October + Linux Mint).
Did you just open and close the ED? or do some additional steps?

kresan:
I was first clicking on the projections and then closing the window. Moreover, crash happens not always. First 2 tries were successful.
Maybe we need more crash tests, on other systems.

ChristianTackeGSI:
@kresan Does the patch in this PR improve the crash situation at least for you? Or is it as random as without?

kresan:
The changes in this PR do not have noticeable effect on the seg fault (as random as without) and do not break the functionality we used to have.

ChristianTackeGSI:
I *personally* would say to merge it, because it removes seemingly unneeded code.

kresan:
It works as expected on Ubuntu 18.04. The window can be closed without crash of ROOT. Green light from my side.

@TobiasStockmanns, do you want to test with PandaRoot or can we merge this PR?

karabowi (approved):

ChristianTackeGSI:
@DanielWielanek thanks for you contribution! It took a bit longer for us to review and check some corner cases, but now it's in.

PR #1153, opened 2022-03-03
Deprecate FairTask::*InputPersistance

ChristianTackeGSI:
* There are only setters, no getters
* Only derived classes could read fInputPersistance
* We're not aware of anyone using this

---

Checklist:


base/steer/FairTask.h
```diff
@@ -21,8 +21,6 @@
 #ifndef FAIRTASK_H
 #define FAIRTASK_H
 
-#include "FairRootManager.h"   // for FairRootManager
```
karabowi, line 24: The removal of FairRootManager include from the FairTask header file
forces few additions of same include in the PandaRoot project.
Is it beneficial to remove it here?

karabowi: I think such an addition in the changelog is not needed.
Since the FairRootManager is used in these PandaRoot tasks, the corresponding include should be there anyways.

karabowi (approved):

PR #1156, opened 2022-03-16
ParRootFileIo: Refactor Multi File Merging

ChristianTackeGSI:
Refactor the part of `FairParRootFileIo::open(TList *)` that merges multiple ROOT files from a list of names into one file into a new static `mergefile` method.

---

Checklist:

### DIFF
--- a/parbase/FairParRootFileIo.cxx
+++ b/parbase/FairParRootFileIo.cxx
@@ -28,15 +28,17 @@
 #include "FairRtdbRun.h"     // for FairRtdbRun
 #include "FairRuntimeDb.h"   // for FairRuntimeDb
 
-#include <TCollection.h>   // for TIter
-#include <TDatime.h>       // for TDatime
-#include <TKey.h>          // for TKey
-#include <TList.h>         // for TListIter, TList
-#include <TObjString.h>    // for TObjString
-#include <TObject.h>       // for TObject
-#include <TString.h>       // for TString, Form
-#include <iostream>        // for operator<<, basic_ostream, etc
-#include <memory>          // for std::unique_ptr
+#include <TCollection.h>         // for TIter
+#include <TDatime.h>             // for TDatime
+#include <TDirectory.h>          // for TDirectory::TContext
+#include <TKey.h>                // for TKey
+#include <TList.h>               // for TListIter, TList
+#include <TObjString.h>          // for TObjString
+#include <TObject.h>             // for TObject
+#include <TString.h>             // for TString, Form
+#include <fairlogger/Logger.h>   // for LOG
+#include <iostream>              // for operator<<, basic_ostream, etc
+#include <memory>                // for std::unique_ptr
 
 using std::cerr;
 using std::cout;
@@ -146,6 +148,12 @@ Bool_t FairParRootFileIo::open(const Text_t* fname, Option_t* option, const Text
     return kFALSE;
 }
 
+/**
+ * \brief Combine a list of parfiles into one, and open it
+ * \warning Creation of the allParams file has a race condition.
+ *   Create the filename yourself, then use \ref MergeFiles,
+ *   and finally \ref open.
+ */
 Bool_t FairParRootFileIo::open(const TList* fnamelist, Option_t* option, const Text_t* ftitle, Int_t compress)
 {
     TDatime currentDate;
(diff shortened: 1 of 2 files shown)


klenze:
Hi @ChristianTackeGSI, thanks for the ping. 

From what I understand, this patch basically allows the user to work around the race condition (or inability to write into the directory given in the last entry of the TList) by
* manually calling MergeFiles() with a uniquely named TFile (or TMemFile instance)
* using the merged file on open(const Text_t* fname, ...)

(Which is what you suggested in https://github.com/FairRootGroup/FairRoot/pull/1090#issuecomment-900191293 )

Also, it provides users the ability to cleanly merge their parameter files, which is way better than "just open them with open(TList...), then search for the most recent allParams_*.root". 

In Python, I would probably want a named argument outfile to open(TList...)  (with the default None corresponding to using the allParams filename), but using four positional parameters in a C++ function constructor seems excessive. :-)

I think the TObjectString* string (I tend to avoid naming my instances after classes in the std namespace) might be uninitialized if the TList is empty. I would either initialize it with {}/nullptr so we only get a segfault instead of UB if that happens, or initialize it to point to a static const TObjectString(""). Not an issue introduced with this PR, though. 

The body of the while loop assigning string in open(TList...) seems unneeded, MergeFiles will already emit a warning for invalid files, no need to annoy people who take the shortcut "set the last item in the TList to the directory where you want your merged file to appear" twice. 

Feel free to close my PR when this is merged. 
Good night,
  Philipp

ChristianTackeGSI:
Hi @klenze,

Many thanks for your review!

> From what I understand, this patch basically allows the user to work around the race condition (or inability to write into the directory given in the last entry of the TList) by
> 
> * manually calling MergeFiles() with a uniquely named TFile (or TMemFile instance)
> 
> * using the merged file on open(const Text_t* fname, ...)

Remember to close the file before calling `open(fname)`.

Maybe we should once have some `open(TFile*)`?

> (Which is what you suggested in [#1090 (comment)](https://github.com/FairRootGroup/FairRoot/pull/1090#issuecomment-900191293) )

It is one part of that suggestion. I thought, I'll start and try to implement the basics.

> In Python, I would probably want a named argument outfile to open(TList...) (with the default None corresponding to using the allParams filename), but using four positional parameters in a C++ function constructor seems excessive. :-)

I wanted to avoid those "cook coffee" APIs. I rather like to provide people with easy to handle tools, so that they can plug them together and cook coffee or do something else. As you said, people could for example once use MergeFiles to merge things into a TMemFile.

> I think the TObjectString* string (I tend to avoid naming my instances after classes in the std namespace) might be uninitialized if the TList is empty. I would either initialize it with {}/nullptr so we only get a segfault instead of UB if that happens, or initialize it to point to a static const TObjectString(""). Not an issue introduced with this PR, though.

I doubt that. If the TList is empty, the first (and final) call to `myIter.Next()` should return a nullptr.

After that, `open` is called with `""`. Maybe not what sounds like a good idea. But that was broken before. 

```c++
    if (!newParFile) {
        return kFALSE;
    }

    MergeFiles(newParFile.get(), fnamelist);
```

What about this?

> The body of the while loop assigning string in open(TList...) seems unneeded, MergeFiles will already emit a warning for invalid files, no need to annoy people who take the shortcut "set the last item in the TList to the directory where you want your merged file to appear" twice.

Well, the code chooses the *first writeable* directory from the input list. That was the old behaviour and I didn't want to break that.

> Feel free to close my PR when this is merged. Good night, Philipp

Thanks! Let's see, when / how it gets merged.

klenze:
Sorry, I completely misread the block of the while loop last night. Somehow thought it just went to line 169. So yes, the first location where the file can be created is used.

I still think lines 165-169 are unneeded, inFile will be used by  MergeFiles anyhow. 

Returning false if no writable location could be found depends on people checking the returned value (which they mostly don't, in my experience). People copying others macros and using parameter files from their colleagues are then in for a hard to debug surprise if they want to access their parameters down the road. Raising an exception or writing an ERROR level log message would probably be kinder.

ChristianTackeGSI:
So, what about something like this?
```c++
    if (!newParFile) {
        throw std::runtime_error("FairParRootFileIo::open: Could not generate merged file");
    }
```

ChristianTackeGSI:
@klenze What do you think about  `LOG(error)` (as now in the current patch proposal in this PR)?

karabowi (approved):
Tested this functionality with proof, for which it was originally created.

PR #1173, opened 2022-04-22
Do not Use "using namespace"

ChristianTackeGSI:
"using namespace foo" isn't really good.  It will import
all names from that nemespace, including any future ones
that might conflict with local names.

So avoid this construct.

---

Checklist:

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,4 +2,9 @@
 HeaderFilterRegex: '.*'
 # disable clang-analyzer-core.UndefinedBinaryOperatorResult
 #   ROOT throws lots of them in their headers
-Checks: '-clang-analyzer-core.UndefinedBinaryOperatorResult,modernize-make-unique'
+# enable google-build-using-namespace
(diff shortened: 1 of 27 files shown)


dennisklein (approved):

fuhlig1 (commented):
Sometimes you exchange 

`using namespace std;`

by e.g.

```
using std::cout;
using std::endl;
```

and sometimes you add the std namespace to a type or function in the code

`const string& -> const std::string&`

instead of adding

`using std::string;`


Is there a reason why you do this differently?

ChristianTackeGSI:
@fuhlig1 thanks for looking at this!

Mostly amount of "lines that need changing".

If there is only one or two uses of `string`, then I replaced it with `std::string`. Other files have a big list of things using `cout`, so I thought it would be better to say: okay, this seems to be an important ingredient of this file, let's make it a global name ("shortcut").
The main point being that it is now explicit.

Or maybe in another terms: Reduce the amount of diff that needs to be reviewed.

So after all, this is purely subjective and fully open for changes.

PR #1174, opened 2022-04-22
Rutherford: Default to Geant4

ChristianTackeGSI:
eventDisplay.C already uses Geant4 by default Switch run_rutherford.C and run_rad.C to use Geant4 by default as well.

---

Checklist:

### DIFF
--- a/examples/simulation/rutherford/src/FairRutherford.h
+++ b/examples/simulation/rutherford/src/FairRutherford.h
@@ -33,27 +33,27 @@ class FairRutherford : public FairDetector
     FairRutherford();
 
     /**       destructor     */
-    virtual ~FairRutherford();
+    ~FairRutherford() override;
 
     /**      Initialization of the detector is done here    */
-    virtual void Initialize();
+    void Initialize() override;
 
     /**       this method is called for each step during simulation
      *       (see FairMCApplication::Stepping())
      */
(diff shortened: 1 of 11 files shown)


examples/simulation/rutherford/src/FairRutherford.h
```diff
@@ -67,30 +67,30 @@ class FairRutherford : public FairDetector
...
     active volume.
     */
-    Int_t fTrackID;        //!  track index
-    Int_t fVolumeID;       //!  volume id
-    TLorentzVector fPos;   //!  position at entrance
-    TLorentzVector fMom;   //!  momentum at entrance
-    Double32_t fTime;      //!  time
-    Double32_t fLength;    //!  length
-    Double32_t fELoss;     //!  energy loss
+    Int_t fTrackID = -1;        //!  track index
```
fuhlig1, line 87: What about using consistently {} for initialisation for all members?

ChristianTackeGSI: Good point!

I was considering to use `=` for base types and `{}` for classes. But maybe we should really just use `{}` all around?

ChristianTackeGSI: Okay, changed over. Should be resolved for this one.

karabowi (approved):

PR #1176, opened 2022-04-26
Convert Some Member Variables to Local Ones

ChristianTackeGSI:
On some classes, a few member variables are only really used in one (or two) methods to handle local data. So convert those member variables to local ones in that method.

This improves readability and locality of the code.
It also shrinks the instance's size.

---

Checklist:


base/sim/FairMCApplication.cxx
```diff
@@ -579,13 +568,13 @@ void FairMCApplication::Stepping()
     // In any case call the ProcessHits function for this specific detector.
     Int_t copyNo;
     Int_t id = fMC->CurrentVolID(copyNo);
-    Bool_t InMap = kFALSE;
-    fDisVol = 0;
-    fDisDet = 0;
     Int_t fCopyNo = 0;
-    fVolIter = fVolMap.find(id);
```
fuhlig1, line 586: What is the time needed to create the iterator? The stepping is the function which is called most during the transport simulation so it could well be that the iterator was a class member. 
Please  do a large scale transport simulation to  check the effect of the change on the runtime.
@MohammadAlTurany, what do you mean?

ChristianTackeGSI: Well, take a look at the generated assembly:

https://godbolt.org/z/87fdfnenW

1. Instanatiating a `std` iterator seems to be as simple as setting one qword to zero.
2. The assembly with the local variable looks longer and more convoluted to me. I haven't taken the time to analyze it fully, but it looks like it's very much the same, except for having to store the results in the member variable.

karabowi (approved):

base/sim/FairMCApplication.cxx
karabowi: The transport simulation runtime is comparable, no substantial time differences.

PR #1199, opened 2022-06-09
fix: new boost does not play nice with old-boost system yaml-cpp

klenze:
While trying to compile FairRoot on lxir136.gsi.de using FairSoft release from april 2022 (using gcc9):
```
[ 66%] Generating G__MCConfigurator.cxx, ../../lib/G__MCConfigurator_rdict.pcm, ../../lib/libMCConfigurator.rootmap
In file included from input_line_9:120:
In file included from /usr/include/yaml-cpp/yaml.h:17:
In file included from /usr/include/yaml-cpp/node/impl.h:11:
In file included from /usr/include/yaml-cpp/node/iterator.h:13:
/usr/include/yaml-cpp/node/detail/iterator.h:48:54: error: no member named 'next' in namespace 'boost'
  void increment() { this->base_reference() = boost::next(this->base()); }
                                              ~~~~~~~^
Error: /u/land/klenze/2022-06/inst/FairSoft/bin/rootcint: compilation failure (/u/land/klenze/2022-06/src/FairRoot/build/fairtools/MCConfigurator/G__MCConfigurator39d6ad62e4_dictUmbrella.h)
fairtools/MCConfigurator/CMakeFiles/MCConfigurator.dir/build.make:74: recipe for target 'fairtools/MCConfigurator/G__MCConfigurator.cxx' failed
make[2]: *** [fairtools/MCConfigurator/G__MCConfigurator.cxx] Error 1
CMakeFiles/Makefile2:1250: recipe for target 'fairtools/MCConfigurator/CMakeFiles/MCConfigurator.dir/all' failed
make[1]: *** [fairtools/MCConfigurator/CMakeFiles/MCConfigurator.dir/all] Error 2
Makefile:145: recipe for target 'all' failed
make: *** [all] Error 2

```
(Side note: Could you please make the LinkDef depend on all the object files of the current library, thus making sure that the first compiler parsing the code is a grown-up compiler which is actually able to tell the top level file? Handling rootcint errors is bad enough when they are actually rootcint specific. For code rejected even by decent compilers, a decent error message should be generated. Do you expect every user to know that they need VERBOSE=1 make to get any idea what went wrong with rootcint?)

If we are using a system yaml-cpp which was compiled with an older version of boost, it might just include boost utility instead of boost/next_prior.hpp, which will not play well with the latest and greatest version of boost included in FairSoft. Manually included correct header for boost::next so yaml-cpp/yaml.h does not crash & burn.

(Side remark: YAML for configuration? Even when ROOT does not yet provide a pointless TYAML interface class? What happened to "ROOT macros are the ideal format to store configuration settings"? Just kidding.)

---

Checklist:

* [ ] Rebased against `dev` branch -> Nah, it's a fix for a released version. 
* [ ] My name is in the resp. CONTRIBUTORS/AUTHORS file -> Not for a 1 line change.
* [ ] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules) -> Possibly, accidentially?

### DIFF
--- a/fairtools/MCConfigurator/FairYamlVMCConfig.h
+++ b/fairtools/MCConfigurator/FairYamlVMCConfig.h
@@ -15,6 +15,10 @@
 
 #include "FairGenericVMCConfig.h"
 
+// in case system yaml-cpp was compiled with an older boost version where boost::next was defined
+// in utility.hpp:
+#include <boost/next_prior.hpp>
+
 #include <string>
 #include <yaml-cpp/yaml.h>
 


ChristianTackeGSI:
Thanks for the patch and the issue!

Just so that we can try to reproduce the issue:
* What version of Debian do you run on `lxir136`?
* yaml-cpp from the system pacakges?
* Do you know any other special settings?

I think, I would try to setup a CI container with some alike settings afterwards.

klenze:
lxir136 is running stretch (don't blame me, blame IT). 
The header files seem to be from:
libyaml-cpp-dev/oldoldstable,now 0.5.2-4 amd64 [installed,automatic]

Additionally, I am using a custom-compiled gcc-9, but I think it is somewhat unlikely that this changes anything. 

Can you login at lxir136.gsi.de from within GSI? That would probably be quicker than setting up a container. 

https://www.google.com/search?q=boost+next+yaml-c%2B%2B finds several other projects where people encountered this issue. Fortunately, I think we have some advantage here as we are shipping a recent boost as a dependency of FairRoot, so the include should never fail: Either the user has a newer yaml-c++ which uses std::next, in which case we waste a few microseconds by including unrelated code, or it relies on boost::next, in which case we make sure it is defined. 

A cleaner way to solve this would be for fairsoft to require (and provide) yaml-cpp 0.6.0, which dropped boost for c++11.

dennisklein:
Hm, I am against merging this PR. We have already so much technical debt, adding more hacks does not improve the situation.

> A cleaner way to solve this would be for fairsoft to require (and provide) yaml-cpp 0.6.0, which dropped boost for c++11.

I would agree, if Stretch would still be supported by FairSoft `apr22` which it is not, because `apr22` requires C++17. However, if you really want to go down the road and backport FairSoft `apr22` with your custom compiler to Stretch, with the following patch you can include a newer yaml-cpp:

```diff
--- a/cmake/legacy.cmake
+++ b/cmake/legacy.cmake
@@ -114,6 +114,18 @@ ExternalProject_Add(faircmakemodules
   ${LOG_TO_FILE}
 )

+list(APPEND packages yaml-cpp)
+set(yaml-cpp_version "0.7.0")
+ExternalProject_Add(yaml-cpp
+  GIT_REPOSITORY https://github.com/jbeder/yaml-cpp GIT_TAG yaml-cpp-${yaml-cpp_version}
+  ${CMAKE_DEFAULT_ARGS} CMAKE_ARGS
+    "-DBUILD_SHARED_LIBS=ON"
+    "-DYAML_CPP_FORMAT_SOURCE=OFF"
+    "-DYAML_CPP_BUILD_TESTS=OFF"
+  DEPENDS ${extract_source_cache_target}
+  ${LOG_TO_FILE}
+)
+
 list(APPEND packages asio)
 set(asio_version "1.19.1")
 ExternalProject_Add(asio
```

Hope, this helps as a short-term solution. (I haven't tried compiling the full stack in your environment, feel free to contact us, if something is still missing)

The even better solution should be to no longer run oldoldstable, but that is out of scope of FairRoot/FairSoft. (I think LTS support ends in two weeks, so an OS upgrade of your machine would anyways be desirable, if you need an argument for your administrator)

klenze:
@dennisklein I totally get your point about technological debt and hacks. 

Looks like IT is already successfully running Buster on lxpool (even though it is not even EOL yet). Within my collaboration, we are running a mixture of Wheezy i686 (DAQ), Jessie and Stretch. I seriously doubt that there will be much enthusiastic support for upgrading to Buster (or, more reasonably, Bullseye) -- it would force people to recompile their code and fix any issues which appear. 
My hope is that at some point we will get a new computer and install a more recent Debian.

ChristianTackeGSI:
I wonder, whether we should raise the required yaml-cpp version to 0.6 for FairRoot/dev (upcoming FairRoot v19)?

dennisklein:
Hm, why?

dennisklein:
As far as I understand the issue is that old yaml-cpp doesn't play nice with new boost. Which is unrelated to FairRoot, right?

ChristianTackeGSI:
Possible reasons (not saying they are good / valid):
* Require versions that are more "stable" (are not broken with newer boost)
* If I got it right, the newer yaml-cpp is based on `std::` and no longer on boost? So maybe to reduce boost requirements in the long run?

As I said, I was wondering, not proposing.

Issue #1203, opened 2022-06-20
Corrupted FairRoot prefix after calling cmake twice


TobiasStockmanns:
**Describe the bug**
If cmake is called twice in PandaRoot either explicity via `cmake -S ./source -B ./build` or implicitly via make the paths to the `generated_dicitionary_root.sh.in` file gets corrupted.
The error message is:
```
CMake Error: File /Users/tstockmanns/PandaSoftware/FairRoot/dev/install/lib/cmake/FairRoot-19.0.0/share/fairbase/cmake/scripts/generate_dictionary_root.sh.in does not exist.
CMake Error at /Users/tstockmanns/PandaSoftware/FairRoot/dev/install/share/fairbase/cmake/modules/ROOTMacros.cmake:107 (Configure_File):
  Configure_File Problem configuring file
  ```
The correct path would be `/Users/tstockmanns/PandaSoftware/FairRoot/dev/install/share/fairbase/cmake/scripts/generate_dictionary_root.sh.in`.

Looking at the output of the cmake call the variable: `FairRoot prefix` changes between the first and second call:
- `--   FairRoot prefix            : /Users/tstockmanns/PandaSoftware/FairRoot/dev/install`
- `--   FairRoot prefix            : /Users/tstockmanns/PandaSoftware/FairRoot/dev/install/lib/cmake/FairRoot-19.0.0`

Most probably this is not a problem of PandaRoot but caused by the cmake scripts coming with FairRoot.

ChristianTackeGSI:
Thanks for the report!

I am putting some notes here, what I suspect. I will try to look deeper in the next days.

I guess, the problem is in FindFairRoot.cmake:
https://github.com/FairRootGroup/FairRoot/blob/5fab55df73fb3a51bdd81160279b08548851fc4b/cmake/modules/FindFairRoot.cmake#L11-L12

* In the first run, `FairRoot_DIR` is not set. All works normally.
* In the first run a config mode `find_package` finally happens:
  https://github.com/FairRootGroup/FairRoot/blob/5fab55df73fb3a51bdd81160279b08548851fc4b/cmake/modules/FindFairRoot.cmake#L68
  This sets `FairRoot_DIR` as a cache variable to the path containing the `FairRootConfig.cmake` config mode package (or I would assume so).
* The cache variable persits…
* During the second run, the above `if(FairRoot_DIR` code is executed and redirects `FAIRROOTPATH` to the wrong path
* Things go wrong

[...]

PR #1204, opened 2022-06-23
fix(cmake): Handle Double CMake in Using Projects

ChristianTackeGSI:
A typical "ExperimentRoot" project uses FairRoot by using its find-Module (by adding the right path) and calling find_package(FairRoot).

This setup worked for the first call to CMake (configure step), but reconfiguring (running CMake again) did not work correctly.

Add a test for this (that failed without this fix).

The find module recursively calls find_package(FairRoot CONFIG) to load the new config mode package.
This sets FairRoot_DIR to the directory containing the config mode package. This variable is a cache variable and persists into the next run of CMake.

The old find module used to think, that FairRoot_DIR could be the prefix of the FairRoot installation.

Rewrite:

* Use FAIRROOTPATH and ENV{FAIRROOTPATH} only to extend CMAKE_PREFIX_PATH to search the config mode package
* Use FairRoot_PREFIX from the config mode package to guide further old style detection
* Use FindPackageHandleStandardArgs

fixes: #1203

---

Checklist:

### DIFF
--- a/cmake/modules/FindFairRoot.cmake
+++ b/cmake/modules/FindFairRoot.cmake
@@ -1,5 +1,5 @@
  ################################################################################
- #    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    #
+ # Copyright (C) 2014-2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  #
  #                                                                              #
  #              This software is distributed under the terms of the             # 
  #              GNU Lesser General Public Licence (LGPL) version 3,             #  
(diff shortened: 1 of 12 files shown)


karabowi (approved):
I have checked with PandaRoot, and finally rebuilding does not constant removal of CMakeCache.

karabowi (approved):

ChristianTackeGSI:
Testing with and without double cmake doubled the warnings from the templates. So fixed all of them as well.

karabowi:
Got a green light from Tobias via private email.

Issue #1184, opened 2022-05-17
Add virtual finalize method for FairGenerator.


mgoncerz:
For my use case, I have interfaced external Monte Carlo generators to our FairRoot-based software. Many of them provide additional information when finalizing. 

While it's very easy to implement it manually and call it from a macro, it would be convenient to have a virtual Finalize method in the base FairGenerator class, that would be always called somewhere at the end of the production run.

karabowi:
After short discussion I think we can add the virtual Finish() to FairGenerator. 
It is quite interesting, that this feature was not requested by anyone for almost 20 years.

PR #1214, opened 2022-07-15
Implement Finish function for generator

karabowi:
Solves the issue #1184.

---

Checklist:


base/sim/FairGenerator.h
```diff
@@ -52,6 +52,9 @@ class FairGenerator : public TNamed
     /**Initialize the generator if needed */
     virtual Bool_t Init() { return kTRUE; }
 
+    /**Finalize the generator if needed */
+    virtual Bool_t Finish() { return kTRUE; }
```
ChristianTackeGSI, line 56: The return value is not used at all in our code.

So let's just return `void`, please.

In exceptional cases, people can still throw exceptions.

base/sim/FairPrimaryGenerator.h
```diff
@@ -52,6 +52,9 @@ class FairPrimaryGenerator : public TNamed
     /** Initialize the generater (if needed!)*/
     virtual Bool_t Init();
 
+    /** Finalize the generater (if needed!)*/
+    virtual Bool_t Finish();
```
ChristianTackeGSI, line 56: This one always returns `kTRUE`. Are people expected to check the return value?
If not, let's return `void`?

Are we expecting this function to be overriden at some point? If not, we could keep it non-virtual to start with, maybe?

base/sim/FairPrimaryGenerator.cxx
```diff
@@ -153,6 +153,19 @@ Bool_t FairPrimaryGenerator::Init()
...
+
+Bool_t FairPrimaryGenerator::Finish()
+{
+    /** Initialize list of generators*/
+    for (Int_t i = 0; i < fGenList->GetEntries(); i++) {
+        FairGenerator *gen = static_cast<FairGenerator *>(fGenList->At(i));
+        if (gen) {
+            gen->Finish();
+        }
+    }
```
ChristianTackeGSI, line 165: Maybe use TRangeDynCast?
I know, it's doing a `dynamic_cast`, which means a bit of performance. But this is in FinishRun, so maybe not as bad?

I think, in the long run, we should switch to `std::vector`.

```suggestion
    for (auto gen : TRangeDynCast<FairGenerator>(fGenList)) {
        if (gen) {
            gen->Finish();
        }   
    }
```

ChristianTackeGSI (approved):

PR #1222, opened 2022-09-14
Change EveDisplay check for tracks to fatal

DanielWielanek:
Describe your proposal.
1. When FairGeoTracksDraw/FairMCTrackDraw does not find data, it returns kFATAL - which closes the event display. I propose to change it into the kERROR - thanks to this rest of the event display tasks should still work.
2. Another change is using TClass to check if MCTrack branch contains FairMCTrack-based class - if not - then deactivate the task. This prevents the situation when branch "MCTrack" is present but contains a different class (which leads to unexpected behavior).

Mention any issue this PR is resolves or is related to.

---

Checklist:

### DIFF
--- a/eventdisplay/tracks/FairEveGeoTracks.cxx
+++ b/eventdisplay/tracks/FairEveGeoTracks.cxx
@@ -48,9 +48,11 @@ FairEveGeoTracks::FairEveGeoTracks()
 InitStatus FairEveGeoTracks::Init()
 {
     FairRootManager *mngr = FairRootManager::Instance();
-    fContainer = (TClonesArray *)mngr->GetObject("GeoTracks");
-    if (fContainer == nullptr)
-        return kFATAL;
+    fContainer = dynamic_cast<TClonesArray *>(mngr->GetObject("GeoTracks"));
+    if (!fContainer) {
+        LOG(warning) << "Geo tracks not found ! FairGeoTracksDraw will be deactivated.";
+        return kERROR;
+    }
     fBranch = mngr->GetInTree()->GetBranch("GeoTracks");
     FairGetEventTime::Instance().Init();
     return FairEveTracks::Init();
@@ -99,8 +101,8 @@ void FairEveGeoTracks::DrawAnimatedTrack(TGeoTrack *tr, double t0)
     FairEveTrack *track = new FairEveTrack(p, p->GetPdgCode(), trList->GetPropagator());
     track->SetElementTitle(Form("p={%4.3f,%4.3f,%4.3f}", p->Px(), p->Py(), p->Pz()));
     track->SetMainColor(color);
-    Double_t x, y, z, t;       // currentPoint
-    Double_t xp, yp, zp, tp;   // previousPoint
+    Double_t x, y, z, t;                       // currentPoint
+    Double_t xp = 0, yp = 0, zp = 0, tp = 0;   // previousPoint
     bool firstPoint = true;
     bool previousPoint = false;
     for (int i = 0; i < tr->GetNpoints(); i++) {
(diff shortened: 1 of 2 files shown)


eventdisplay/tracks/FairEveGeoTracks.cxx
```diff
@@ -49,8 +49,10 @@ InitStatus FairEveGeoTracks::Init()
 {
     FairRootManager *mngr = FairRootManager::Instance();
     fContainer = (TClonesArray *)mngr->GetObject("GeoTracks");
-    if (fContainer == nullptr)
-        return kFATAL;
+    if (fContainer == nullptr) {
```
dennisklein, line 52:
```suggestion
    if (!fContainer) {
```

Let's use this opportunity and apply [ES.87](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions). Thx!

examples/common/eventdisplay/FairEveMCTracks.cxx
```diff
@@ -149,8 +149,15 @@ InitStatus FairEveMCTracks::Init()
 {
     FairRootManager *mngr = FairRootManager::Instance();
     fContainer = (TClonesArray *)mngr->GetObject("MCTrack");
-    if (fContainer == nullptr)
-        return kFATAL;
+    if (fContainer == nullptr) {
```
dennisklein, line 152:
```suggestion
    if (!fContainer) {
```

Same as above, see [ES.87](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions).

examples/common/eventdisplay/FairEveMCTracks.cxx
```diff
@@ -149,8 +149,15 @@ InitStatus FairEveMCTracks::Init()
...
+        LOG(warning) << "MCTrack branch not found !";
+        return kERROR;
+    }
+    TClass *classTrack = fContainer->GetClass();
+    if (!classTrack->InheritsFrom("FairMCTrack")) {
```
dennisklein, line 157:
```suggestion
    if (classTrack && !classTrack->InheritsFrom("FairMCTrack")) {
```

Unless you can make a compelling argument, why this cannot return `nullptr` here?

dennisklein (changes requested):
See my comments. Otherwise, I approve the suggested changes. But better let's wait for one more approval by one of the experts.

eventdisplay/tracks/FairEveGeoTracks.cxx
```diff
@@ -49,8 +49,10 @@ InitStatus FairEveGeoTracks::Init()
 {
     FairRootManager *mngr = FairRootManager::Instance();
     fContainer = (TClonesArray *)mngr->GetObject("GeoTracks");
-    if (fContainer == nullptr)
-        return kFATAL;
+    if (!fContainer) {
+        LOG(warning) << "Geo tracks not found !";
```
fuhlig1, line 53: Could you please add a short explanation what will happen due to the absence of the Geo Tracks. Probably something like "deactivating task".

Please due the same also for the other error messages. It would help the users to understand the consequence.

fuhlig1 (commented):
The PR is okay for me after the log messages were updated.

examples/common/eventdisplay/FairEveMCTracks.cxx
```diff
@@ -149,8 +149,15 @@ InitStatus FairEveMCTracks::Init()
 {
     FairRootManager *mngr = FairRootManager::Instance();
     fContainer = (TClonesArray *)mngr->GetObject("MCTrack");
```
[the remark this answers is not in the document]
DanielWielanek: I decided to use dynamic_cast, I also added this into GeoTracks

ChristianTackeGSI (commented):
This looks good to me.

BUT:
* Someone should please verify that the GUI still works (it really should, but nevertheless).
* @DanielWielanek would you mind rebasing / squashing this into a single commit instead of six?

karabowi (approved):

dennisklein:
https://cdash.gsi.de/testDetails.php?test=12340355&build=347034 is an odd failure, the log output looks just fine. I have seen this already in the past, I suspect it is related to the CMake. However, I think we can exclude that it is cause by this PR at least ;)

PR #1230, opened 2022-09-30
Structure the repo into compile-time configurable components

dennisklein:
Attempt at adding missing naming consistency and repo structure to introduce a compile-time configurable component structure. Still very WIP will add more rationale later. But any early feedback welcome (I recommend to review by commit, the total changeset is too large).

proposed source tree conventions:
* one top-level directory per library
* directory name in lower-case
* library name in came-case, but same name as directory
* no "Fair" prefix, we are within the FairRoot source tree, it is redundant (exception wrapper target `FairMQ` where it stands as an atomic name)
* generate dictionary from just `LinkDef.h`, no need to prefix the library name here again
* example directory structure stays unchanged

proposed install tree changes:
* install library headers to `<prefix>/include/fairroot` (pure install tree technicality, the user interface stays untouched, meaning includes still done without dir prefix as before, because our headers are prefixed - `#include <Fair*.h>`)
* install example headers to `<prefix>/share/fairroot/examples/include`
* prefix library file names: `libfairroot-<library>.so` where `<library>` matches the directory name in the source tree (kebab-case lib naming seems popular, but it's a matter of taste, snake-case/camel-case also possible, important is the prefix - alternative to the prefix would be a directory `<prefix>/lib/fairroot/libBase.so` but this seems more configuration intensive). this breaks the current user interface, but with the v19 release we anyways require people to switch to cmake targets, which from then on hide the on-disk names from the user.

general changes:
* cmake project renamed to camel-case `FairRoot` to match cmake package name (and the github repo name and camel-case style known in our ecosystem of Fair* projects)
* usage of relative `PROJECT_*` family of variables in favor of explicit `FairRoot_*` (former `FAIRROOT_*`) for maintainability


ChristianTackeGSI:
Looking at the titles of some of the commits, I would prefer them to be in separate PRs, so that they can be reviewed/merged independently.

For example the one to rename the project to camelcase.
This doesn't look too breaking and would still fit well into the FairRoot 19 release.

Many other things need discussion, whether we want even more "breaking changes" in FairRoot 19, or whether we want to move those changes to FairRoot 20.
We could release FairRoot 20 four or six months after FairRoot 19? This gives people the opportunity to go step by step with the breaking changes.

dennisklein:
sure, i can put things into multiple PRs, just let me know what you want first.

The questions what content to release in which quantum is out of scope of this PR, or why do you ask it here?

dennisklein:
> For example the one to rename the project to camelcase.

#1232

PR #1236, opened 2022-10-11
Replace the YAML config with XML using FairXmlVMCConfig

karabowi:
Got rid of external dependency on yaml-cpp
by using the ROOT's native XML parser.
Replaced YAML config files with XML files.
Removed the YAML references in the main CMakeLists.txt. 
Replaced fSimulationConfig pointer with unique pointer.
Changed the macros, as well as some MQ devices
to use the FairVMCConfig.

---

Checklist:


examples/common/gconfig/FairVMCConfig.cxx
```diff
@@ -18,18 +18,26 @@
...
+        LOG(debug) << "Setting SIM:FairStack:StoreSecondaries to \"" << (ConvertTo<bool>(xmlNode) ? "true" : "false")
+                   << "\"";
+        stack->StoreSecondaries(ConvertTo<bool>(xmlNode));
+    }
+    if ((xmlNode = GetPointer("SIM:FairStack:MinPoints", fXmlConfig))) {
+        LOG(debug) << "Setting SIM:FairStack:MinPoints to \"" << ConvertTo<float>(xmlNode) << "\"";
+        stack->SetMinPoints(ConvertTo<float>(xmlNode));
+    }
+
     TVirtualMC::GetMC()->SetStack(stack);
```
dennisklein, line 41:
```suggestion
    auto stack = std::make_unique<FairStack>(1000);

    if (fXmlConfig) {
        if (auto const storeSecondariesNode = GetPointer("SIM:FairStack:StoreSecondaries", fXmlConfig);
            node)
        {
            auto const storeSecondaries = ConvertTo<bool>(storeSecondariesNode);
            LOG(debug) << "Setting SIM:FairStack:StoreSecondaries to \"" << (storeSecondaries ? "true" : "false")
                       << "\"";
            stack->StoreSecondaries(storeSecondaries);
        }
        if (auto const minPointsNode = GetPointer("SIM:FairStack:MinPoints", fXmlConfig);
            minPointsNode)
        {
            auto const minPoints = ConvertTo<float>(minPointsNode);
            LOG(debug) << "Setting SIM:FairStack:MinPoints to \"" << minPoints << "\"";
            stack->SetMinPoints(minPoints);
        }
    } else {
        // Handle this case, e.g. set defaults? warn, error whatever...
    }

    TVirtualMC::GetMC()->SetStack(stack.release());
```
1. Do not use owning raw pointer! (Specifically, nothing that is called here is `noexcept`, so `stack` would leak if an exception is thrown)
2. Handle the case when `fXmlConfig` is `nullptr`!
3. Call the `ConvertTo` functions only once!
4. Do not repeat the type name in the variable name! Prefer readable variable names.

---

The `FairXmlVMCConfig::GetPointer(..., XMLNodePointer_t, ...)` feels awkwardly low level to be used here. I suggest to add a more high-level API, e.g.
```cpp
template <typename T>
auto FairXmlVMCConfig::GetValue<T>(std::string const & path, bool verbose = false)
{
    if (fXmlConfig) {
        if (auto const node = GetPointer(path, fXmlConfig);
            node)
        {
            return ConvertTo<T>(node, verbose);
        }
    }

    // Handle this case, e.g. error, throw whatever...
}
```

then the above code simplifies to:

```cpp
#include <iomanip>   // for std::quoted
// ...


    auto const storeSecondaries = GetValue<bool>("SIM:FairStack:StoreSecondaries");
    auto const minPoints = GetValue<float>("SIM:FairStack:MinPoints");
    auto stack = std::make_unique<FairStack>(1000);

    LOG(debug) << "Setting SIM:FairStack:StoreSecondaries to " << std::quoted(storeSecondaries ? "true" : "false");
    stack->StoreSecondaries(storeSecondaries);

    LOG(debug) << "Setting SIM:FairStack:MinPoints to " << std::quoted(minPoints);
    stack->SetMinPoints(minPoints);

    TVirtualMC::GetMC()->SetStack(stack.release());
    // or, if we also add an overload `SetStack(std::unique_ptr<FairStack>)`
    TVirtualMC::GetMC()->SetStack(std::move(stack));
```

dennisklein (changes requested):
In general, the XMLIO module is very antiquated with lots of `void*`s. Not so sure, if we really wanted to adopt this in 2022 🤔

base/sim/FairGenericVMCConfig.h
```diff
@@ -19,6 +19,7 @@ class FairGenericVMCConfig
 {
   public:
     FairGenericVMCConfig();
+    FairGenericVMCConfig(const FairGenericVMCConfig& config);
     virtual ~FairGenericVMCConfig();
```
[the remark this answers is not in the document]
dennisklein: Please also apply [C.21](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all) to the new class `FairXmlVMCConfig`.

fairtools/MCConfigurator/FairXmlVMCConfig.h
```diff
@@ -0,0 +1,80 @@
...
+
+    string fMCEngine;
+
+  protected:
+    void SetupGeant3();
+    void SetupGeant4();
+    virtual void SetupStack() = 0;
+    void SetCuts();
+
+    TXMLEngine* fXmlEngine;
```
dennisklein, line 51: This object is leaking ...

No need for a heap object here anyways, let's keep it a plain simple value member.

```suggestion
    TXMLEngine fXmlEngine;
```

fairtools/MCConfigurator/FairXmlVMCConfig.cxx
```diff
@@ -0,0 +1,429 @@
...
+
+    if ((xmlNode = GetPointer("SIM:G4_PostInit:Commands", fXmlConfigPostInit))) {
+        std::vector<string> strVect = ConvertToVector<string>(xmlNode);
+        LOG(debug) << "FairXmlVMCConfig: SIM:G4_PostInit:Commands got " << strVect.size() << " values:";
+        TGeant4* geant4 = dynamic_cast<TGeant4*>(TVirtualMC::GetMC());
+        for (auto val : strVect) {
+            LOG(debug) << "   -> \"" << val << "\"";
+            geant4->ProcessGeantCommand(val.data());
+        }
+    }
```
dennisklein, line 132: * Please use descriptive names describing the variable content and **not** the type
* Make local variables `const` by default

```cpp
    if ((xmlNode = GetPointer("SIM:G4_PostInit:Commands", fXmlConfigPostInit))) {
        std::vector<string> const commands = ConvertToVector<string>(xmlNode);
        LOG(debug) << "FairXmlVMCConfig: SIM:G4_PostInit:Commands got " << commands.size() << " values:";
        TGeant4* geant4 = dynamic_cast<TGeant4*>(TVirtualMC::GetMC());
        for (auto const command : commands) {
            LOG(debug) << "   -> \"" << command << "\"";
            geant4->ProcessGeantCommand(command.data());
        }
    }
```

fairtools/MCConfigurator/FairXmlVMCConfig.cxx
```diff
@@ -0,0 +1,429 @@
...
+
+    XMLDocPointer_t xmldoc = fXmlEngine->ParseFile(ConfigMacro.Data());
+    if (!xmldoc)
+        return;
+
+    // take access to main node
+    fXmlConfigPostInit = fXmlEngine->DocGetRootElement(xmldoc);
+
+    XMLNodePointer_t xmlNode = 0;
```
dennisklein, line 122:
```suggestion
    XMLNodePointer_t xmlNode = nullptr;
```
See [ES.47](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es47-use-nullptr-rather-than-0-or-null)

fairtools/MCConfigurator/FairXmlVMCConfig.cxx
```diff
@@ -0,0 +1,429 @@
...
+        LOG(fatal) << "FairXmlVMCConfig::Setup() Engine \"" << mcEngine << "\" unknown!";
+    }
+
+    string xmlFileName = ObtainXmlFileName(mcEngine);
+
+    // Parse xml file
+    // Only file with restricted xml syntax are supported
+    XMLDocPointer_t xmldoc = fXmlEngine->ParseFile(xmlFileName.data());
+    if (!xmldoc)
+        return;
```
dennisklein, line 55: How does this work out? If parsing failed, we just silently skip the rest without any error? 🤯

fairtools/MCConfigurator/FairXmlVMCConfig.cxx
```diff
@@ -0,0 +1,429 @@
...
+    return configFileWithPath.Data();
+}
+
+void FairXmlVMCConfig::StoreXmlInfo()
+{
+    std::ostringstream nodestring;
+    nodestring << fMCEngine << "\n";
+    nodestring << fXmlConfig;
+    nodestring << "\n";
+    TObjString* configObject = new TObjString(nodestring.str().c_str());
```
dennisklein, line 356: Who owns `configObject`?

fairtools/MCConfigurator/FairXmlVMCConfig.cxx
```diff
@@ -0,0 +1,429 @@
...
+    FairRunSim::Instance()->GetSink()->WriteObject(configObject, "SimulationSetup", TObject::kSingleKey);
+
+    LOG(debug) << "FairXmlVMCConfig::StoreXmlInfo() done.";
+}
+
+XMLNodePointer_t FairXmlVMCConfig::GetPointer(string path, XMLNodePointer_t node, int verbose)
```
dennisklein, line 362: * The naming does not help at all, maybe rename to `GetXmlNode` or `QueryXmlNode`?
* This function needs a big multiline comment documenting what it does. Especially, how the `path` argument is defined!

fairtools/MCConfigurator/FairXmlVMCConfig.cxx
```diff
@@ -0,0 +1,429 @@
...
+{
+    // this function display all accessible information about xml node and its children
+    if (verbose)
+        cout << "GetPointer with path \"" << path << "\"" << endl;
+    path.replace(0, path.find(':') + 1, "");
```
dennisklein, line 367: It immediately discards the first element from the `path`? Meaning it just accepts any root element?

fairtools/MCConfigurator/FairXmlVMCConfig.cxx
```diff
@@ -0,0 +1,429 @@
...
+
+    XMLNodePointer_t child = fXmlEngine->GetChild(node);
+    while (child != 0) {
+        if (verbose)
+            cout << "got child with name " << fXmlEngine->GetNodeName(child) << " and the path is \"" << path << "\""
+                 << endl;
+        if (path.find(fXmlEngine->GetNodeName(child)) != 0) {
```
dennisklein, line 374: `std::string::find` matches any occurrence the search term, but what you want, is to match only everything in front of the first `:` separator, right?

dennisklein: ok, ignore that, i see it now

fairtools/MCConfigurator/FairXmlVMCConfig.cxx
dennisklein: `path.find(':') + 1` this also deserves a comment IMHO that the integer overflow is intended (if it is).

fairtools/MCConfigurator/FairXmlVMCConfig.cxx
dennisklein: The way this function is implemented right now, would basically be unmaintainable on arrival. This needs to be rewritten. I cannot approve this in this form. It lacks
* documentation as stated above, and
* clear, readable recursion and terminating cases.

Also, this would be a prime use case for [`std::string_view`](https://en.cppreference.com/w/cpp/string/basic_string_view) to avoid all those string copies and allocations.

CMakeLists.txt
```diff
@@ -222,22 +222,6 @@ set(FAIRROOT_LIBRARY_PROPERTIES ${FAIRROOT_LIBRARY_PROPERTIES}
...
-if(    yaml-cpp_FOUND
-   AND TARGET yaml-cpp
-   AND EXISTS YAML_CPP_INCLUDE_DIR
-   AND yaml-cpp_VERSION VERSION_LESS 0.6.3
-)
-  set_target_properties(yaml-cpp PROPERTIES
-    INTERFACE_INCLUDE_DIRECTORIES "${YAML_CPP_INCLUDE_DIR}"
-  )
-endif()
-
```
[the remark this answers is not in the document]
karabowi: Let's discuss the possible options in person.

karabowi:
Closed after redesign decisions.

PR #1240, opened 2022-10-13
FMCApp: Do not Leak FairRunSim

ChristianTackeGSI:
CloneForWorker creates a FairRunSim. But this was not cleaned up anywhere.

So clean it up during destruction of FairMCApplication.

Also use the fRun member variable more instead of FairRun{,Sim}::Instance.

---

Checklist:

### DIFF
--- a/base/sim/FairMCApplication.cxx
+++ b/base/sim/FairMCApplication.cxx
@@ -64,7 +64,7 @@ class TParticle;
 #include <float.h>    // for DBL_MAX
 #include <mutex>      // std::mutex
 #include <stdlib.h>   // for getenv, exit
-#include <utility>    // for pair
+#include <utility>    // for pair, move
 std::mutex mtx;       // mutex for critical section
 
 using std::pair;
@@ -103,11 +103,12 @@ FairMCApplication::FairMCApplication(const char* name, const char* title, TObjAr
     , listActiveDetectors()
     , listDetectors()
     , fMC(nullptr)
-    , fRun(nullptr)
+    , fRun(FairRunSim::Instance())
     , fSaveCurrentEvent(kTRUE)
     , fState(FairMCApplicationState::kUnknownState)
     , fRunInfo()
     , fGeometryIsInitialized(kFALSE)
+    , fWorkerRunSim()
 {
     // Standard Simulation constructor
 
@@ -115,7 +116,6 @@ FairMCApplication::FairMCApplication(const char* name, const char* title, TObjAr
 
     LOG(debug) << "FairMCApplication-ctor " << this;
 
-    fRun = FairRunSim::Instance();
     fModules = ModList;
     fModIter = fModules->MakeIterator();
     // Create and fill a list of active detectors
@@ -148,7 +148,7 @@ FairMCApplication::FairMCApplication(const char* name, const char* title, TObjAr
 }
 
 //_____________________________________________________________________________
-FairMCApplication::FairMCApplication(const FairMCApplication& rhs)
+FairMCApplication::FairMCApplication(const FairMCApplication& rhs, std::unique_ptr<FairRunSim> otherRunSim)
     : TVirtualMCApplication(rhs.GetName(), rhs.GetTitle())
     , fActiveDetectors(nullptr)
     , fFairTaskList(nullptr)
@@ -181,16 +181,19 @@ FairMCApplication::FairMCApplication(const FairMCApplication& rhs)
     , listActiveDetectors()
     , listDetectors()
     , fMC(nullptr)
-    , fRun(nullptr)
     , fSaveCurrentEvent(kTRUE)
     , fState(FairMCApplicationState::kUnknownState)
     , fRunInfo()
     , fGeometryIsInitialized(kFALSE)
+    , fWorkerRunSim(std::move(otherRunSim))
 {
     // Copy constructor
     // Do not create Root manager
 
     LOG(debug) << "FairMCApplication-copy-ctor " << this;
+
+    fParent = &rhs;
+    fRun = fWorkerRunSim.get();
 
     // Create an ObjArray of Modules and its iterator
     fModules = new TObjArray();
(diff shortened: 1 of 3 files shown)


base/sim/FairMCApplication.cxx
```diff
@@ -490,11 +497,7 @@ TVirtualMCApplication* FairMCApplication::CloneForWorker() const
     }
 
     // Create new FairMCApplication object on worker
-    FairMCApplication* workerApplication = new FairMCApplication(*this);
-    workerApplication->SetGenerator(fEvGen->ClonePrimaryGenerator());
```
rbx, line 494: `SetGenerator()` just goes away? You don't call it anywhere else?

rbx: Ok, I see you're just assigning directly.

rbx:
From `fRun(FairRunSim::Instance())` I assume that `FairRunSim` is a singleton. Why is there a copy of it needed for the "worker"?

dennisklein:
> From `fRun(FairRunSim::Instance())` I assume that `FairRunSim` is a singleton. Why is there a copy of it needed for the "worker"?

Because of this magic ...

https://github.com/FairRootGroup/FairRoot/blob/a3f91e09509988ca1d8532e712d246df0c4d4887/base/steer/FairRunSim.cxx#L367

ChristianTackeGSI:
I have tried to
* keep all the current "logic" as it was.
* move some things into nicer places (use the special copy constructor for more stuff)
* finally clean up the (copied) FairRunSim

As I don't yet fully understand the whole picture, that seemed the safest way to go.

rbx:
> > From `fRun(FairRunSim::Instance())` I assume that `FairRunSim` is a singleton. Why is there a copy of it needed for the "worker"?
> 
> Because of this magic ...
> 
> https://github.com/FairRootGroup/FairRoot/blob/a3f91e09509988ca1d8532e712d246df0c4d4887/base/steer/FairRunSim.cxx#L367

Doesn't that create a copy for each thread "automatically"?

rbx:
> I have tried to
> 
>     * keep all the current "logic" as it was.
> 
>     * move some things into nicer places (use the special copy constructor for more stuff)
> 
>     * finally clean up the (copied) FairRunSim
> 
> 
> As I don't yet fully understand the whole picture, that seemed the safest way to go.

I guess a change in the logic that I see is that before this class was asking for an instance of FairRunSim, while now it will ask for the copied object through fRun? It is still not very clear to me if those are the same thing. Or if they will stay the same thing if the implementation of FairRunSim changes.

dennisklein:
> > > From `fRun(FairRunSim::Instance())` I assume that `FairRunSim` is a singleton. Why is there a copy of it needed for the "worker"?
> > 
> > 
> > Because of this magic ...
> > https://github.com/FairRootGroup/FairRoot/blob/a3f91e09509988ca1d8532e712d246df0c4d4887/base/steer/FairRunSim.cxx#L367
> 
> Doesn't that create a copy for each thread "automatically"?

Unfortunately not, see ctor

https://github.com/FairRootGroup/FairRoot/blob/a3f91e09509988ca1d8532e712d246df0c4d4887/base/steer/FairRunSim.cxx#L49-L87

and "singleton" getter

https://github.com/FairRootGroup/FairRoot/blob/a3f91e09509988ca1d8532e712d246df0c4d4887/base/steer/FairRunSim.cxx#L116

It is not a classic Singleton pattern from the text book. The ctors are public, so a user can create multiple instances still (and is required to call a ctor first), but somehow all but the first one (per thread ;) ) are kinda broken...

ChristianTackeGSI:
> I guess a change in the logic that I see is that before this class was asking for an instance of FairRunSim, while now it will ask for the copied object through fRun (in places where you use the fRun member variable instead of FairRun{,Sim}::Instance)? It is still not very clear to me if those are the same thing. Or if they will stay the same thing if the implementation of FairRunSim changes.

For threading:
* Old:
  * CloneForWorker runs in a new thread. No "instance" of FairRunSim in that thread. So it did a `new FairRunSim` (`::Instance` will return that in this thread).
  * CloneForWorker copies over some stuff (especially the MC name).
  * All code in this thread gets the new FairRunSim via `::Instance`.
  * New FairRunSim being leaked
* New:
  * CloneForWorker creates a new FairRunSim and makes sure that `fRun` points to it
  * CloneForWorker copies things over
  * All the code in the thread gets the new FairRunSim via `fRun`
  * When the FMCApp is cleaned up, the FairRunSim is also cleaned up.

For "classic":
* Old: There is only one FairRunSim. `::Instance` returns it.
* New: There is still the same one FairRunSim. Just `fRun` points to it (also).

Possible problems:
* Only if code starts in one thread and continues in another. But I guess, that was problematic before as well?

rbx:
Thanks for the clear explanation!

I don't see problems with this PR.

But this thread local + weird singleton combo seems crazy, fragile and dangerous. If we can, we should try to isolate it as much as possible from the user and from the rest of the framework 🤯. (or get rid of it 😉)

rbx (approved):

ChristianTackeGSI:
> But this thread local + weird singleton combo seems crazy, fragile and dangerous. If we can, we should try to isolate it as much as possible from the user and from the rest of the framework exploding_head. (or get rid of it wink)

That's, why I am running around and trying to replace `Foo::Instance` by something else, trying to mark `::Instance` as deprecated, etc…

karabowi:
The CloneForWorkers is needed solely by the Geant4MT, which currently does not work, so testing is difficult.

However, when I run `root -l -q 'run_tutorial1.C(10,"TGeant4",true)'` with your branch, the root session ends with segmentation violation, which is not the case with FairRoot dev.
Can you test it?

ChristianTackeGSI:
`run_tutorial1.C(10,"TGeant4",true)` works now for me.

karabowi (approved):

PR #1245, opened 2022-10-21
Add testing Geant4 Multithreaded

karabowi:
Check the output of the Geant4MT mode.

---

Checklist:

### DIFF
--- a/base/event/FairMCEventHeader.h
+++ b/base/event/FairMCEventHeader.h
@@ -82,21 +82,7 @@ class FairMCEventHeader : public TNamed
     void Reset();
 
     /** Clone this object (used in MT mode only) */
-    FairMCEventHeader* CloneMCEventHeader() const
-    {
-        FairMCEventHeader* clonedHeader = new FairMCEventHeader();
-        clonedHeader->SetRunID(fRunId);
-        clonedHeader->SetEventID(fEventId);
-        clonedHeader->SetVertex(fX, fY, fZ);
-        clonedHeader->SetTime(fT);
-        clonedHeader->SetB(fB);
-        clonedHeader->SetNPrim(fNPrim);
-        clonedHeader->MarkSet(fIsSet);
-        clonedHeader->SetRotX(fRotX);
-        clonedHeader->SetRotY(fRotY);
-        clonedHeader->SetRotZ(fRotZ);
-        return clonedHeader;
-    }
+    FairMCEventHeader* CloneMCEventHeader() const { return new FairMCEventHeader(*this); }
 
     /** Register the class as data branch to the output */
     virtual void Register();
--- a/base/event/FairMCEventHeader.h
+++ b/base/event/FairMCEventHeader.h
@@ -81,6 +81,23 @@ class FairMCEventHeader : public TNamed
     /** Reset all members **/
     void Reset();
 
+    /** Clone this object (used in MT mode only) */
+    FairMCEventHeader* CloneMCEventHeader() const
+    {
+        FairMCEventHeader* clonedHeader = new FairMCEventHeader();
+        clonedHeader->SetRunID(fRunId);
+        clonedHeader->SetEventID(fEventId);
+        clonedHeader->SetVertex(fX, fY, fZ);
+        clonedHeader->SetTime(fT);
+        clonedHeader->SetB(fB);
+        clonedHeader->SetNPrim(fNPrim);
+        clonedHeader->MarkSet(fIsSet);
+        clonedHeader->SetRotX(fRotX);
+        clonedHeader->SetRotY(fRotY);
+        clonedHeader->SetRotZ(fRotZ);
+        return clonedHeader;
+    }
+
     /** Register the class as data branch to the output */
     virtual void Register();
 
--- a/base/sim/FairMCApplication.cxx
+++ b/base/sim/FairMCApplication.cxx
@@ -233,6 +233,8 @@ FairMCApplication::FairMCApplication(const FairMCApplication& rhs, std::unique_p
         fEvGen = rhs.fEvGen->ClonePrimaryGenerator();
     }
 
+    fMCEventHeader = rhs.fMCEventHeader->CloneMCEventHeader();
+
     // Create a Task list
     // Let's try without it
     // fFairTaskList= new FairTask("Task List", 1);
@@ -488,6 +490,7 @@ TVirtualMCApplication* FairMCApplication::CloneForWorker() const
     auto workerRun = std::make_unique<FairRunSim>(kFALSE);
     workerRun->SetName(fRun->GetName());   // Transport engine
     workerRun->SetSink(std::unique_ptr<FairSink>{fRun->GetSink()->CloneSink()});
+    workerRun->SetMCEventHeader(fRun->GetMCEventHeader()->CloneMCEventHeader());
 
     // Trajectories filter is created explicitly as we do not call
     // FairRunSim::Init on workers
@@ -900,7 +903,6 @@ void FairMCApplication::RegisterOutput()
     } else {
         LOG(warn) << "Stack is not registered ";
     }
-
     /** SetSpecialPhysicsCuts for FairDetector objects and all passive modules inheriting from FairModule */
     // initialize and register FairDetector objects in addition
     // Note: listActiveDetectors or fActiveDetectors not used to include passive modules in the same loop.
@@ -923,6 +925,13 @@ void FairMCApplication::RegisterOutput()
         }
     }
     fModIter->Reset();
+
+    if (fRootManager) {
+        fMCEventHeader->Register();
+    }
+    if (fEvGen) {
+        fEvGen->SetEvent(fMCEventHeader);
+    }
 }
 
 //_____________________________________________________________________________
@@ -943,19 +952,6 @@ void FairMCApplication::InitGeometry()
     Int_t id = 0;
     fModIter->Reset();
 
-    // Register stack
-    if (fRootManager && !fParent) {
-        RegisterOutput();
-    }
-
-    /**Tasks has to be initialized here, they have access to the detector branches and still can create objects in the
-     * tree*/
-    /// There is always a Main Task  !
-    /// so .. always a InitTasks() is called <D.B>
-    if (fFairTaskList) {
-        InitTasks();
-    }
-
     // store the EventHeader Info
     // Get and register EventHeader
     UInt_t runId = fRun->GetRunId();
@@ -965,13 +961,7 @@ void FairMCApplication::InitGeometry()
     // Get and register the MCEventHeader
     fMCEventHeader = fRun->GetMCEventHeader();
     fMCEventHeader->SetRunID(runId);
-    if (fRootManager) {
-        fMCEventHeader->Register();
-    }
 
-    if (fEvGen) {
-        fEvGen->SetEvent(fMCEventHeader);
-    }
     fTrajFilter = FairTrajFilter::Instance();
     if (nullptr != fTrajFilter) {
         fTrajFilter->Init();
@@ -986,9 +976,18 @@ void FairMCApplication::InitGeometry()
         fRadGridMan->Init();
     }
 
-    // save Geo Params in Output file
+    // Register output
     if (fRootManager && !fParent) {
+        RegisterOutput();
         fRootManager->WriteFolder();
+    }
+
+    /**Tasks has to be initialized here, they have access to the detector branches and still can create objects in the
+     * tree*/
+    /// There is always a Main Task  !
+    /// so .. always a InitTasks() is called <D.B>
+    if (fFairTaskList) {
+        InitTasks();
     }
 
     // Get static thread local svList
--- a/base/sim/FairMCApplication.cxx
+++ b/base/sim/FairMCApplication.cxx
@@ -485,6 +485,11 @@ TVirtualMCApplication* FairMCApplication::CloneForWorker() const
     std::lock_guard guard(mtx);
     LOG(info) << "FairMCApplication::CloneForWorker ";
 
+    if (fRun->GetNTasks() >= 1) {
+        LOG(fatal) << "Adding tasks to simulation in multithreaded mode currently unavailable. Contact FairRoot "
+                      "developers for possible solutions.";
+    }
+
     // Create new FairRunSim object on worker
     // and pass some data from master FairRunSim object
     auto workerRun = std::make_unique<FairRunSim>(kFALSE);
--- a/base/sim/FairMCApplication.cxx
+++ b/base/sim/FairMCApplication.cxx
@@ -495,7 +495,6 @@ TVirtualMCApplication* FairMCApplication::CloneForWorker() const
     auto workerRun = std::make_unique<FairRunSim>(kFALSE);
     workerRun->SetName(fRun->GetName());   // Transport engine
     workerRun->SetSink(std::unique_ptr<FairSink>{fRun->GetSink()->CloneSink()});
-    workerRun->SetMCEventHeader(fRun->GetMCEventHeader()->CloneMCEventHeader());
 
     // Trajectories filter is created explicitly as we do not call
     // FairRunSim::Init on workers
--- a/base/sim/FairMCApplication.cxx
+++ b/base/sim/FairMCApplication.cxx
@@ -57,7 +57,6 @@
 #include <TROOT.h>              // for TROOT, gROOT
 #include <TRefArray.h>          // for TRefArray
 #include <TSystem.h>            // for TSystem, gSystem
-#include <TVirtualMC.h>         // for TVirtualMC
 #include <TVirtualMCStack.h>    // for TVirtualMCStack
 class TParticle;
 
@@ -512,7 +511,12 @@ void FairMCApplication::InitOnWorker()
     // and create a new sink on worker
     // moved to CloneForWorker and specific sink->CloneSink();
 
-    fRootManager->InitSink();
+    {
+        std::lock_guard guard(mtx);
+        fRootManager->InitSink();
+        RegisterOutput();
+        fRootManager->WriteFolder();
+    }
 
     // Cache thread-local gMC
     fMC = gMC;
@@ -749,7 +753,10 @@ Double_t FairMCApplication::TrackingZmax() const
     return DBL_MAX;
 }
 //_____________________________________________________________________________
-void FairMCApplication::SetField(FairField* field) { fxField = field; }
+void FairMCApplication::SetField(FairField* field)
+{
+    fxField = field;
+}
 //_____________________________________________________________________________
 void FairMCApplication::ConstructOpGeometry()
 {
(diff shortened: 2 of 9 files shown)


examples/simulation/Tutorial1/macros/run_tutorial1.C
```diff
@@ -160,4 +150,27 @@ void run_tutorial1(Int_t nEvents = 10,
     cout << "Macro finished successfully." << endl;
 
     // ------------------------------------------------------------------------
+
+    auto file = TFile::Open(outFile.data());
+    auto tree = static_cast<TTree*>(file->Get("cbmsim"));
```
ChristianTackeGSI, line 155: Could you try this?

```suggestion
    auto tree = file->Get<TTree*>("cbmsim");
```

examples/simulation/Tutorial1/macros/run_tutorial1.C
```diff
@@ -160,4 +150,27 @@ void run_tutorial1(Int_t nEvents = 10,
     cout << "Macro finished successfully." << endl;
 
     // ------------------------------------------------------------------------
+
+    auto file = TFile::Open(outFile.data());
```
ChristianTackeGSI, line 154: Could you try this?

```suggestion
    std::unique_ptr<TFile> file{TFile::Open(outFile.c_str()};
```

examples/simulation/Tutorial1/macros/run_tutorial1.C
```diff
@@ -160,4 +150,27 @@ void run_tutorial1(Int_t nEvents = 10,
...
+    if (nofT < nEvents * nofPart) {
+        cout << "Not enough tracks (" << nofT << " < " << nEvents * nofPart << ") in the output tree." << endl;
+        return 1;
+    }
+    if (nofP < nEvents * nofPart) {
+        cout << "Not enough points (" << nofP << " < " << nEvents * nofPart << ") in the output tree." << endl;
+        return 1;
+    }
+    cout << "Simulation successful." << endl;
+    file->Close();
```
ChristianTackeGSI, line 173:
```suggestion
```

This is probably not needed, because the unique_ptr will destruct (and thus close) the file, hopefully.

ChristianTackeGSI (commented):
I like new test coverage!

I added a few comments.

I haven't looked in detail, why some tests on some machines fail, while other work out.

It *might* be due to `.data()` instead of `.c_str()`. Because `data` returns a pointer to some internal data that might not be NUL-terminated. While `c_str` is guaranteed to be NUL-terminated.

Could we first fix that and then take a deeper looker at the failures?

karabowi:
Hi @ChristianTackeGSI, thanks for the quick comments.

I expect the tests to fail on all the G4MT installations of FairSoft.
Two days ago I have discovered, that G4MT&FairRoot does not save any data to the output tree.
I have not been able to find the reason why, so I've decided to write this short test.

ChristianTackeGSI:
Okay, wow! Now looking at it. Only the none MT FairSoft builds seem to work.

Well, one non-MT build also fails. I don't understand that either.

Are we using Sinks to write the output?

I guess, one reason could be FairRun in the threads: We actually create a new FairRun for the thread. And that one does not have any Source/Sink connected to it. See #1240 for example, which works this out even more.

Maybe we should stop creating a new FairRun in the threads and use one for all threads? Just pondering. Still I don't understand all the complete design.

ChristianTackeGSI:
Some first ideas, so that we can continue:

* Maybe you comment out the `return 1` from all the error cases. That way the test still succeeds (for the CI run), but we have the test in CDash and can still view it and observe differences, if we start to fix things.

* Then we could merge it and have easier test coverage for everyone on this.

* Then I will rebase #1240 on it and we'll see, wether it changes anything at all. If not, we could maybe merge it?

* And the we could work further from there and analyze more things?

karabowi:
I have changed the return value.
I think we can safely merge it and try the #1240 fix on it.
The only drawback will be the failing FairRoot test on MT FairSofts.

ChristianTackeGSI:
Could you please at least fix all the `.data()` to `.c_str()`?

ChristianTackeGSI:
Hi @karabowi,

I added a commit to your branch that has most of my suggestions in it. Please take a look. If you like it, please squash it into your commits. If you don't, drop it.

I hope it was okay to add it that way and let the CI test the suggestions.

base/sim/FairMCApplication.cxx
```diff
@@ -419,6 +419,7 @@ void FairMCApplication::FinishRun()
         UndoGeometryModifications();
     }
 
+    FairRunSim::Instance()->SetIsMT(fMC->IsMT());
```
ChristianTackeGSI, line 422: Could you check, whether it is possible to move this into `CloneForWorker`?

```c++
    workerRun->SetIsMT(fRun->IsMT());
```

base/sim/FairMCApplication.cxx
```diff
@@ -512,6 +513,7 @@ void FairMCApplication::InitOnWorker()
     // and create a new sink on worker
     // moved to CloneForWorker and specific sink->CloneSink();
 
+    mtx.lock();
```
ChristianTackeGSI, line 516: I am not so happy, that you lock a mutex in `InitOnWorker` and unlock it in `InitGeometry`. What if things break in between and `InitGeometry` is never called?

To be honest, I don't understand, what you're protecting. If each thread has its own TFile, then there shouldn't be any races?

Anyway: Could you limit the area that needs locking to a very located area maybe?

base/sink/FairRootFileSink.cxx
```diff
@@ -107,8 +107,7 @@ Bool_t FairRootFileSink::InitSink()
 
     // FairRun* fRun = FairRun::Instance();
     /**Check if a simulation run!*/
-    fOutFolder = gROOT->GetRootFolder()->AddFolder(
-        Form("%s_%d", FairRootManager::GetFolderName(), FairRootManager::Instance()->GetInstanceId()), "Main Folder");
+    fOutFolder = gROOT->GetRootFolder()->AddFolder(Form("%s", FairRootManager::GetFolderName()), "Main Folder");
```
ChristianTackeGSI, line 110: From what I understand, this changes an API towards the user?
The generated Folders (and below the trees) now have different names?

Could you add a small entry to the "Breaking Changes" section in the changelog please?

examples/simulation/Tutorial1/macros/run_tutorial1.C
```diff
@@ -160,4 +158,34 @@ void run_tutorial1(Int_t nEvents = 10,
     cout << "Macro finished successfully." << endl;
 
     // ------------------------------------------------------------------------
+
+    TChain chain("cbmsim");
+    if (run->IsMT()) {
+        cout << endl << "MT mode, reading two files: t1 and t2" << endl;
+        chain.Add((outFile.replace(outFile.find(".root"), 5, "_t1.root")).c_str());
+        chain.Add((outFile.replace(outFile.find("_t1.root"), 8, "_t2.root")).c_str());
```
ChristianTackeGSI, line 166: Is Geant4 always creating two threads? Just wondering.

base/sink/FairRootFileSink.cxx
ChristianTackeGSI: Okay, I start to get it, after reading the docs on TFolder.

Maybe this code should be locked with a mutex, and check wether the Folder already exists and only create it, if it does not yet exist?

Well, still I feel like I am missing something.

examples/simulation/Tutorial1/macros/run_tutorial1.C
karabowi: Currently yes, no idea if there are plans for more...

base/sink/FairRootFileSink.cxx
karabowi: The TFolder may (and in FairRoot is) be used to automatically create the tree structure. We used to create folders with different names for the thread workers, but then created trees are not read properly when TChain'ed.

base/sim/FairMCApplication.cxx
karabowi: This change passes the actual setting of the flag from the `TVirtualMC` to the user.
Even though the user might have requested to run in the MT mode, it is not certain that this setting was actually applied (f.e. Geant4 compiled in single threaded mode, or running with TGeant3).
Therefore it cannot be called in `CloneForWorker` as this function is only called in Geant4MT.

ChristianTackeGSI: If we want to inform the user, what has happened really, I would propose to have a different channel from the "what the user would like to happen".

base/sim/FairMCApplication.cxx
```diff
@@ -515,6 +516,9 @@ void FairMCApplication::InitOnWorker()
 
     mtx.lock();
     fRootManager->InitSink();
+    RegisterOutput();
+    fRootManager->WriteFolder();
+    mtx.unlock();
```
ChristianTackeGSI, line 521: Consider using `std::lock_guard`, please. It unlocks the mutex automatically when leaving the block (even in case of errors or exceptions).

```suggestion
    {
        std::lock_guard guard(mtx);
        fRootManager->InitSink();
        RegisterOutput();
        fRootManager->WriteFolder();
    }
```

base/sim/FairMCApplication.cxx
```diff
@@ -936,6 +922,30 @@ void FairMCApplication::InitGeometry()
...
+
+    /// Initialize geometry
+
+    /** Register stack and detector collections*/
+    FairVolume* fv = 0;
+    Int_t id = 0;
+    fModIter->Reset();
+
+    // Register stack
+    if (fRootManager && fRootManager->GetInstanceId() == 0) {
```
ChristianTackeGSI, line 946: I assume, your check for the InstanceId is for checking that this is the main thread or something? What about using `fParent` for this?

```suggestion
    if (fRootManager && !fParent) {
```

base/sim/FairMCApplication.h
```diff
@@ -115,6 +115,8 @@ class FairMCApplication : public TVirtualMCApplication
     TTask* GetListOfTasks();
     FairGenericStack* GetStack();
     TChain* GetChain();
+    /** Register output */
+    void RegisterOutput();
```
ChristianTackeGSI, line 119: If this is only needed internally, could you consider moving it into the `private:` section?

ChristianTackeGSI (commented):
Things look good to me.
I have added a few smaller things to consider.

As this now changes some internal things in FairRoot, someone more knowledgeable about those things might have a second look? @fuhlig1 maybe?

examples/simulation/Tutorial1/macros/run_tutorial1.C
fuhlig1: @karabowi,

does it make sense to have the check in the same macro? Wouldn't it be better to have the check in another macro?

karabowi: @fuhlig1 the checking of the output file belongs to this macro, in my opinion.
Why should we create another macro just to check the output file?

base/sink/FairRootFileSink.cxx
```diff
@@ -262,13 +261,11 @@ void FairRootFileSink::WriteFolder()
...
-        // FairRun* fRun = FairRun::Instance();
-        fOutTree =
-            new TTree(FairRootManager::GetTreeName(),
-                      Form("/%s_%d", FairRootManager::GetFolderName(), FairRootManager::Instance()->GetInstanceId()),
-                      99);
+        fOutTree = new TTree(FairRootManager::GetTreeName(), Form("/%s", FairRootManager::GetFolderName()), 99);
+
         TruncateBranchNames();
         CreatePersistentBranchesAny();
+        gROOT->GetRootFolder()->Remove(fOutFolder);
```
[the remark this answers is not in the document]
karabowi: done

base/sim/FairMCApplication.cxx
ChristianTackeGSI: @karabowi Would you like to comment on this?

base/sink/FairRootFileSink.cxx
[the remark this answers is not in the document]
fuhlig1: Thanks.

examples/simulation/Tutorial1/macros/run_tutorial1.C
fuhlig1: In my opinion you check here two things at once. First you check if you are able to run the multithreaded Geant4 and second you check if you can read the results of the simulation. For me this are two distinct tests but I am also fine to keep the macro as it is.

base/sim/FairMCApplication.cxx
karabowi: Is that enough?
` // Pass on the actual MT flag from the VMC to the user
    FairRunSim::Instance()->SetIsMT(fMC->IsMT());`

ChristianTackeGSI: Could we have a new member variable like `MTWasUsed` on FairRunSim please?

From a user point of view, it is not obvious, that FairRunSim::IsMT() before a Run gives the indended MT flag, while after the run it gives the actual one.

If you *really* want this behaviour, please write Doxygen documentation at the IsMT() method, please.

ChristianTackeGSI: Generally looks good, resolved.

base/steer/FairRunSim.h
```diff
@@ -232,6 +242,7 @@ class FairRunSim : public FairRun
     TString fUserConfig;                            //!                        /** Macro for geant configuration*/
     TString fUserCuts;                              //!                          /** Macro for geant cuts*/
     Bool_t fIsMT;                                   //!                              /** MT mode option (Geant4 only)*/
+    Bool_t fWasMT;                                  //!                              /** Actual MT mode used */
```
ChristianTackeGSI, line 245: Maybe?

```suggestion
    bool fWasMT{false};                             //!                              /** Actual MT mode used */
```

And I think, you might even move that to the `private:` section?

ChristianTackeGSI (commented):
Looks very good to me. Just some small ideas.

base/steer/FairRunSim.h
```diff
@@ -186,8 +186,18 @@ class FairRunSim : public FairRun
...
+     */
     void SetIsMT(Bool_t isMT) { fIsMT = isMT; }
+    /**
+     * Read the MultiThreaded mode
+     */
     Bool_t IsMT() const { return fIsMT; }
+    /**
+     * Read the used transport MultiThreaded mode
+     */
+    Bool_t WasMT() const { return fWasMT; }
```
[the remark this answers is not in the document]
fuhlig1: I don't  think it is a good idea to use auto in an interface. How should I know which type to expect without reading the code?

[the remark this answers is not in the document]
ChristianTackeGSI: Yeah, you're right, as I also suggest to make `fWasMT` private and private member variables do not currently end up in Doxygen.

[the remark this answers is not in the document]
ChristianTackeGSI: Then please:

```suggestion
    bool WasMT() const { return fWasMT; }
```

base/steer/FairRunSim.h
ChristianTackeGSI: Please initialize `fWasMT` (probably to `false`), so that `WasMT()` returns a defined value even before a run?

base/steer/FairRunSim.cxx
```diff
@@ -72,8 +72,9 @@ FairRunSim::FairRunSim(Bool_t isMaster)
     , fMeshList(new TObjArray())
     , fUserConfig("")
     , fUserCuts("SetCuts.C")
-    , fIsMT(kFALSE)
     , fImportTGeoToVMC(kTRUE)
+    , fIsMT(false)
+    , fWasMT(false)
```
ChristianTackeGSI, line 77: Just move the simple initialization over into the `.h`. See below.

```suggestion
```

base/steer/FairRunSim.h
```diff
@@ -205,6 +215,8 @@ class FairRunSim : public FairRun
     void SetMCConfig();
     void CheckFlukaExec();
 
+    bool fWasMT;   //!                              /** Actual MT mode used */
```
ChristianTackeGSI, line 218:
```suggestion
    bool fWasMT{false};   //!                              /** Actual MT mode used */
```

base/steer/FairRunSim.h
```diff
@@ -231,7 +243,7 @@ class FairRunSim : public FairRun
     TObjArray* fMeshList;                           //!                          /** radiation grid scoring
     TString fUserConfig;                            //!                        /** Macro for geant configuration*/
     TString fUserCuts;                              //!                          /** Macro for geant cuts*/
-    Bool_t fIsMT;                                   //!                              /** MT mode option (Geant4 only)*/
+    bool fIsMT;                                     //!                              /** MT mode option (Geant4 only)*/
```
ChristianTackeGSI, line 246:
```suggestion
    bool fIsMT{false};                              //!                              /** MT mode option (Geant4 only)*/
```

ChristianTackeGSI (commented):
Please let's start to move simple initializers over to the `.h`.

This even avoids warnings like this:

> initializer order does not match the declaration order

(see also C++ Core Guideliens [C.47](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-order))

ChristianTackeGSI (dismissed):
Looks good to me now!

Don't know, if anybody else would like to, or should look at this again?

base/sim/FairMCApplication.cxx
```diff
@@ -234,6 +233,8 @@ FairMCApplication::FairMCApplication(const FairMCApplication& rhs, std::unique_p
         fEvGen = rhs.fEvGen->ClonePrimaryGenerator();
     }
 
+    fMCEventHeader = rhs.fMCEventHeader->CloneMCEventHeader();
```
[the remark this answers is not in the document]
ChristianTackeGSI: Can you now use
```cpp
fMCEventHeader = fRun->GetMCEventHeader();;
```
here?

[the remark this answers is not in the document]
ChristianTackeGSI: Okay, I admit, that I don't understand this fully.

* For me this looks like a memory leak currently. Could you add a [\todo](https://www.doxygen.nl/manual/commands.html#cmdtodo) comment in front a la?

  ```cpp
   /**
    * \todo Check whether this is a memory leak
    */
   ```

* As I understand it, the RunID should not be 0 in the output. Could you please add a simple check to the macro, so when we later try to refactor this, we'll have proper checking in place?

base/event/FairMCEventHeader.h
```diff
@@ -81,6 +81,23 @@ class FairMCEventHeader : public TNamed
...
+        clonedHeader->SetEventID(fEventId);
+        clonedHeader->SetVertex(fX, fY, fZ);
+        clonedHeader->SetTime(fT);
+        clonedHeader->SetB(fB);
+        clonedHeader->SetNPrim(fNPrim);
+        clonedHeader->MarkSet(fIsSet);
+        clonedHeader->SetRotX(fRotX);
+        clonedHeader->SetRotY(fRotY);
+        clonedHeader->SetRotZ(fRotZ);
+        return clonedHeader;
```
ChristianTackeGSI, line 98: I think, the default copy constructor should handle all of this?
```suggestion
        return new FairMCEventHeader(*this);
```

karabowi:
Redone in PR #1331 .

PR #1246, opened 2022-10-21
New online lib

fuhlig1:
Create a new library Online which contains the class FairRunOnline as well as any additional classes which are needed only when running online. This removes the dependency on ROOT::RHTTP from the Base library.
---

Checklist:

* [ x] Rebased against `dev` branch
* [ x] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)


CMakeLists.txt
```diff
@@ -106,9 +106,14 @@ ENDIF(NOT UNIX)
...
 option(BUILD_UNITTESTS "Build all unittests and add them as new tests" OFF)
 option(ENABLE_GEANT3_TESTING "Enable tests utilizing Geant3" OFF)
-option(BUILD_MBS "Build MBS" OFF)
 option(BUILD_PROOF_SUPPORT "Support ROOT::Proof (deprecated)" OFF)
 
+option(BUILD_ONLINE "Build the online library" ON)
+option(BUILD_MBS "Build MBS" OFF)
+if (BUILD_MBS)
+  set(BUILD_ONLINE ON)
+endif()
```
ChristianTackeGSI, line 115: I think, MBS (at least the part in the MbsAPI directory) can be built without ONLINE? So please try removing this?

CMakeLists.txt
```diff
@@ -249,9 +257,6 @@ else()
   message(STATUS "eventdisplay will not be built, because ROOT has no opengl support.")
 endif()
 
-if(BUILD_MBS)
-  add_subdirectory(MbsAPI)
-endif()
```
ChristianTackeGSI, line 254: Dennis suggested to have all "components" (exported targets) in sibling directories.
So please just keep MbsAPI in its current directory?

online/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
+ ################################################################################
+ #    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    #
```
ChristianTackeGSI, line 2:
```suggestion
 #    Copyright (C) 2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    #
```

online/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+ #                                                                              #
+ #              This software is distributed under the terms of the             #
+ #              GNU Lesser General Public Licence (LGPL) version 3,             #
+ #                  copied verbatim in the file "LICENSE"                       #
+ ################################################################################
+
+if(BUILD_MBS)
+  add_subdirectory(MbsAPI)
+endif()
+
```
ChristianTackeGSI, line 12:
```suggestion
```

If MbsAPI is a top level directory, then this is not needed here.

online/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+target_compile_definitions(${target} PRIVATE
+  $<$<BOOL:${BUILD_MBS}>:BUILD_MBS>
+  $<$<BOOL:${ROOT_gdml_FOUND}>:ROOT_HAS_GDML>
+  # $<$<PLATFORM_ID:Linux>:Linux>
+)
+
+fairroot_target_root_dictionary(${target}
+  HEADERS ${headers}
+  LINKDEF OnlineLinkDef.h
```
ChristianTackeGSI, line 65:
```suggestion
  LINKDEF LinkDef.h
```

As there is only one LinkDef in this directory, could we start naming it just `LinkDef.h`? I think, ROOT lately adopted this scheme as well.

online/OnlineLinkDef.h
```diff
@@ -0,0 +1,29 @@
+/********************************************************************************
+ *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
```
ChristianTackeGSI, line 2:
```suggestion
 *    Copyright (C) 2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
```

ChristianTackeGSI (changes requested):
I like it, that this seems to work! Good!

I added a few comments.

CMakeLists.txt
fuhlig1: You can compile MbsAPI without Online for sure but what is the sense of it? You need the MbsAPI only to read some data files which you can't without the Online library. 

So formally you can compile MbsAPI without Online but logically it doesn't make sense.

CMakeLists.txt
fuhlig1: Okay. Will do so.

online/CMakeLists.txt
fuhlig1: Thanks. I completely forgot about it. I will update also the one in base.

online/OnlineLinkDef.h
fuhlig1: Done.

online/CMakeLists.txt
fuhlig1: Should I also do the same in the base directory?

online/CMakeLists.txt
fuhlig1: Done.

CMakeLists.txt
ChristianTackeGSI: Could you look into [cmake_dependent_option](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html#command:cmake_dependent_option) and see whether you can use it to model this?

online/CMakeLists.txt
ChristianTackeGSI: If you like to? Yeah, sure.

CMakeLists.txt
fuhlig1: In my opinion this will not do what is intended. If I make BUILD_ONLINE depend on BUILD_MBS the option BUILD_ONLINE will be only available if BUILD_MBS is TRUE. In the other case BUILD_MBS is build always since the default value for BUILD_ONLINE is TRUE.

What I want to have is to avoid the case BUILD_MBS=TRUE && BUILD_ONLINE=FALSE since it doesn't make sense from the logical point of view. The code forces that case to BUILD_MBS=TRUE && BUILD_ONLINE=TRUE. All other cases are also valid.

ChristianTackeGSI: > avoid the case BUILD_MBS=TRUE && BUILD_ONLINE=FALSE

What about making MBS depend on ONLINE? That should exactly disallow that case?

fuhlig1: There a valid cases to build the Online library without MBS support. For example in CBM we need to run FairRunOnline with FairSources which reads our own file format, so we need FairRunOnline but not the MbsAPI.

I would like to get rid of the MbsAPI which we only copied from another project to allow R3B  to read their data. I am not aware of another experiment still using the old MBS data aquisition.

ChristianTackeGSI: > In the other case when BUILD_MBS depends on BUILD_ONLINE the MbsAPI is build always since the default value for BUILD_ONLINE is TRUE. This is something I don't want.

I read the documentation of that CMake feature a bit differently. I think, if MBS depends on ONLINE, then:

* If ONLINE is not enabled, MBS will not be available at all (see your requirement)
* If ONLINE is available, then MBS will be set to `<value>`, which can be `OFF`.

```cmake
cmake_dependent_option(BUILD_MBS "Build MBS" OFF "BUILD_ONLINE" OFF)
```

fuhlig1: This results in the following 

```
cmake  <sourcedir>
=>
BUILD_MBS:BOOL=OFF
BUILD_ONLINE:BOOL=ON

cmake  <sourcedir> --DBUILD_ONLINE=OFF
=> CMake crash since BUILD_MBS isn't known any longer 
CMake Error at examples/advanced/MbsTutorial/CMakeLists.txt:30 (target_link_libraries):
  Target "ExMbs" links to:

    FairRoot::Online

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```
The crash is understandable since it is checked if BUILD_MBS is TRUE to build some of the tests. Changing the check to

`if(BUILD_MBS AND BUILD_ONLINE)`

solves the issue.

I still don't like the proposed change. First off all I have to use an additional include file and more important I have to read the documentation to understand the new command and its logic. I such still prefer the implemented if statement. 

If you insist on the new function I will change the code.

ChristianTackeGSI: First of all: This was only a suggestion to make the dependencies more "coded" and have CMake do the work.

The crash above confuses me a bit, because BUILD_MBS should be OFF, if BUILD_ONLINE is OFF. Anyway.

As MBS will hopefully go away soon, go ahead with your original idea.

ChristianTackeGSI (changes requested):
Could you please add a simple `online/README.dox` a la `base/README.dox`, so that we get a nice new entry in the modules list in Doxygen (the CI runs Doxygen for you, so that you see the result!).

CMakeLists.txt
fuhlig1: The problem is that BUILD_MBS isn't available if BUILD_ONLINE is OFF. The option is only available if the other option is ON.

karabowi (approved):

fuhlig1:
Please don't merge it yet. I will update the PR.

dennisklein:
image

Here github hides a link with which you may convert your PR to a draft status which prevents merging. (I just pressed it)

image

Later you convert it back with this button "Ready for review"

fuhlig1:
> Could you please add a simple `online/README.dox` a la `base/README.dox`, so that we get a nice new entry in the modules list in Doxygen (the CI runs Doxygen for you, so that you see the result!).

Added with the latest commit. 
Also squashed everything to just one commit.

ChristianTackeGSI:
> > Could you please add a simple `online/README.dox` a la `base/README.dox`, so that we get a nice new entry in the modules list in Doxygen (the CI runs Doxygen for you, so that you see the result!).
> 
> Added with the latest commit. Also squashed everything to just one commit.

Things look very good!

I am currently unable to find the `README.dox`?

Oh, and could you add a short statement to the ChangeLog? Something about FairRunOnline and friends moved and people need to add FairRoot::Online as a new dependency target to their builds (probably under Breaking Changes)?

fuhlig1:
> > > Could you please add a simple `online/README.dox` a la `base/README.dox`, so that we get a nice new entry in the modules list in Doxygen (the CI runs Doxygen for you, so that you see the result!).
> > 
> > 
> > Added with the latest commit. Also squashed everything to just one commit.
> 
> Things look very good!
> 
> I am currently unable to find the `README.dox`?
> 
> Oh, and could you add a short statement to the ChangeLog? Something about FairRunOnline and friends moved and people need to add FairRoot::Online as a new dependency target to their builds (probably under Breaking Changes)?

Done.

ChristianTackeGSI (approved):

karabowi (dismissed):
@fuhlig1 , can you please squash into one commit?

dennisklein:
Squashed it, lgtm ;)

Issue #790, opened 2018-06-08
Can `FairRootManager::RegisterAny` consume `T*` instead of `T*&` as second argument?


dennisklein:
Followup on one aspect identified in #789.

Disadvantages of `T*&`:
* If people want to register object that are referenced by a smart pointer, they are currently told to extract the raw pointer and extend its lifetime with a local static variable.
* The calling method will effectively only work once.
* You cannot pass rvalues.

sawenzel:
I agree that there is room for improvement here and I appreciate the initiative. I remember it simply did not work with just `T*`. Some limitations come in my understanding from the way you need to pass addresses to the TBranch.

PR #1249, opened 2022-11-01
TCA's 2.0: preparation work

klenze:
These are changes in the FairTask and FairRootManager classes to support type safe (and optionally associative) TClonesArrays. 
FairRootManager
* I did not find a good way to fetch objects registered with RegisterAny, so I make GetMemoryBranchAny public.
* FairRootManager::RegisterAny() previously took the address of the pointer passed by reference (which happened to live on my local stack) and fed that to AddMemoryBranchAny(), so GetMemoryBranchAny() later returned an pointer to the stack. I am unsure why the double pointer would be required? In my mind, the process should be simple: someone creates some object and passes FRM a pointer to it for safekeeping. Later, someone else asks FRM for the object and get the pointer back. 

FairTask:
* While we can get away with fetching TCAs used as an input on first use, we must register TCAs used for output during Init, because subsequent (old school TCA using) tasks may depend on them being around during their Init phase. Having the user maintain their Init() manually is exactly the boilerplate code I want to avoid. Thus, additionally to Init(), there is a ``std::vector<std::function<void(void)>> fOnInit`` whose functions also get executed by InitTask. A public method ``AddOnInit std::function<void(void)>`` allows to add stuff to that vector. 
* During the constructor of a FairTask, member variables which have to handle TCA creation and registration are passed the this pointer of the FairTask. In their constructor, they register (via AddOnInit) a callback to a lambda expression which does the necessary steps during Init. 
* I will give an example in the related issue I opened. 
---

Checklist:

### DIFF
--- a/fairtools/MCConfigurator/FairYamlVMCConfig.h
+++ b/fairtools/MCConfigurator/FairYamlVMCConfig.h
@@ -16,6 +16,11 @@
 #include "FairGenericVMCConfig.h"
 
 #include <string>
+
+// prevent yaml-cpp/node/detail/iterator.h:48:54:
+//  error: no member named 'next' in namespace 'boost'
+#include <boost/next_prior.hpp>
+
 #include <yaml-cpp/yaml.h>
 
 class FairYamlVMCConfig : public FairGenericVMCConfig
--- a/base/steer/FairRootManager.h
+++ b/base/steer/FairRootManager.h
@@ -163,7 +163,7 @@ class FairRootManager : public TObject
 
     /** create a new branch based on an arbitrary type T (for which a dictionary must exist) **/
     template<typename T>
-    void RegisterAny(const char* name, T*& obj, Bool_t toFile);
+    void RegisterAny(const char* name, T* obj, Bool_t toFile);
 
     void RegisterInputObject(const char* name, TObject* obj);
 
@@ -273,6 +273,10 @@ class FairRootManager : public TObject
     /**Read one event from source to find out which RunId to use*/
     Bool_t SpecifyRunId();
 
+    template<typename T>
+    T GetMemoryBranchAny(const char* name) const;
+
+  
   private:
     // helper struct since std::pair has problems with type_info
     struct TypeAddressPair
@@ -300,9 +304,7 @@ class FairRootManager : public TObject
     void AddMemoryBranch(const char*, TObject*);
 
     template<typename T>
-    void AddMemoryBranchAny(const char* name, T** obj);
-    template<typename T>
-    T GetMemoryBranchAny(const char* name) const;
+    void AddMemoryBranchAny(const char* name, T* obj);
 
     template<typename T>
     void RegisterImpl(const char* name, const char* Foldername, T* obj, Bool_t toFile);
@@ -407,13 +409,18 @@ class FairRootManager : public TObject
 
 // FIXME: move to source since we can make it non-template dependent
 template<typename T>
-void FairRootManager::AddMemoryBranchAny(const char* brname, T** obj)
+void FairRootManager::AddMemoryBranchAny(const char* brname, T* obj)
 {
-    if (fAnyBranchMap.find(brname) == fAnyBranchMap.end()) {
+    if (fAnyBranchMap.find(brname) == fAnyBranchMap.end())
+      {
         auto& ot = typeid(T*);
         auto& pt = typeid(T);
         fAnyBranchMap[brname] = std::unique_ptr<TypeAddressPair const>(new TypeAddressPair(ot, pt, (void*)obj));
-    }
+      }
+    else
+      {
+        LOG(fatal) << "Failed to add branch: " << brname << ", it already exists!\n";
+      }
 }
 
 // try to retrieve an object address from the registered branches/names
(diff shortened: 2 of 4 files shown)


dennisklein:
> I am unsure why the double pointer would be required?

I have asked the same question in the past: https://github.com/FairRootGroup/FairRoot/issues/790. The problem seems to occur when reading the TCA branches back from file. Have you tried reading your data back? I either get `Error in <TTree::SetBranchAddress>: The address for "..." should be the address of a pointer!` or a segfault.

ChristianTackeGSI:
TBH I haven't looked in detail at this yet.

But could you please split this into two or three topicals Pull Requests?

For example the yaml / boost / next thing vaguely reminds me of something that we have already seen or even discussed. Would be helpful to have it in a dedicated PR to allow more structured discussions.

fairtools/MCConfigurator/FairYamlVMCConfig.h
```diff
@@ -16,6 +16,11 @@
 #include "FairGenericVMCConfig.h"
 
 #include <string>
+
+// prevent yaml-cpp/node/detail/iterator.h:48:54:
+//  error: no member named 'next' in namespace 'boost'
+#include <boost/next_prior.hpp>
```
dennisklein, line 22: Could it be that you are using yaml-cpp < 0.6? I guess, we only support 0.6+, which does not depend on Boost any more.

ChristianTackeGSI: This looks a lot like #1199?

dennisklein:
> While we can get away with fetching TCAs used as an input on first use, we must register TCAs used for output during Init, because subsequent (old school TCA using) tasks may depend on them being around during their Init phase. Having the user maintain their Init() manually is exactly the boilerplate code I want to avoid. Thus, additionally to Init(), there is a std::vector<std::function<void(void)>> fOnInit whose functions also get executed by InitTask. A public method AddOnInit std::function<void(void)> allows to add stuff to that vector.

While I understand what you are after, I am not convinced this approach is a good design. I have looked at your `roothacks::DContainer` which uses this interface: Why would you want to have your container type depend on the task?

Have you considered specializing `FairTask` as `R3BTask` which offers this registry? Or in other words, what would be the general use case why we should carry this on FairRoot level?

klenze:
> While I understand what you are after, I am not convinced this approach is a good design. I have looked at your roothacks::DContainer which uses this interface: Why would you want to have your container type depend on the task?

The constructor template is because we have a separate class called FairReader (which does not inherit from R3BReader) which also has an Init method (and now a AddOnInit). Using C++20 concepts, one could probably specify "any class which has an AddOnInit method". Or one could use an interface. 

> Have you considered specializing FairTask as R3BTask which offers this registry? Or in other words, what would be the general use case why we should carry this on FairRoot level?

I don't think this code has much use beyond encapsulating TCAs. The question is if you want to do that in FairRoot in general. 

Historically, the sentiment that it would be nice to have such functionality came up [before](https://github.com/R3BRootGroup/R3BRoot/issues/436#issuecomment-883163995):
> Now, all of that should have been done by FairRoot since the beginning - we really should not have to worry about all these Nullpointer checks, endless casting and type safety, and should leverage all the stuff provided by the stl.

Thinking a bit more, I still see some way to simplify things: The creation of managed TCAs should not depend on the presence of other TCAs. (no "only register this TCA if another TCA was fetched by FRM::GetObject()"). Thus, it would probably be okay to create and register all of the managed TCAs in any order, provided it is immediately before the FairTasks are Init()ed. 
(This would mean the the second argument of the DContainer could simply be bool doCreate=0)

I can see several ways to contain the AddOnInit method:
* A singleton FairTask which manages to register itself so that it is always executed first
* A list in an existing singleton (FairRootManager would be an obvious candidate). 

Would any of them be acceptable to you?

dennisklein:
> > While I understand what you are after, I am not convinced this approach is a good design. I have looked at your roothacks::DContainer which uses this interface: Why would you want to have your container type depend on the task?
>
> The constructor template is because we have a separate class called FairReader (which does not inherit from R3BReader) which also has an Init method (and now a AddOnInit). Using C++20 concepts, one could probably specify "any class which has an AddOnInit method". Or one could use an interface.

I think I have actually not understood your use case. I'll leave it to the colleagues/community and retract my opinion on this aspect of your PR.

klenze:
Good news. By creating a pair of  singleton FairTasks, and placing then first and last(ish) in FairRun::fTask, I can make sure that my output TCAs are registered before any other tasks Init and my input TCAs are initialized after them (assuming the FairTasks using them are instantiated after any FairTasks which create them are added). 

With regard to registering non-TObjects, I now use a simple global ``std::map<std::string, boost::any>`` instead, so I don't have to figure out if the correct way to use RegisterAny (and if there is a public way to retrieve objects). 

I still consider storing the address of a value passed by reference in RegisterAny in the line ``AddMemoryBranchAny<T>(brname, &obj);`` to be evil -- if you require a persistent object pointer, don't ask for a reference. Ideally, ask for a shared_ptr, or at least a a pointer. This will force the user to to at least consider that a stack object is not what you want. 

So, don't do (which you did for ObjT=T* in ``RegisterAny(const char* brname, T*& obj, bool persistence)``)
```
void Foo(ObjT& t)
  {
      this->fFoo=&t;
  }
```
which will possibly fail if someone uses 
```
  {
     ObjT temp;
     Someptr->Foo(temp);
   }
   // Someptr->fFoo now dangling
```
but instead
```
void Foo(ObjT* t)
  {
      this->fFoo=t;
  }
```
This would mean that the caller would have to run 
```
     ObjT temp;
     Someptr->Foo(&temp);
```
which rightly looks much more suspicious.
Just my 2 cents.

Issue #1228, opened 2022-09-28
Add build switch(es) to disable feature/example depending on Boost


dennisklein:
The idea would be to have `BUILD_<component>` switches. Naturally, one switch per source tree directory and maybe a few more for the more complex modules like Base and for `examples/` directory. The granularity of those switches should allow to effectively disable any dependency to Boost.

dennisklein:
#1251 scheduled for release in v18.8.

PR #1251, opened 2022-11-04
Support disabling `BaseMQ` and dependencies

dennisklein:
solves #1228

#### Motivation

AliceO2 does not use the `FairRoot::BaseMQ` target. Because it cannot be disabled, FairRoot pulls in quite heavy dependencies like Boost and FairMQ, which are not used through FairRoot. @ktf has requested a possibility to disable those libraries in FairRoot.

One approach is certainly to exactly define a build switch based on the request, such as `DISABLE_BOOST=ON` or similar. However, longterm maintenance of such a dependency-based build switch may grow complicated as FairRoot is a collection of dozens of libraries which may each have their own rationale when and why they want to depend on a Boost library.

Therefore, I propose to add a simple build switch `BUILD_BASEMQ=ON/OFF` which just disables the `FairRoot::BaseMQ` library and all that depends on it (`FairRoot::FairMQ`, `FairRoot::ParMQ` and some examples). Technically, `FairRoot::Base` also depended on Boost via its Boost.Serialization support for the `FairHit`, `FairMCPoint`, and `FairTimeStamp` classes. I proposed in this PR to refactor this feature into its [non-intrusive variant](https://www.boost.org/doc/libs/1_80_0/libs/serialization/doc/tutorial.html#nonintrusiveversion) and make it optional from the point of view of the `FairRoot::Base` library. When `BUILD_BASEMQ=ON` is configured, then the Boost lookup will require the Boost.Serialization library as well, effectively controlling this feature in the `FairRoot::Base` library (which is the exact use case it was introduced for in the first place).

In the last commit I change the lookup logic of the FairMQ and Boost dependencies based on build options only. Together with this change, a user like Alice may now configure FairRoot with `-DBUILD_BASEMQ=OFF -DBUILD_EXAMPLES=OFF` and effectively disable all targets that depend on either FairMQ or Boost (in future releases an existing or new FairRoot library may - of course - still opt-in to have a new dependency on FairMQ or Boost, then a user shall be ideally presented with a suitable `BUILD_<lib>` switch as well, if it makes sense).

The general motivation to provide `BUILD_<lib>` switches and then in turn `REQUIRE` dependencies is to enable reproducible builds that do not depend on the environment. The build shall fail fast (at configure time), if the given environment is not suitable for the chosen configuration. If we would chose to adopt this logic, adding a complete set of `BUILD_<lib>` switches and transforming all bits in the FairRoot build system to this scheme consistently, is out of scope of this PR - however, the general adoption of such a change should be discussed before merging this PR - otherwise it would perhaps add just more inconsistencies as there are already present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Some headers were cleaned up and now `#include` fewer
   other headers. You might have to add some `#includes`s
   in your code.
+* Added a new build switch `BUILD_BASEMQ` for controlling whether `FairRoot::BaseMQ` and
+  dependent targets are built. It is enabled by default and will now **require**
+  its external package dependencies FairMQ and Boost.
 
 
 ### Deprecations
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,6 +113,7 @@ option(BUILD_MBS "Build MBS" OFF)
 if (BUILD_MBS)
   set(BUILD_ONLINE ON)
 endif()
+option(BUILD_BASEMQ "Build the 'FairRoot::BaseMQ' library" ON)
 
 # searches for needed packages
 # REQUIRED means that cmake will stop if this packages are not found
@@ -263,13 +264,11 @@ add_subdirectory(datamatch)
 
 add_subdirectory(templates)
 
-If (FairMQ_FOUND AND Boost_FOUND)
+if(BUILD_BASEMQ)
   add_subdirectory(fairmq)
   add_subdirectory(basemq)
   add_subdirectory(parmq)
-Else ()
-  Message(STATUS "fairmq, basemq and parmq will not be built, because FairMQ and/or Boost was not found.")
-EndIf ()
+endif()
 
 Option(BUILD_DOXYGEN "Build Doxygen" OFF)
 if(BUILD_DOXYGEN)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,33 +170,44 @@ if(NOT DEFINED GTEST_ROOT)
   set(GTEST_ROOT ${SIMPATH})
 endif()
 
-find_package2(PUBLIC FairMQ VERSION 1.4.0)
+if(BUILD_BASEMQ)
+  find_package2(PUBLIC FairMQ VERSION 1.4.0 REQUIRED)
+endif()
+
 find_package2(PUBLIC DDS)
 find_package2(PUBLIC FairLogger VERSION 1.2.0 REQUIRED)
 
-if(NOT DEFINED Boost_NO_SYSTEM_PATHS)
-  Set(Boost_NO_SYSTEM_PATHS TRUE)
-endif()
-if(Boost_NO_BOOST_CMAKE)
-  # If an older version of boost is found both of the variables below are
-  # cached and in a second cmake run, a good boost version is found even
-  # if the version is to old.
-  # To overcome this problem both variables are cleared before checking
-  # for boost.
-  Unset(Boost_INCLUDE_DIR CACHE)
-  Unset(Boost_LIBRARY_DIRS CACHE)
-endif()
+if(BUILD_BASEMQ OR BUILD_UNITTESTS OR BUILD_EXAMPLES)
+  if(NOT DEFINED Boost_NO_SYSTEM_PATHS)
+    Set(Boost_NO_SYSTEM_PATHS TRUE)
+  endif()
+  if(Boost_NO_BOOST_CMAKE)
+    # If an older version of boost is found both of the variables below are
+    # cached and in a second cmake run, a good boost version is found even
+    # if the version is to old.
+    # To overcome this problem both variables are cleared before checking
+    # for boost.
+    Unset(Boost_INCLUDE_DIR CACHE)
+    Unset(Boost_LIBRARY_DIRS CACHE)
+  endif()
 
-list(APPEND boost_dependencies filesystem serialization)
+  set(boost_dependencies)
+  if(BUILD_BASEMQ)
+    list(APPEND boost_dependencies filesystem serialization program_options)
+  endif()
+  if(BUILD_UNITTESTS)
+    list(APPEND boost_dependencies unit_test_framework)
+  endif()
+  if(BUILD_EXAMPLES)
+    list(APPEND boost_dependencies program_options)
+  endif()
 
-if(BUILD_UNITTESTS)
- list(APPEND boost_dependencies unit_test_framework)
+  find_package2(PUBLIC Boost
+    VERSION 1.67
+    COMPONENTS ${boost_dependencies}
+    REQUIRED
+  )
 endif()
-
-find_package2(PUBLIC Boost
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -268,10 +268,10 @@ add_subdirectory(templates)
 
 If (FairMQ_FOUND AND Boost_FOUND)
   add_subdirectory(fairmq)
-  add_subdirectory(base/MQ)
+  add_subdirectory(basemq)
   add_subdirectory(parmq)
 Else ()
-  Message(STATUS "base/MQ and parmq will not be built, because FairMQ and/or Boost was not found.")
+  Message(STATUS "basemq and parmq will not be built, because FairMQ and/or Boost was not found.")
```
ChristianTackeGSI: Please mention FairMQ as well.

ChristianTackeGSI (commented):
You know, I love smaller PRs…

What about moving the first two commits into its own PR? I would approve it immediately.

Also we could use that PR to discuss about starting to move all of the new top level directories (which I like!) into a new dedicated `src/` directory? (So fairmq, basemq, and parmq would be the first entries in the new `src/`.)

All of that is purely "internal" and would not affect any user of FairRoot.

CMakeLists.txt
```diff
@@ -169,37 +170,48 @@ if(NOT DEFINED GTEST_ROOT)
...
+  endif()
+  if(Boost_NO_BOOST_CMAKE)
+    # If an older version of boost is found both of the variables below are
+    # cached and in a second cmake run, a good boost version is found even
+    # if the version is to old.
+    # To overcome this problem both variables are cleared before checking
+    # for boost.
+    Unset(Boost_INCLUDE_DIR CACHE)
+    Unset(Boost_LIBRARY_DIRS CACHE)
+  endif()
```
ChristianTackeGSI, line 195: Could we move this boost package finding setup into some `cmake/private/` thing like fairroot_prepare_boost_find_packege()`?

CHANGELOG.md
```diff
@@ -68,6 +68,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Some headers were cleaned up and now `#include` fewer
   other headers. You might have to add some `#includes`s
   in your code.
+* Added a new build switch `BUILD_BASEMQ` for controlling whether `FairRoot::BaseMQ` and
+  dependent targets are built. It is enabled by default and will now **require**
+  its external package dependencies FairMQ and Boost.
```
ChristianTackeGSI, line 73: Can you please spell out the dependent targets?

ChristianTackeGSI (commented):
Some small other notes.

About the rest:
* I have no clue on this serialization business! I just wonder, wether one could move the new "external" serialization support into its own header, and if that would be useful?
* You know, I like explicit and reproducible stuff.

CMakeLists.txt
dennisklein: We should rather clean it up. E.g.

```
  if(NOT DEFINED Boost_NO_BOOST_CMAKE AND CMAKE_VERSION VERSION_LESS 3.15)
    Set(Boost_NO_BOOST_CMAKE TRUE)
  endif()
```
can go away, since `dev` requires CMake 3.15+

```
  if(NOT DEFINED Boost_NO_SYSTEM_PATHS)
    Set(Boost_NO_SYSTEM_PATHS TRUE)
  endif()
```
is anyways very debatable

```
  if(Boost_NO_BOOST_CMAKE)
    # If an older version of boost is found both of the variables below are
    # cached and in a second cmake run, a good boost version is found even
    # if the version is to old.
    # To overcome this problem both variables are cleared before checking
    # for boost.
    Unset(Boost_INCLUDE_DIR CACHE)
    Unset(Boost_LIBRARY_DIRS CACHE)
  endif()
```
we should check, if this is really needed still with the supported Boost versions - it also sounds as it came around the 1.69 to 1.70 releases (+ the current CMake versions at that time, pre 3.15) when Boost started to ship its own CMake package.

CHANGELOG.md
dennisklein: Sure, however, let's first discuss whether we want to adopt the general `BUILD_<lib>` option scheme. Depending on the details this text will change anyways. Because, if we add build switches for - let's say `BUILD_PARMQ` - what will happen, if the user configures `-DBUILD_BASEMQ=OFF -DBUILD_PARMQ=ON`. In FairMQ I have a helper [`fairmq_build_option`](https://github.com/FairRootGroup/FairMQ/blob/master/cmake/FairMQBuildOption.cmake). But it is rather stupid, it is not recursive and will not switch dependencies on or off automatically (but complain, if a dependency is off e.g.). I remember you discussed such details heavily with @fuhlig1 in #1246, but I have not read and followed it in detail.

dennisklein:
> What about moving the first two commits into its own PR? I would approve it immediately.

Done, see #1252. Will rebase here, once the other one is merged.

CMakeLists.txt
dennisklein: done, see #1252

dennisklein:
> I just wonder, wether one could move the new "external" serialization support into its own header, and if that would be useful?

Yes, we could, but would you want to break the user interface - should the user explicitely include this header? Or purely as a organizational, internal measure to separate the code?

CMakeLists.txt
dennisklein: partly addressed in 1cadcbac4672d684861f87bc680dac2879c75fbb in #1252

ChristianTackeGSI:
> Yes, we could, but would you want to break the user interface - should the user explicitely include this header? Or purely as a organizational, internal measure to separate the code?

First: I don't know, really!

*I* would consider breaking the user experience and letting them include the extra "boost serialization header". Maybe even making the new headers part of BaseMQ? But then, really, I have not taken any look at actual use cases and the consequences.

In reality, I don't understand, what this feature targets? Why exactly only those three classes?

This was mostly an academic question. So we can declare this part as "resolved" and postpone to another PR.

dennisklein:
> In reality, I don't understand, what this feature targets? Why exactly only those three classes?

It was introduced in commit 1fd40f8e2be78ba068c15496a680cbafed986840, maybe this gives more context?

base/event/FairTimeStamp.h
```diff
@@ -139,6 +141,8 @@ void save(Archive& ar, FairTimeStamp const& time, const unsigned int)
 
 }   // namespace boost::serialization
 
+BOOST_SERIALIZATION_SPLIT_FREE(FairTimeStamp)
+
 #endif   // FAIRROOT_HAS_BOOST_SERIALIZATION
```
dennisklein, line 147: Reminder to self: This is the wrong line!

karabowi:
I would not like to have our examples depend on boost.
We can easily remove all the serialization stuff from advanced/propagator example - it is not used anyhow.
As for simulation/Tutorial1 - is there some easy to use replacement for boost::options?

dennisklein:
> As for `simulation/Tutorial1` - is there some easy to use replacement for `boost::options`?

https://github.com/CLIUtils/CLI11 I used in the past.

karabowi (approved):
The removal of examples dependency on boost will follow.

ChristianTackeGSI (approved):

PR #1254, opened 2022-11-09
Make member pointer fEvtHeader of FairRun into unique pointer

### LINKED ISSUE #1242: Ownership transferring needs to be explicitly expressed
As C++ Core guidelines (I.11) suggests:
> Never transfer ownership by a raw pointer (T*) or reference

## Problem:
For example, the event header, which is denoted as `fEvtHeader`, is owned by class `FairRun` as it's deleted in the deconstructor.

But at the same time, there is a method, which can reseat the fEvtHeader pointer to another user-defined object:
https://github.com/FairRootGroup/FairRoot/blob/8501262650150c547e4aa6dbaff204e476ab6e41/base/steer/FairRun.h#L79

Here we have two problems: 
1. If fEvtheader is already pointing to an existing event header, that existing one won't get deleted in the end and hence we have a memory leak.
2. Users have absolutely no idea whether they need to delete the pointer they passed to the method or not. If they don't go deep into the source files, the default is to delete the pointer because it's passed as a raw pointer.

## Suggestions:
1. Whenever a class owns an object, encapsulate the pointer into `std::unique_ptr<>`. ( this also increases exception safety. )
2. When a class need to consume an object, make the parameter as the type `std::unique_ptr<>`
    As in the case of the example, following code is much better:
    ```cpp
    void SetEventHeader(std::unique_ptr<FairEventHeader> NewEvHeader) 
    { 
        fEvtHeader = std::move(NewEvHeader); // reseating the unique_ptr
    }
    ```

    With this, the two problem mentioned above are solved. First, old event header get deleted automatically. Second, User knows this is an ownership transferring and he doesn't need to delete the pointer created by him.

    For the user, all he needs to do is:
    ```cpp
    auto eventHeader = std::unique_ptr<R3BEventHeader>(new R3BEventHeader());
    FairRun::Instance()->SetEventHeader(std::move(eventHeader));    // ownership transferring.
    ```

YanzhaoW:
Encapsulate member pointer FairEventHeader from FairRun inside `std::unique_ptr`. This specifies the ownership of the event header and prevents it from being deleted multiple times.

Fixes #1242

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
     in addition.
 * Renamed our `ROOT_GENERATE_DICTIONARY` to `FAIRROOT_GENERATE_DICTIONARY`.
   (It's not used in many places anyway, it seems.)
+* `fEvtHeader` member variable now is a private unique pointer owned by `FairRun`. To access
+  the event header, please use the public member function `GetEventHeader()`.
 * The following files have been deleted. As far as we know they were not used anywhere:
   * basemq/baseMQtools/baseMQtools.h
   * basemq/policies/Sampler/FairMQFileSource.h
@@ -78,6 +80,9 @@ file an issue, so that we can see how to handle this.
 * Deprecated FairEventBuilder and FairEventBuilderManager
   * The functionality, introduced to enable event reconstruction, is not used.
   * It can be enabled with `-DBUILD_EVENT_BUILDER=ON`.
+* Deprecated `FairRun::SetEventHeader(FairEventHeader*)`
+   * Use `FairRun::SetEventHeader(std::unique_ptr<FairEventHeader> EvHeader)`, which 
+     indicates the ownership transferring.
 
 ### Other Notable Changes
 * Consider calling `fairroot_check_root_cxxstd_compatibility()`
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -32,6 +32,7 @@ Shahoyan, Ruben
 Stockmanns, Tobias
 Ustyuzhanin, Andrey
 von Haller, Barthélémy
+Wang, Yanzhao [https://orcid.org/0000-0002-7006-7986]
 Wenzel, Sandro
 Wielanek, Daniel
 Winckler, Nicolas
--- a/fairroot/base/steer/FairRun.h
+++ b/fairroot/base/steer/FairRun.h
@@ -75,7 +75,9 @@ class FairRun : public TNamed
      *       Set the experiment dependent run header
      *       for each run
      */
-    void SetEventHeader(FairEventHeader* EvHeader) { fEvtHeader = EvHeader; }
+    void SetEventHeader(std::unique_ptr<FairEventHeader> EvHeader);
+    [[deprecated("Use SetEventHeader(std::unique_ptr<FairEventHeader>) instead")]] void SetEventHeader(
+        FairEventHeader* EvHeader);
     /**
      * return a pointer to the RuntimeDB
      */
@@ -206,6 +208,8 @@ class FairRun : public TNamed
     FairRun& operator=(const FairRun&) { return *this; }
     /** Number of Tasks added*/
     Int_t fNTasks;
+    /** MC Event Header */
+    std::unique_ptr<FairEventHeader> fEvtHeader;   //!
 
     FairLinkManager fLinkManager{};   //!
 
@@ -228,8 +232,6 @@ class FairRun : public TNamed
     UInt_t fRunId;   //!
     /** true for Anaylsis session*/
     Bool_t fAna;   //!
-    /** MC Event Header */
-    FairEventHeader* fEvtHeader;   //!
     /** File  Header */
     FairFileHeader* fFileHeader;
     /** true if RunInfo file should be written*/
--- a/fairroot/base/steer/FairRunAna.cxx
+++ b/fairroot/base/steer/FairRunAna.cxx
@@ -202,20 +202,16 @@ void FairRunAna::Init()
 
     // Assure that basic info is there for the run
     //  if(par && fInputFile) {
+    auto* evtHeader = GetEventHeader();
     if (par && fInFileIsOpen) {
 
         LOG(info) << "Parameter and input file are available, Assure that basic info is there for the run!";
         fRootManager->SpecifyRunId();
-
-        //    fEvtHeader = GetEventHeader();
-        GetEventHeader();
-
         FillEventHeader();
-
         fRunId = GetEvtHeaderRunId();
 
         // Copy the Event Header Info to Output
-        fEvtHeader->Register(GetSink() ? fStoreEventHeader : false);
+        evtHeader->Register(GetSink() ? fStoreEventHeader : false);
 
         // Init the containers in Tasks
         LOG(info) << "--- Initialize with RunId  --- " << fRunId;
@@ -225,12 +221,11 @@ void FairRunAna::Init()
         fTask->SetParTask();
     } else {   // end----- if(fMixedInput)
         LOG(info) << "Initializing without input file or Mixed input";
-        FairEventHeader* evt = GetEventHeader();
-        evt->Register(GetSink() ? fStoreEventHeader : false);
+        evtHeader->Register(GetSink() ? fStoreEventHeader : false);
         FairRunIdGenerator genid;
         fRunId = genid.generateId();
         fRtdb->addRun(fRunId);
-        evt->SetRunId(fRunId);
+        evtHeader->SetRunId(fRunId);
         fTask->SetParTask();
         if (!fRtdb->initContainers(fRunId)) {
             LOG(error) << "FairRunAna::Init: fRtdb->initContainers failed";
--- a/fairroot/base/steer/FairRunAnaProof.cxx
+++ b/fairroot/base/steer/FairRunAnaProof.cxx
@@ -13,7 +13,6 @@
 #include "FairRunAnaProof.h"
 
 #include "FairBaseParSet.h"
-#include "FairEventHeader.h"
 #include "FairFieldFactory.h"
 #include "FairFileHeader.h"
 #include "FairLogger.h"
@@ -170,19 +169,18 @@ void FairRunAnaProof::Init()
 
     // Assure that basic info is there for the run
 
+    auto* evtHeader = GetEventHeader();
     if (par && fInFileIsOpen) {
 
         LOG(info) << "Parameter and input file are available, Assure that basic info is there for the run!";
         fRootManager->SpecifyRunId();
-
-        GetEventHeader();
 
         FillEventHeader();
 
         fRunId = GetEvtHeaderRunId();
 
         // Copy the Event Header Info to Output
-        fEvtHeader->Register(GetSink() ? fStoreEventHeader : false);
+        evtHeader->Register(GetSink() ? fStoreEventHeader : false);
 
         // Init the containers in Tasks
         LOG(info) << "--- Initialize with RunId  --- " << fRunId;
@@ -192,12 +190,11 @@ void FairRunAnaProof::Init()
         fTask->SetParTask();
     } else {   // end----- if(fMixedInput)
         LOG(info) << "Initializing without input file or Mixed input";
-        FairEventHeader* evt = GetEventHeader();
-        evt->Register(GetSink() ? fStoreEventHeader : false);
+        evtHeader->Register(GetSink() ? fStoreEventHeader : false);
         FairRunIdGenerator genid;
         fRunId = genid.generateId();
         fRtdb->addRun(fRunId);
-        evt->SetRunId(fRunId);
+        evtHeader->SetRunId(fRunId);
         fTask->SetParTask();
         if (!fRtdb->initContainers(fRunId)) {
             LOG(error) << "FairRunAnaProof::Init: fRtdb->initContainers failed";
@@ -238,19 +235,18 @@ void FairRunAnaProof::InitContainers()
     fRtdb = GetRuntimeDb();
     FairBaseParSet* par = dynamic_cast<FairBaseParSet*>(fRtdb->getContainer("FairBaseParSet"));
 
+    auto* evtHeader = GetEventHeader();
     if (par && fInFileIsOpen) {
         fRootManager->SpecifyRunId();
-
-        GetEventHeader();
 
         FillEventHeader();
 
         fRunId = GetEvtHeaderRunId();
 
-        LOG(info) << "RKGOT FEH " << fEvtHeader << " (" << fRunId << ") FROM FRM " << fRootManager << ".";
+        LOG(info) << "RKGOT FEH " << evtHeader << " (" << fRunId << ") FROM FRM " << fRootManager << ".";
 
         // Copy the Event Header Info to Output
-        fEvtHeader->Register(GetSink() ? fStoreEventHeader : false);
+        evtHeader->Register(GetSink() ? fStoreEventHeader : false);
 
         // Init the containers in Tasks
         LOG(info) << "--- Initialize with RunId  --- " << fRunId;
--- a/fairroot/online/steer/FairRunOnline.cxx
+++ b/fairroot/online/steer/FairRunOnline.cxx
@@ -130,10 +130,9 @@ void FairRunOnline::Init()
         fRootManager->ReadEvent(0);
     }
 
-    GetEventHeader();
-
+    // --- Get event header from Run
+    auto* evtHeader = GetEventHeader();
     FillEventHeader();
-
     if (0 == fRunId)   // Run ID was not set in run manager
     {
         if (0 == GetEvtHeaderRunId())   // Run ID was not set in source
@@ -187,14 +186,9 @@ void FairRunOnline::Init()
         LOG(error) << "FairRunOnline::Init: fRtdb->initContainers failed";
     }
 
-    // --- Get event header from Run
-    if (!fEvtHeader) {
-        LOG(fatal) << "FairRunOnline::Init() No event header in run!";
-        return;
-    }
-    LOG(info) << "FairRunOnline::Init() Event header at " << fEvtHeader;
-    fRootManager->Register("EventHeader.", "Event", fEvtHeader, (nullptr != GetSink()));
-    fEvtHeader->SetRunId(fRunId);
+    LOG(info) << "FairRunOnline::Init() Event header at " << evtHeader;
+    fRootManager->Register("EventHeader.", "Event", evtHeader, (nullptr != GetSink()));
+    evtHeader->SetRunId(fRunId);
 
     if (!GetSource()->InitUnpackers()) {
         LOG(fatal) << "FairRunOnline->Init() InitUnpackers() failed!";
@@ -222,6 +216,7 @@ Int_t FairRunOnline::EventLoop()
 
     fRootManager->StoreWriteoutBufferData(fRootManager->GetEventTime());
     fTask->ExecuteTask("");
+    // TODO: Fill the event header again? Is this necessary?
     FillEventHeader();
     Fill();
     fRootManager->DeleteOldWriteoutBufferData();
--- a/fairroot/base/steer/FairRun.cxx
+++ b/fairroot/base/steer/FairRun.cxx
@@ -37,13 +37,13 @@ FairRun* FairRun::Instance() { return fRunInstance; }
 FairRun::FairRun(Bool_t isMaster)
     : TNamed()
     , fNTasks(0)
+    , fEvtHeader()
     , fRtdb(FairRuntimeDb::instance())
     , fTask(new FairTask("FairTaskList"))
     , fRootManager(0)
     , fUserOutputFileName()
     , fRunId(0)
     , fAna(kFALSE)
-    , fEvtHeader(nullptr)
     , fFileHeader(new FairFileHeader())
     , fGenerateRunInfo(kFALSE)
     , fIsMaster(isMaster)
@@ -96,7 +96,6 @@ FairRun::~FairRun()
         // who is responsible for the RuntimeDataBase?
         delete fRtdb;
     }
-    delete fEvtHeader;
     if (fRunInstance == this) {
         // Do not point to a destructed object!
         fRunInstance = nullptr;
@@ -146,12 +145,22 @@ FairTask* FairRun::GetTask(const char* taskName)
     return dynamic_cast<FairTask*>(task);
 }
 
+void FairRun::SetEventHeader(FairEventHeader* EvHeader)
+{
+    fEvtHeader.reset(EvHeader);
+}
+
+void FairRun::SetEventHeader(std::unique_ptr<FairEventHeader> EvHeader)
+{
+    fEvtHeader = std::move(EvHeader);
+}
+
 FairEventHeader* FairRun::GetEventHeader()
 {
-    if (nullptr == fEvtHeader) {
-        fEvtHeader = new FairEventHeader();
+    if (!fEvtHeader) {
+        fEvtHeader = std::make_unique<FairEventHeader>();
     }
-    return fEvtHeader;
+    return fEvtHeader.get();
 }
 
 void FairRun::SetUseFairLinks(Bool_t val) { fRootManager->SetUseFairLinks(val); }
@@ -237,7 +246,7 @@ void FairRun::FillEventHeader()
         return;
     }
 
-    fSource->FillEventHeader(fEvtHeader);
+    fSource->FillEventHeader(fEvtHeader.get());
 }
 
 void FairRun::AddAlignmentMatrices(const std::map<std::string, TGeoHMatrix>& alignmentMatrices, bool invertMatrices)


YanzhaoW:
For a short remark, I would suggest there should be no more `delete` operator in the FairRoot. For the new operator, except for **some** ROOT build-in classes, instead use 
```cpp
std::make_unique<typename> ptr;
// or
std::make_shared<typename> ptr;
```

base/steer/FairRunAnaProof.h
```diff
@@ -82,6 +82,7 @@ class FairRunAnaProof : public FairRunAna
     FairRunAnaProof operator=(const FairRunAnaProof&);
 
     FairFileSource* fProofFileSource;
+    FairEventHeader* fEvtHeader = nullptr;
```
ChristianTackeGSI, line 85: Which code in FairRunAnaProof initalizes it to the real pointer?

ChristianTackeGSI (commented):
Many thanks for your PR!
I like most of it!

I added a few comments here and there.

And you might possibly want to wait for #1255 getting merged and then rebase on it?

base/steer/FairRun.h
```diff
@@ -205,6 +207,8 @@ class FairRun : public TNamed
     FairRun& operator=(const FairRun&) { return *this; }
     /** Number of Tasks added*/
     Int_t fNTasks;
+    /** MC Event Header */
+    std::unique_ptr<FairEventHeader> fEvtHeader;   //!
```
ChristianTackeGSI, line 211: I really appreciate, that you're making this private.

But maybe let's do things step by step and keep it protected for the time being and fix up all those users that can't handle a `unique_ptr`? The major one is probably FillEventHeader(). I had #1255 in my private tree anyway and thought it's now a good time for it. It should make this much simpler. So maybe you might want to wait for it getting merged and then rebase?

YanzhaoW:
@ChristianTackeGSI Thanks for your comments. Sure, let's pend this PR until #1255 is merged.

karabowi:
@YanzhaoW , can you squash your commits into one, and then add one extra commit with your name in the CONTRIBUTORS file?

YanzhaoW:
@karabowi Sorry. Yes, sure. But I still need some time and to rebase it on c4ced68 commit. After I finish, I will squash it to one commit.

base/steer/FairRunAnaProof.h
YanzhaoW: Initialisation of event header is in `FairRun`. And pointer `fEvtHeader` in `FairRunAnaProof` is not onwed.

base/steer/FairRun.h
YanzhaoW: I would rather like to keep it private. If I make it protected (, which I tried in the very first beginning), lots of places would need to be revised because of of the naming. If other derived class need to use this private pointer, just call the function `GetEventHeader()`

CONTRIBUTORS
```diff
@@ -32,6 +32,7 @@ Shahoyan, Ruben
 Stockmanns, Tobias
 Ustyuzhanin, Andrey
 von Haller, Barthélémy
+Wang, Yanzhao
```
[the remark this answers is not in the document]
YanzhaoW: no, I don't have this. ;D

base/steer/FairRunAnaProof.h
YanzhaoW: Oh, it seems I misunderstood the "initialisation" of the pointer. If it means assigning a valid address to the pointer, ah, yes, only FairRunAnaProof can initialise it:
 https://github.com/FairRootGroup/FairRoot/blob/95b6b43de7ff33308989473c2cf90132effc3fa8/base/steer/FairRunAnaProof.cxx#L243

That's the only place when it's assigned. And in the `Init()` function the pointer cannot be guaranteed to have a valid memory address. I will make a change about it very soon. (also in other classes).

base/steer/FairRun.h
YanzhaoW: How about changing it to 
```cpp
fEvtHeader = GetEventHeader();
```

YanzhaoW:
Ok, I made additional changes to prevent calling the member functions of `FairEventHeader` when it's nullptr.

I'm not sure which of them are really necessary and which are not. The event header is used in all derived classes of FairRun and some of them are doing exactly the same thing. I would suggest better put those steps in the member function of `FairRun` and call it when necessary. This could make the code much more maintainable.

base/steer/FairRunAna.cxx
```diff
@@ -503,6 +507,8 @@ void FairRunAna::Run(Long64_t entry)
 {
     UInt_t tmpId = 0;
     fRootManager->ReadEvent(entry);
+    if(!fEvtHeader)
+        fEvtHeader = GetEventHeader();
     tmpId = fEvtHeader->GetRunId();
```
ChristianTackeGSI, line 514: I don't really like that we have to check each instance of `->GetRunId()` for fEvtHeader being set…

What do you think about [this](https://github.com/ChristianTackeGSI/FairRoot/commit/fa238b90f8a27edeae66c34056ef900ffb50cf34) (you might want to cherry-pick it into your branch?)

CONTRIBUTORS
ChristianTackeGSI, line 35: BTW, if you have an ORCiD, append it like this
`Family, Give [https://orcid.org/0000-XXXX-XXXX-XXXX]`.

base/steer/FairRun.h
ChristianTackeGSI: Thinking again about it…
```cpp
fEvtHeader = dynamic_cast<FairEventHeader*>(fRootManager->GetObject("EventHeader."));
```

I think, this assumed (in the current version), that `GetObject` returns a fresh/new object. Because currently the destructor of `FairRun` deletes it.

So what about replacing this just with `SetEventHeader(…)`? Then the unique_ptr will do the `delete` (again).

base/steer/FairRunAna.h
```diff
@@ -100,6 +100,7 @@ class FairRunAna : public FairRun
   private:
     FairRunAna(const FairRunAna& M);
     FairRunAna& operator=(const FairRunAna&) { return *this; }
+    FairEventHeader* fEvtHeader = nullptr;
```
ChristianTackeGSI, line 103: Having another thought about this as well:

Could you try to not use member variables, but use local variables in the methods that need `fEvtHeader`?

Like
```cpp
FairRunAna::DoFoo() {
    auto fEvtHeader = GetEventHeader();
    …
    // code that actually needs fEvtHeader
```

that way the relevant methods havr access to the current value and not just an old one?

ChristianTackeGSI (commented):
Hi, excuse me for being late…

By now, we had another round of fEvtHeader usage cleanup: f7091f7d86e8737f28913830713606da488a51cb

So could you try a rebase, please?

CONTRIBUTORS
YanzhaoW: Ok, thanks for the info.

base/steer/FairRun.h
YanzhaoW: Yeah, I would also do this. Event header should be directly obtained from `FairRun` every time for the consistency.

base/steer/FairRunAna.h
YanzhaoW: Yes, this is a better way.

YanzhaoW:
Hi, @ChristianTackeGSI

Here are some changes from the latest commit (82c9a19):
1. From CHANGELOG file, you can see I deprecate two member functions in FairRun to make them more readable and safer.
2. Remove private member of FairEventHeader from FairRunAna, FairRunOnline and FairRunAnaProof.
3. Delete all lines which obtain the eventHeader from `FairRunManager::GetObject()`. Which makes the eventHeader of the whole program consistent with the one from FairRun::GetEventHeader().

Please check the commit and see whether the changes break anything not wanted.

base/steer/FairRunAnaProof.cxx
```diff
@@ -237,20 +237,16 @@ void FairRunAnaProof::InitContainers()
...
     FairBaseParSet* par = dynamic_cast<FairBaseParSet*>(fRtdb->getContainer("FairBaseParSet"));
 
+    auto* evtHeader = GetEventHeader();
     if (par && fInFileIsOpen) {
         fRootManager->ReadEvent(0);
 
-        fEvtHeader = dynamic_cast<FairEventHeader*>(fRootManager->GetObject("EventHeader."));
-
-        if (nullptr == fEvtHeader)
-            LOG(fatal) << "Could not read event header.";
```
ChristianTackeGSI, line 246: Also relevant to this.

online/steer/FairRunOnline.cxx
```diff
@@ -217,13 +210,13 @@ void FairRunOnline::InitContainers()
     if (nullptr == par)
         LOG(warn) << "FairRunOnline::InitContainers: no  'FairBaseParSet' container !";
 
+    // --- Get event header from Run
+    auto evtHeader = GetEventHeader();
     if (par) {
-        fEvtHeader = static_cast<FairEventHeader*>(fRootManager->GetObject("EventHeader."));
```
ChristianTackeGSI, line 221: Also relevant to this.

ChristianTackeGSI:
Hi @YanzhaoW!

We concluded this morning, that GetObject does not return an owning pointer (as already stated in #1346). @karabowi will take a look at FairRun and friends, how GetObject is used for the EventHeader.

Many thanks for your patience!

YanzhaoW:
Hi, @ChristianTackeGSI. Are there any updates from @karabowi about this PR?

ChristianTackeGSI:
Hi!

Thanks for asking!

We're still stuck in the first point of #1346 (looking at the semantics of GetObject in FairRunAnaProof).
I hope @karabowi can take a look at it next week maybe. Let's see, what he says…

karabowi:
Experiment tests are being performed now to check if we can change the `FairRunOnline` and remove `GetObject`.

fairroot/base/steer/FairRunAna.cxx
```diff
@@ -200,20 +203,18 @@ void FairRunAna::Init()
...
-
-        FillEventHeader();
+        FillEventHeader(evtHeader);
 
-        fRunId = GetEvtHeaderRunId();
+        fRunId = GetEvtHeaderRunId(evtHeader);
 
         // Copy the Event Header Info to Output
-        fEvtHeader->Register(GetSink() ? fStoreEventHeader : false);
+        evtHeader->Register(fStoreEventHeader);
```
ChristianTackeGSI, line 217: We added the `GetSink` checks on purpose. Can you please keep them?

ChristianTackeGSI (commented):
Please take a look at the compile errors from the CI.

Please consider squashing your commits together.

I added a few more comments inline.

fairroot/base/steer/FairRunAna.cxx
```diff
@@ -432,7 +433,7 @@ void FairRunAna::RunEventReco(Int_t Ev_start, Int_t Ev_end)
         fRootManager->StoreWriteoutBufferData(fRootManager->GetEventTime());
         fTask->ExecuteTask("");
 
-        FillEventHeader();
+        FillEventHeader(GetEventHeader());
```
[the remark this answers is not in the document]
YanzhaoW: Hi. By doing so, my idea was that the function FillEventHeader takes the event header as a parameter and therefore is independent of the event header intrinsically. It's a very subtle change. IMO, it's more readable and less error-prone. Otherwise, one possibility to make an error is to change the event header inside the `FillEventHeader()`.

But now I think we should get just rid of it. Instead we could add a getter for the fSource and directly fill the event header from the source. I think it's more clearer in this way.

fairroot/base/steer/FairRunAna.cxx
```diff
@@ -482,7 +483,7 @@ void FairRunAna::RunMQ(Long64_t entry)
    it read a certain event and call the task exec, but no output is written
    */
     fRootManager->ReadEvent(entry);
-    auto const tmpId = GetEvtHeaderRunId();
+    auto const tmpId = GetEventHeader()->GetRunId();
```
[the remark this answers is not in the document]
YanzhaoW: This is the same reason as above.

fairroot/base/steer/FairRunAnaProof.cxx
```diff
@@ -167,19 +169,21 @@ void FairRunAnaProof::Init()
...
-        FillEventHeader();
+        fRunId = GetEvtHeaderRunId(evtHeader);
 
-        fRunId = GetEvtHeaderRunId();
+        // Copy the Event Header Info to Output
+        evtHeader->Register(kTRUE);
 
         // Copy the Event Header Info to Output
-        fEvtHeader->Register(GetSink() ? fStoreEventHeader : false);
+        evtHeader->Register();
```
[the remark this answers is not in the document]
YanzhaoW: Yeah, thanks.

fairroot/online/steer/FairRunOnline.h
```diff
@@ -8,6 +8,13 @@
 #ifndef FAIRRUNONLINE_H
 #define FAIRRUNONLINE_H
 
+/**
+ * Configure and manage the Analysis
+ * @author M. Al-Turany D. Bertini
+ * @version 0.1
+ * @since 28.02.05
+ */
+
```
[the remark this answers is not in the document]
YanzhaoW: Sorry, I made a mistake when doing the rebasing.

YanzhaoW:
Hi, @ChristianTackeGSI

 Thanks for your comments. Sorry for the delay. I made few corrections and will squash the commits once everything is considered ok. Please check my replies and ask me if you have more questions or suggestions.

fairroot/base/steer/FairRunAna.cxx
YanzhaoW: Sure

YanzhaoW:
The compilation is successful. But I got a segmentation error from ctest "ex_tutorial3_reco_timebased_TGeant4". Other ctests work fine.

fairroot/base/steer/FairRunAna.cxx
```diff
@@ -531,7 +532,7 @@ void FairRunAna::RunTSBuffers()
             fRootManager->ReadNonTimeBasedEventFromBranches(globalEvent++);
         }
         fTask->ExecuteTask("");
-        FillEventHeader();
+        GetSource()->FillEventHeader(GetEventHeader());
```
ChristianTackeGSI, line 535: I think, this one is causing the final segmentation fault, because GetSource might return a `nullptr`. The old `FillEventHeader()` catches this (withut giving a warning…)

ChristianTackeGSI (commented):
Just a short quick note… I will look into this more on Friday (hopefully).

fairroot/base/steer/FairRun.h
```diff
@@ -247,16 +249,19 @@ class FairRun : public TNamed
     /**
      * Call FillEventHeader on the source
      */
-    void FillEventHeader()
+    [[deprecated("Use GetSource()->FillEventHeader()")]] void FillEventHeader()
```
dennisklein, line 252: If we recommend to unconditionally dereference the return value of `GetSource()`, we should also make sure it never returns a `nullptr`, right?

The C++ idiomatic way to communicate this would be to return a reference instead of a raw pointer, but FairRoot rarely uses references ..

Perhaps we should add new *safer-to-use* APIs to support call chains?
```cpp
FairSource& GetDataSource();
FairSource const& GetDataSource() const;
```

---

`GetEventHeader()`'s implementation currently guarantees not to return a `nullptr`, but I would still add a comment to the header file documenting this intended post condition that it does not get violated in the future by accident. Additionally, we should add unit tests.

fairroot/base/steer/FairRunAna.cxx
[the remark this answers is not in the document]
ChristianTackeGSI: I learned about the Facade pattern from @dennisklein. And yes, I think, `FillEventHeader()` and `GetEvtHeaderRunId()` are facades.

And we're not trying to hide anything away. First, these are internal member functions. Not meant for public use (at least not yet). Second, the full source is available anyway.

I tried to improve documentation with #1413.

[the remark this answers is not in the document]
YanzhaoW: From my understanding, the Facade design pattern involves with a Facade object that provides a simpler public interface to a complicated underlying system. But here the function `FillEventHeader()` just provide an additional way to modify its private member variable.

Protected functions are still not private, which means any derived class could expose the functionality to the public.

However the question remains: there are two paths/ways to do one thing. This could further cause additional problems:

1. Confusing to users. 
   Which way should I choose? 
  ```cpp
  auto* evtHeader = run->GetEventHeader();
  auto* source = run->GetSource();
  source->FillEventHeader(evtHeader);
  ```
  or
  ```cpp
  run->FillEventHeader();
  ```
2. Bug-prone and difficult to maintain. 
    Supposing that in the future there is a change for the function `FillEventHeader()`, the code using the second way could have this new implementation. But other code using the first method still has the old implementation even though these two method are meant to do exactly the same thing.

fairroot/base/steer/FairRun.h
YanzhaoW: Yes, the common way for a getter function is to return a const ref. Returning a raw pointer requires the caller to constantly check whether it's a nullptr.

We could implement additional getters to return references. But for `GetSource()`, is it possible to do one of these?
1. Throw an exception `std::runtime_error()` if source is nullptr
2. Use return gsl::not_null<FairSource*>. This will call std::terminate automatically when fSource is a nullptr. By this way we don't need to change the API.

dennisklein: I would love to fix some of the return types of existing member functions, but how can one do that gracefully in C++ (with a deprecation period)?

YanzhaoW: Assuming we are not going to break API, `gsl::not_null` is pretty good shot for those member functions returning a pointer. 

For reference, I think we need two getters returning const ref and ref:
```cpp
auto Get_cref() -> const Type&;
auto Get_ref() -> Type&;
``` 
I wouldn't say the following is a good choice:
```cpp
auto Get() const -> const Type&;
auto Get() -> Type&;
```
because in many cases when the object is not a const, we still need to get a const ref to its member variable for the reading. Some classes like FairRun can never be a const. But get a const ref of, for example, event header for its reading is still necessary.

ChristianTackeGSI:
@YanzhaoW wow, good catch!

Could you please open an issue for this?

Some of the main developers should take a look at this, really.

I still wonder, whether running without a Source is a valid scenario at all. (Running without a Sink has lately been declared a valid scenario.)

fairroot/base/steer/FairRunAna.cxx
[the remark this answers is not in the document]
ChristianTackeGSI: > Protected functions are still not private, which means any derived class could expose the functionality to the public.

Those functions have been declared (via documentation) as internal in #1413. If someone outside FairRoot uses them, they're entirely on their own. If this is actually a concern, we could mark them as `private:` and use `friend` to allow all the relevant users inside FairRoot. But TBH I don't like that either: At some point, I like it, if things are private even inside FairRoot.

Point (1) then isn't relevant: Only possibly confusing to FairRoot developers. My take on that one: Use the shorter API, so that the caller is easier to read (which is really an important point in our current codebase).

Point (2) is also only relevant to FairRoot developers. But that conern is a problematic point of the facade pattern at all. So yes, FairRoot devs have to take care.

YanzhaoW:
Hi, @ChristianTackeGSI 

I have changed them back to `FillEventHeader()` and also rebased them to the head of the dev branch.

For this PR, I only made eventHeader to be a private unique_ptr of FairRun. Other issues, such as filling the event header many times and its mysterious ownership from FairRootManager (see #1418 ) , is not addressed.

If everything is alright, I will squash my commits for the merging.

fairroot/base/steer/FairRun.h
```diff
@@ -258,7 +260,10 @@ class FairRun : public TNamed
      *
      * \todo Should `fEvtHeader` be checked?
      */
-    UInt_t GetEvtHeaderRunId() const { return fEvtHeader->GetRunId(); }
+    [[deprecated("Use GetEventHeader()->GetRunId()")]] UInt_t GetEvtHeaderRunId() const
+    {
+        return fEvtHeader->GetRunId();
+    }
```
ChristianTackeGSI, line 266: Can we please also keep `GetEvtHeaderRunId()` for the same reason we keep `FillEventHeader()`?

fairroot/base/steer/FairRunAna.cxx
```diff
@@ -225,12 +221,11 @@ void FairRunAna::Init()
         fTask->SetParTask();
     } else {   // end----- if(fMixedInput)
         LOG(info) << "Initializing without input file or Mixed input";
-        FairEventHeader* evt = GetEventHeader();
-        evt->Register(GetSink() ? fStoreEventHeader : false);
+        evtHeader->Register(fStoreEventHeader);
```
ChristianTackeGSI, line 224:
```suggestion
        evtHeader->Register(GetSink() ? fStoreEventHeader : false);
```

fairroot/base/steer/FairRunAnaProof.cxx
```diff
@@ -192,12 +190,11 @@ void FairRunAnaProof::Init()
         fTask->SetParTask();
     } else {   // end----- if(fMixedInput)
         LOG(info) << "Initializing without input file or Mixed input";
-        FairEventHeader* evt = GetEventHeader();
-        evt->Register(GetSink() ? fStoreEventHeader : false);
+        evtHeader->Register();
```
ChristianTackeGSI, line 193:
```suggestion
        evtHeader->Register(GetSink() ? fStoreEventHeader : false);
```

fairroot/base/steer/FairRunAnaProof.cxx
```diff
@@ -238,19 +235,18 @@ void FairRunAnaProof::InitContainers()
     fRtdb = GetRuntimeDb();
     FairBaseParSet* par = dynamic_cast<FairBaseParSet*>(fRtdb->getContainer("FairBaseParSet"));
 
+    auto* evtHeader = GetEventHeader();
     if (par && fInFileIsOpen) {
-        fRootManager->SpecifyRunId();
-
-        GetEventHeader();
+        fRootManager->ReadEvent(0);
```
ChristianTackeGSI, line 240: I am not quite sure that `SpecifyRunId` and `ReadEvent(0)` do the same thing? I am confused.

fairroot/base/steer/FairRunAnaProof.cxx
```diff
@@ -238,19 +235,18 @@ void FairRunAnaProof::InitContainers()
...
 
-        fRunId = GetEvtHeaderRunId();
+        fRunId = evtHeader->GetRunId();
 
-        LOG(info) << "RKGOT FEH " << fEvtHeader << " (" << fRunId << ") FROM FRM " << fRootManager << ".";
+        LOG(info) << "RKGOT FEH " << evtHeader << " (" << fRunId << ") FROM FRM " << fRootManager << ".";
 
         // Copy the Event Header Info to Output
-        fEvtHeader->Register(GetSink() ? fStoreEventHeader : false);
+        evtHeader->Register();
```
ChristianTackeGSI, line 249:
```suggestion
        evtHeader->Register(GetSink() ? fStoreEventHeader : false);
```

fairroot/online/steer/FairRunOnline.cxx
```diff
@@ -130,21 +130,20 @@ void FairRunOnline::Init()
         fRootManager->ReadEvent(0);
     }
 
-    GetEventHeader();
-
-    FillEventHeader();
+    GetSource()->FillEventHeader(GetEventHeader());
 
+    auto* evtHeader = GetEventHeader();
```
ChristianTackeGSI, line 135: Maybe that's much simpler?

```suggestion
    auto evtHeader = GetEventHeader();
    FillEventHeader();
```

fairroot/online/steer/FairRunOnline.cxx
```diff
@@ -188,13 +187,13 @@ void FairRunOnline::Init()
     }
 
     // --- Get event header from Run
-    if (!fEvtHeader) {
+    if (evtHeader == nullptr) {
         LOG(fatal) << "FairRunOnline::Init() No event header in run!";
         return;
     }
```
ChristianTackeGSI, line 193:
```suggestion
```

I wonder, whether this is needed at all? `GetEventHEader()` ensures that a header is returned always, right?

ChristianTackeGSI (commented):
Hi!

Many thanks for keeping track on this, really!

I appreciate your work and that you keep with this stuff.

My comments are really meant to improve things. And if you think, that something is too much work, please let me know, then I'll try to take over that part.

fairroot/base/steer/FairRun.h
YanzhaoW: Hmmm, I still don't understand this style. 

`FairEventHeader` has other getters, like `GetInputFiledID`, `GetEventTime`, etc. Does that mean we should also define `GetEvtHeaderInputFieldID`, `GetEvtEventTime`, etc in `FairRun`? If so, we should do the same for `fSource`. And in the end, `FairRun` could be filled with more than hundreds of getters and setters.

YanzhaoW:
@ChristianTackeGSI 
Thanks for your comments. Most of issues you found are probably due to some confusions when I resolve merge conflicts. I will fix them as soon as possible.

fairroot/online/steer/FairRunOnline.cxx
YanzhaoW: Yes, you are right.

fairroot/base/steer/FairRun.h
ChristianTackeGSI: These are only intended for internal use in FairRun{,Ana,Sim} and not for public use. And so we only have "helpers" for exactly the few interactions with FairEventHeader, that happen often enough in FairRun\*, that it is worth to have an internal helper that simplifies internal use. And yes, for FairSource or FairSink, we could have those internal helpers, if they simplify things.

At least that was my idea when I introduced some of these.

YanzhaoW: Sorry, I can't agree that those internal helper functions simplify things. There is nothing to simplify if there are just one or two lines of code. And on the contrary, it makes things more unreadable and less transparent. For example, from the name of `FillEventHeader()`, it's difficult to say to where this event header is filled and why it's related to a file source. Same goes for `GetEvtHeaderRunId()`. It means just one line of code `evtHeader->GetRunId()` and makes the code less transparent.

I think we should try to write a clean code. And a clean code includes no unnecessary interface or internal method. Sorry about wasting the time for this argument. Maybe a third person opinion could help us address this issue more quickly? @dennisklein

dennisklein: Unfortunately, I don't have a good argument to add, which would favor one over the other :see_no_evil:

However, I would like to repeat the comment [I made earlier](https://github.com/FairRootGroup/FairRoot/pull/1254#discussion_r1237288780). If such a recommendation like `GetEventHeader()->GetRunId()` would be added, let's make some effort it is not going to be dangerous code. (I know there is already a good share of unsafe APIs and not-so-safe coding practices in FairRoot/ROOT, but if we never start to address them, the situation will never improve, even though it is only tiny steps)

YanzhaoW: I agree ;D

YanzhaoW:
Hi, @ChristianTackeGSI 
 I have squashed all commits into one. If there is nothing else, it's ready to be merged.

CHANGELOG.md
```diff
@@ -83,6 +83,11 @@ file an issue, so that we can see how to handle this.
 * Consider calling `fairroot_check_root_cxxstd_compatibility()`
   in your `CMakeLists.txt`.
 * `fairsoft-config` isn't searched for and not needed any more.
+* `fEvtHeader` member variable now is a private unique pointer owned by `FairRun`. To access
+  the event header, please use the public member function `GetEventHeader()`.
+* `FairRun::SetEventHeader(FairEventHeader*)`, 
+use `FairRun::SetEventHeader(std::unique_ptr<FairEventHeader> EvHeader)`, which 
+indicates the ownership transferring.
```
ChristianTackeGSI, line 90: 1. Some really small suggestions:

```suggestion
* `fEvtHeader` member variable now is a private unique pointer owned by `FairRun`. To access
  the event header, please use the public member function `GetEventHeader()`.
* `FairRun::SetEventHeader(FairEventHeader*)` got deprecated, 
  use `FairRun::SetEventHeader(std::unique_ptr<FairEventHeader> EvHeader)`, which 
  indicates the ownership transferring.
```

2. Please reoder a bit:
   * The change of fEvtHeader is actually breaking (because derived classes can't any longer acces it). Could you move that bullet point under the "Breaking Changes" subsection?
   * And the other one is a deprecation. We also have a subsection for deprecationbs. Cizld you move it there?

ChristianTackeGSI (commented):
I like this! We can merge this really, really soon.

I have only a small stylistic request (really tidying up things) for the CHANGELOG.

CHANGELOG.md
YanzhaoW: Sure, no problem.

YanzhaoW: Fixed.

ChristianTackeGSI (approved):

ChristianTackeGSI:
@YanzhaoW many thanks for this! This looks so good!

I have created #1423 which depends on your stuff. Maybe you could leave a Review there?

PR #1264, opened 2022-11-13
Fix warnings

dennisklein:
I verified 3e31ea4b4b698715865e47e108885443e75fa615 manually with Tutorial1, the GUI element didn't show any difference and worked as before, so I assume, there was no hidden side effect.

### DIFF
--- a/base/steer/FairTSBufferFunctional.h
+++ b/base/steer/FairTSBufferFunctional.h
@@ -1,5 +1,5 @@
 /********************************************************************************
- *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ * Copyright (C) 2014-2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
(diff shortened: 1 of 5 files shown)


base/steer/FairTSBufferFunctional.h
```diff
@@ -32,9 +32,13 @@ class TTree;
...
  */
 
-class BinaryFunctor : public std::binary_function<FairTimeStamp*, double, bool>
+class BinaryFunctor
 {
   public:
+    using first_argument_type = FairTimeStamp*;
+    using second_argument_type = double;
+    using result_type = bool;
+
```
ChristianTackeGSI, line 41: I wonder, whether those are actually used? Can we mark them deprecated?

ChristianTackeGSI (approved):
LGTM

base/steer/FairTSBufferFunctional.h
dennisklein: It looks like they are not used, but before deprecating this on such a fine-grained level, I would rather just deprecate/remove the whole `BinaryFunctor` type and replace it with `std::function<bool(FairTimeStamp*,double)>`.

ChristianTackeGSI: Okay. Let's postpone this for another time.

PR #1265, opened 2022-11-13
Attempt at fixing the `ConstructASCIIGeometry` helper

### LINKED ISSUE #1145: FairModule::ConstructASCIIGeometry: "warning: parameter 'dataType1' set but not used"
```console
base/sim/FairModule.h:170:44: warning: parameter 'dataType1' set but not used [-Wunused-but-set-parameter]
 void FairModule::ConstructASCIIGeometry(T* dataType1, TString containerName, U*)
                                         ~~~^~~~~~~~~
```

* FairRoot dev branch
* Added `-Wunused-but-set-parameter` in `cmake/modules/CheckCompiler.cmake`
* FairSoft's `test/env/fairroot_develop`
* Debian-10

dennisklein:
Proposal to fix #1145

~~Had to use different function name, because the existing helper's default arguments would have overlapped with my proposed "overload".~~

edit: The original approach grew into hundreds of lines of code changes, so I aborted it (due to lack of test coverage) and went now with the minimal fix I can think of.

### DIFF
--- a/base/sim/FairModule.h
+++ b/base/sim/FairModule.h
@@ -1,5 +1,5 @@
 /********************************************************************************
- *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ * Copyright (C) 2014-2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -90,9 +90,14 @@ class FairModule : public TNamed
     /** Finish worker run (used in MT mode only) */
     virtual void FinishWorkerRun() const { ; }
 
-    /**template function to construct geometry. to be used in derived classes.*/
+    /** @deprecated template function to construct geometry. to be used in derived classes.
+     * The first and the third argument are meaningless, just pass nullptr */
     template<class T, class U>
-    void ConstructASCIIGeometry(T* dataType1, TString containerName = "", U* datatype2 = nullptr);
+    [[deprecated("Broken signature, use ConstructASCIIGeometry(TString) instead")]] void
+        ConstructASCIIGeometry(T*, TString containerName = "", U* = nullptr);
+    /** Helper function to construct geometry. */
+    template<class T, class U>
+    void ConstructASCIIGeometry(TString containerName = "");
 
     /**Set the sensitivity flag for volumes, called from ConstructASCIIRootGeometry(), and has to be implimented for
      * detectors which use ConstructASCIIRootGeometry() to build the geometry */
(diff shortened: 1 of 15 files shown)


base/sim/FairModule.h
```diff
@@ -168,51 +179,59 @@ class FairModule : public TNamed
...
-    TList* volList = MGeo->getListOfVolumes();
-    // store geo parameter
-    FairRun* fRun = FairRun::Instance();
-    FairRuntimeDb* rtdb = FairRun::Instance()->GetRuntimeDb();
+template<typename GeoModule, typename ParameterSet>
+void FairModule::AddGeoNodesToParameterSetHelper(GeoModule& geoModule, std::string const& parameterSetName)
+{
+    auto const run = FairRun::Instance();
+    auto const rtdb = run->GetRuntimeDb();
+    auto parameterSet = static_cast<ParameterSet*>(rtdb->getContainer(parameterSetName.c_str()));
```
dennisklein, line 201:
```suggestion
    static_assert(std::is_base_of_v<FairParSet, ParameterSet>, "...");
    auto parameterSet = dynamic_cast<ParameterSet*>(rtdb->getContainer(parameterSetName.c_str()));
```

base/sim/FairModule.h
```diff
@@ -168,51 +179,59 @@ class FairModule : public TNamed
...
     }
+    return geoModule;
+}
 
-    TList* volList = MGeo->getListOfVolumes();
-    // store geo parameter
-    FairRun* fRun = FairRun::Instance();
-    FairRuntimeDb* rtdb = FairRun::Instance()->GetRuntimeDb();
+template<typename GeoModule, typename ParameterSet>
+void FairModule::AddGeoNodesToParameterSetHelper(GeoModule& geoModule, std::string const& parameterSetName)
```
dennisklein, line 197:
```suggestion
template<typename ParameterSet, typename GeoModule>
void FairModule::AddGeoNodesToParameterSetHelper(std::string const& parameterSetName, GeoModule& geoModule)
```

dennisklein: Ideally, it should also be `GeoModule const& geoModule` parameter, but this requires `getListOfVolumes()` to be `const` as well

base/sim/FairModule.h
```diff
@@ -90,9 +90,14 @@ class FairModule : public TNamed
     /** Finish worker run (used in MT mode only) */
     virtual void FinishWorkerRun() const { ; }
 
-    /**template function to construct geometry. to be used in derived classes.*/
+    /**template function to construct geometry. to be used in derived classes.
+     * The first and the third argument are meaningless, just pass nullptr */
```
ChristianTackeGSI, line 94: Can you add a `\deprecated` so it shows up in Doxygen's deprecation list?

ChristianTackeGSI (approved):

karabowi (dismissed):

base/sim/FairModule.h
dennisklein: done

ChristianTackeGSI (approved):

karabowi (approved):

PR #1267, opened 2022-11-14
Simplify boolean expr

dennisklein:
for the fun of it

### DIFF
--- a/generators/FairYPtGenerator.cxx
+++ b/generators/FairYPtGenerator.cxx
@@ -46,7 +46,7 @@ void FairYPtGenerator::SetYPt(const TH2D &yPt)
 
 Bool_t FairYPtGenerator::Init()
 {
-    if (FairBaseMCGenerator::Init() == kFALSE)
+    if (!FairBaseMCGenerator::Init())
         return kFALSE;
     fYPt.ResetStats();
     for (int i = 1; i <= fYPt.GetNbinsX(); i++) {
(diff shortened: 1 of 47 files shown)


ChristianTackeGSI (approved):
Please add `readability-simplify-boolean-expr` to `.clang-tidy`, so that we will not miss anything again.

generators/FairYPtGenerator.cxx
```diff
@@ -46,8 +46,8 @@ void FairYPtGenerator::SetYPt(const TH2D &yPt)
 
 Bool_t FairYPtGenerator::Init()
 {
-    if (FairBaseMCGenerator::Init() == kFALSE)
-        return kFALSE;
+    if (!FairBaseMCGenerator::Init())
+        return fFALSE;
```
ChristianTackeGSI, line 50: Typo…

I wonder, why this compiles on some of our machines.

```suggestion
        return false;
```

dennisklein: > I wonder, why this compiles on some of our machines.

I suspect overlapping success signals from the previous CI builds.

dennisklein: Ya, that's it, https://alfa-ci.gsi.de/blue/organizations/jenkins/FairRootGroup%2FFairRoot/detail/PR-1267/3/pipeline/ here you see that all really failed.

ChristianTackeGSI (approved):

PR #1269, opened 2022-11-14
Forward Port New Target Names

ChristianTackeGSI:
Forward Ported #1253 to `dev`

---

Checklist:

### DIFF
--- a/datamatch/CMakeLists.txt
+++ b/datamatch/CMakeLists.txt
@@ -1,12 +1,12 @@
 ################################################################################
-# Copyright (C) 2014-2019 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  #
+# Copyright (C) 2014-2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  #
 #                                                                              #
 #              This software is distributed under the terms of the             #
 #              GNU Lesser General Public Licence (LGPL) version 3,             #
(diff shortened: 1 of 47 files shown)


datamatch/CMakeLists.txt
```diff
@@ -1,12 +1,12 @@
 ################################################################################
-# Copyright (C) 2014-2019 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  #
+# Copyright (C) 2014-2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  #
 #                                                                              #
```
fuhlig1, line 3: In my opinion we should also deprecate the library. 

If I am correct it is not used by any of the experiments. 
This was a copy of some Panda code long time ago to make it generally available also for other experiments. It is definitely not used by CBM and and I am sure Panda still uses there own implementation (@karabowi could you please confirm).

fuhlig1 (approved):
Since you only changed the target names and not the library names I am file with the PR.

Please consider deprecating DataMatch.

ChristianTackeGSI:
> Since you only changed the target names and not the library names I am file with the PR.

Note that I scheduled renaming the actual filesystem objects for 19.2. The idea is, that everybody should have moved to targets with 19.0.

And 18.8 will already have all the targets with the new names.

> Please consider deprecating DataMatch.

I opened #1270 to keep track of this for now.

PR #1273, opened 2022-11-15
Fix warnings

dennisklein:

PR #1277, opened 2022-11-16
`CONTRIBUTING.md`

dennisklein:
I took @ChristianTackeGSI's [comment](https://github.com/FairRootGroup/FairRoot/pull/1273#discussion_r1023669941) as a spark to finally make of concrete proposal on how to write such things down.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,108 @@
+# Contributing
+
+## Bug reports, feature request, questions etc
+
+Create a [Github Issue](https://github.com/FairRootGroup/FairRoot/issues/new/choose).
+
+## Contributing code
+
+Create a [Github Pull Request](https://github.com/FairRootGroup/FairRoot/compare):
+
+1. By default, develop against the [`dev`](https://github.com/FairRootGroup/FairRoot/tree/dev)
+   branch and base your PR against it.
+   * In rare cases (e.g. backports, some hotfixes) base against the appropriate
+     branch.
+2. Follow [the seven rules of a great Git commit message](https://cbea.ms/git-commit/#seven-rules)!
+   * Use a meaningful commit granularity, e.g. do not mix
+     unrelated changes in a single commit and vice verse squash related commits
+     into one.
+   * Expect that a reviewer will ask you for restructuring your commits! This
+     usually happens towards the end of the lifetime of a PR when it matured
+     enough for merging.
+   * Utilize [the commit message body to explain *what* and *why* and not *how*](https://cbea.ms/git-commit/#why-not-how).
+     Aim your message body content to be meaningful (concise, but complete
+     enough) to a reader in the future!
+   * Utilize [reference keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)
+     both in commit messages as well as in PR comments if applicable.
+3. If you are a first time contributor, add a separate commit in your PR which
+   adds your name to the [`CONTRIBUTORS`](CONTRIBUTORS) file.
+4. Follow our [Coding Guidelines](#coding-guidelines).
+5. Address failed CI checks.
+
+# Coding Guidelines
+
+These guidelines are horribly incomplete at this point in time, but one has to
+start somewhere ;) Exactly because these guidelines are relatively new, do not be
+alarmed if large parts of this repo do not follow them (yet). The idea is to
+apply them incrementally on new code and when existing code is modified. These
+guidelines will also evolve as time passes.
+
+How to reference a rule in an issue or PR discussion:
+```
+// reference to our guideline
+[G.2](https://github.com/FairRootGroup/FairRoot/blob/master/CONTRIBUTING.md#g.2-adopted-c++-core-guidelines)
+// reference to c++ core guideline with prefix CPPCG::
+[CPPCG::I.11](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#i11-never-transfer-ownership-by-a-raw-pointer-t-or-reference-t)
+```
+
+## G - General
+
+### G.1 Our CI runs [`clang-tidy`](https://clang.llvm.org/extra/clang-tidy/)
+
+**Config**: [`.clang-tidy`](.clang-tidy)
+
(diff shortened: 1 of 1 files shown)


ChristianTackeGSI (dismissed):
Very good idea!

CONTRIBUTING.md
```diff
@@ -0,0 +1,70 @@
...
+
+### I - Initialization
+
+1. Prefer [default member initialization](https://en.cppreference.com/w/cpp/language/data_members#Member_initialization)
+   with braces, e.g.
+   ```cpp
+   class FairDetector {
+     int fDetId{0};
+   };
+   ```
```
ChristianTackeGSI, line 70: Maybe?

> Sometimes a classic member initializer is also fine. The obvious case is the specific constructor wanting to set a non default value. But also `unique_ptr`s of incomplete types (for what ever reason) should use member initializers and not default initializers for the time being.

dennisklein: at least that is how I understood [your comment](https://github.com/FairRootGroup/FairRoot/pull/1273#discussion_r1023669941)

dennisklein:
But we should definitely go through the points as a group and decide together, if we even want to adopt what I wrote.

CONTRIBUTING.md
ChristianTackeGSI: Okay, I think, there are two points here:

1. Whether or not to recommend default member initializers at all, or not.
   * This one is by me. I like default member initializers, because they make constructors more concise, avoid member initializer order errors, and avoid repetitions for multiple constructors.
   * Not to mention C++ Core Guidelines [C.48](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers)
   * But then there are some cases, where a "classic" member initializer is better. For one, if not using the _default_ value. For another, older compilers are not completely happy about `unique_ptr`s with an incomplete type.
2. If one actually uses a default member initializer, then there is the question of style. *I* didn't care much about it. When starting to use them, I used `{…};` for objects and ` = …;` for simple types. @fuhlig1 suggested to use one style for everything. Since using ` = …;` for complex objects feels very broken, the `{…};` style was left over.

karabowi:
Great idea!

Small remark. In the Checklist point 2, I would expect 7 stars ;).

dennisklein:
I guess, we could have a dedicated section "Adopted CPPCoreGuidelines". What do you think?

CONTRIBUTING.md
rbx: To me `= ...;` is more readable. Do we have to enforce consistency in this case?
I would vote for making `= ...;` the default/preferred one. And only use the other where it is unavoidable for some reason.

As I understand `= ...;` requires you to make things more explicit sometimes. To quote an example from @mattkretz:
```c++
class A
{
  public:
  explicit A(int);
};

class B
{
  A a1 = 1; // error
  A a2{1}; // ok
};
```
, which can still be solved by writing this instead:
```c++
class A
{
  public:
  explicit A(int);
};

class B
{
  A a1 = A(1);
  A a2 = A(1);
};
```

mattkretz: Right, :+1: for using default member initializers (and then defaulting the default ctor :grin:).

But otherwise, don't restrict initialization style. All have their places. In C++ the most natural syntax for initialization, to me at least, appears to be `Type name = init;`. I've always found `Type name(init);` to be strange - and it leads to the most vexing parse, which is why we got `Type name{init};`. Note that the latter has different semantics, because initializer lists (i.e. braces) require non-narrowing conversions.

dennisklein: I just removed this guideline from this PR. Note: It was also contended by two other persons during today's group meeting. So, it seems unlikely to be adopted at this point in time.

ChristianTackeGSI:
One point from this morning's session:

We should change the PullRequest template to have a note / link so people know that they should comply with the contributors guidelines.

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+     enough for merging.
+   * Utilize [the commit message body to explain *what* and *why* and not *how*](https://cbea.ms/git-commit/#why-not-how).
+     Aim your message body content to be meaningful (concise, but complete
+     enough) to a reader in the future!
+   * Utilize [reference keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)
+     both in commit messages as well as in PR comments if applicable.
+3. If you are a first time contributor, add a separate commit in your PR which
+   adds your name to the [`CONTRIBUTORS`](CONTRIBUTORS) file.
+4. Follow our [Coding Guidelines](#coding-guidelines).
+5. Address failed CI checks.
```
dennisklein, line 30:
```suggestion
2. If you are a first time contributor, add a separate commit in your PR which
   adds your name to the [`CONTRIBUTORS`](CONTRIBUTORS) file.
3. Follow our [Coding Guidelines](#coding-guidelines).
4. Expect that a reviewer will ask you for restructuring your commits! This
   usually happens towards the end of the lifetime of a PR when it matured
   enough for merging.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+time.
+
+### G.2 Adopted C++ Core Guidelines
+
+This shall be an evolving list of explicitely adopted C++ Core Guidelines:
+
+* [CPPCG::I.11: Never transfer ownership by a raw pointer (`T*`) or reference (`T&`)](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#i11-never-transfer-ownership-by-a-raw-pointer-t-or-reference-t)
+   * If an owning raw pointer cannot be avoided for legacy reasons,
+     **you must add a comment documenting the ownership semantics!**
+
```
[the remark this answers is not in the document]
dennisklein: Could also fit as D.2

ChristianTackeGSI:
Could we merge this (at least into dev) sooner than later? That way we have something in the repo to assist discussions. Individual discussions could then be handled using individual PRs.

For example we had the idea to put the "Things to remember for a Release" in here. That could be a nice PR later on.

dennisklein:
Feel free to copy or derive any text from here you would like to have. No need to put me as committer or anything.

Issue #1270, opened 2022-11-15
Deprecate FairRoot::DataMatch?


ChristianTackeGSI:
@fuhlig1 suggests [here](https://github.com/FairRootGroup/FairRoot/pull/1269#pullrequestreview-1180568622) to deprecate FairRoot::DataMatch.

fuhlig1:
Here the answer of Tobias, so we can deprecate FairRoot::DataMatch

> Hallo Florian,
> 
> die Library DataMatch kann auslaufen. Wir benutzen es nicht mehr in PandaRoot (weder über FairRoot noch über PandaRoot), da es zu langsam war.
> 
> LG,
> 
> Tobias
>

ChristianTackeGSI:
@fuhlig1 would you mind writing an appropriate ChangeLog-Entry for 18.8?

For 19 (dev) we might go the MBS way and add a `BUILD_DATAMATCH` option that defaults to OFF?

dennisklein:
wait, too fast, still needs to be picked to v18.8_patches

PR #1283, opened 2022-11-17
Deprecate DataMatch

karabowi:
By default it will not be built.
Enable it by -DBUILD_DATAMATCH=ON

Addresses issue #1270.

---

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,9 @@ file an issue, so that we can see how to handle this.
 * Many items were already deprecated in prior versions.
   Marked them with proper C++14 deprecation warnings.
   Scheduled them for removal in v20.
+* Deprecated DataMatch library
+  * Contact the developers should you need it.
+  * Scheduled to remove without replacement in a future major release of FairRoot.
 
 ### Other Notable Changes
 * CMake [targets](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html)
--- a/datamatch/CMakeLists.txt
+++ b/datamatch/CMakeLists.txt
@@ -39,6 +39,10 @@ target_link_libraries(${target} PUBLIC
   ROOT::Core
 )
 
+target_compile_definitions(${target} PRIVATE
+  FAIR_SILENCE_DATAMATCH_DEPRECATION
+)
+
 fairroot_target_root_dictionary(${target}
   HEADERS ${headers}
   LINKDEF LinkDef.h
--- a/datamatch/FairMCDataCrawler.h
+++ b/datamatch/FairMCDataCrawler.h
@@ -15,6 +15,10 @@
 #ifndef FAIRMCDATACRAWLER_H_
 #define FAIRMCDATACRAWLER_H_
 
+#ifndef FAIR_SILENCE_DATAMATCH_DEPRECATION
+#warning "FairRoot::DataMatch is deprecated and will be removed in a future major release"
+#endif
+
 #include "FairMultiLinkedData.h"   // for FairMultiLinkedData
 
 #include <Rtypes.h>    // for Int_t, Bool_t, etc
--- a/datamatch/FairMCEntry.h
+++ b/datamatch/FairMCEntry.h
@@ -15,6 +15,10 @@
 #ifndef FAIRMCENTRY_H_
 #define FAIRMCENTRY_H_
 
+#ifndef FAIR_SILENCE_DATAMATCH_DEPRECATION
+#warning "FairRoot::DataMatch is deprecated and will be removed in a future major release"
+#endif
+
 #include "FairLink.h"              // for FairLink
 #include "FairMultiLinkedData.h"   // for FairMultiLinkedData
 
--- a/datamatch/FairMCList.h
+++ b/datamatch/FairMCList.h
@@ -15,6 +15,10 @@
 #ifndef FAIRMCLIST_H_
 #define FAIRMCLIST_H_
 
+#ifndef FAIR_SILENCE_DATAMATCH_DEPRECATION
+#warning "FairRoot::DataMatch is deprecated and will be removed in a future major release"
+#endif
+
 #include <Rtypes.h>    // for Int_t, FairMCList::Class, etc
 #include <TObject.h>   // for TObject
 #include <vector>      // for vector
--- a/datamatch/FairMCMatch.h
+++ b/datamatch/FairMCMatch.h
@@ -15,6 +15,10 @@
 #ifndef FAIRMCMATCH_H_
 #define FAIRMCMATCH_H_
 
+#ifndef FAIR_SILENCE_DATAMATCH_DEPRECATION
+#warning "FairRoot::DataMatch is deprecated and will be removed in a future major release"
+#endif
+
 #include "FairMCEntry.h"           // for FairMCEntry
 #include "FairMCResult.h"          // for FairMCResult
 #include "FairMCStage.h"           // for FairMCStage
--- a/datamatch/FairMCMatchCreatorTask.h
+++ b/datamatch/FairMCMatchCreatorTask.h
@@ -19,6 +19,10 @@
 #ifndef FAIRMCMATCHCREATORTASK_H
 #define FAIRMCMATCHCREATORTASK_H
 
+#ifndef FAIR_SILENCE_DATAMATCH_DEPRECATION
+#warning "FairRoot::DataMatch is deprecated and will be removed in a future major release"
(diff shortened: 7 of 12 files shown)


CMakeLists.txt
```diff
@@ -109,6 +109,7 @@ option(ENABLE_GEANT3_TESTING "Enable tests utilizing Geant3" OFF)
 option(BUILD_PROOF_SUPPORT "Support ROOT::Proof (deprecated)" OFF)
 
 option(BUILD_ONLINE "Build the online library" ON)
+option(BUILD_DATAMATCH "Build datamatch" OFF)
```
ChristianTackeGSI, line 112:
```suggestion
option(BUILD_DATAMATCH "Build datamatch" ON)
```

* Could you leave it at ON for the main commit?
* Then add another commit that explicitly disables it?
  That commit maybe should also add to the v19 ChangeLog?
* Then we can backport the first commit straight to 18.8

CHANGELOG.md
```diff
@@ -126,6 +126,11 @@ file an issue, so that we can see how to handle this.
     * It will be disabled by default in FairRoot 19.
     * We plan to remove it completely in FairRoot 20.
   * If you need it, speak up NOW.
+* Deprecating DataMatch
+  * This release deprecates MbsAPI.
```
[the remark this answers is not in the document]
karabowi: Right, fixed.

CMakeLists.txt
karabowi: Done.

dennisklein (changes requested):
The commit messages and their changes are confusing

c03c50a12e05220215fbc05dc5f7a66b77d89036 adds the `BUILD_DATAMATCH` option and declares the library deprecated in the changelog but says nothing about it in its commit message.

But b3fdd2785a6914715e3a9804e940973b4b9c4c0b says `Deprecate DataMatch` and in fact switches the default of the `BUILD_DATAMATCH` option which has nothing to do with the deprecation. It should say `Disable DataMatch by default` or similar, but nothing about deprecation, IMHO.

---

In the release which actually deprecates the library, we could add e.g. in `datamatch/CMakeLists.txt` at the top

```cmake
message(DEPRECATION "The FairRoot::DataMatch library is deprecated and will be removed without replacement in a future major release of FairRoot")
```

---

And on a second thought, declaring something deprecated and still building it by default, is confusing as well if I am honest. We should only declare deprecation after we switched them off by default and removed all our own usages (may happen in the same release of course).

---

And finally, this entry needs different wording IMHO (also the other ones this one is copied from):
```
* Deprecating DataMatch
  * This release deprecates DataMatch.     // what information does this line add to the previous?
    * It will be disabled by default in FairRoot 19. 
    * We plan to remove it completely in FairRoot 20.    // for my taste, let's just say "in a future major release" - no need to be more predictive.
  * If you need it, speak up NOW.   // this i don't understand at all, a changelog should not be a usage questionnaire (*where* and *until when* should a user speak up?) and the confusing part is, that this statement immediately makes me question the stated plan two lines above. what if a user speaks up and says, she needs it, then we revert the deprecation and scrap the plan for removal or what??
```

dennisklein:
There are three distinct steps in exactly the following order:
1. Add the library build switch, but change nothing -> changelog: "CMake: Add `BUILD_DATAMATCH` option (default `ON`)"
2. Disable the library by default -> changelog: "Disabled DataMatch library by default, opt-in via `-DBUILD_DATAMATCH=ON`"
3. Deprecate the library -> changelog: "Deprecated DataMatch library"

If you release 1. and 2. together, then you can just drop the changelog from 1. completely.

If you release 2. and 3. together, just put one changelog line: "Disabled and deprecated DataMatch library. You may still build it via `-DBUILD_DATAMATCH=ON`."

ChristianTackeGSI:
Looking at the way `[[deprecated]]` works in C++, I think, disabling and declaring deprecated should work the other way around:

You first deprecate a symbol in C++. It still fully works as expected, the user doesn't have to change _anything_ (unless they turn deprecation warnings into errors–other story).

I haven't quickly found a way to declare a target in CMake as deprecated. So the next best thing we can do: Write a ChangeLog entry. In the C++ world, you could say, that this is a heads up notice. I think, you sometimes do that in FairMQ as well?

After that we can turn this into an actual error by disabling the thing by default. The user can still get the target back by switching it on again. So that's the extended "warning" way.

And finally we remove it.

In 18.8 I would like to let many things "still just work". So I would prefer a heads up notice in the changelog there. (And we can already have the switch, just defaulting to ON.)

Just my 2 cent…

dennisklein:
> You first deprecate a symbol in C++. It still fully works as expected, the user doesn't have to change anything

No, related to DataMatch: As a user I would ask, why do you spam me with deprecation warnings, I am using *YOUR* defaults.

ChristianTackeGSI:
Spam… ChangeLog.

Well, I don't care too much. Go ahead with whatever seems right for you!

CHANGELOG.md
```diff
@@ -166,4 +168,4 @@ file an issue, so that we can see how to handle this.
   those tests have a probability > 0 for failing.
   If you want to run them anyways, pass
   `-DENABLE_GEANT3_TESTING=ON` to CMake.
-* CMake: added `BUILD_DATAMATCH` build switch, default is `ON`.
+* CMake: added `BUILD_DATAMATCH` build switch, default is `OFF`.
```
dennisklein, line 171: So, what is now the plan? The first commit is released with 18.8? The second commit with 19.0? If yes, then why would a change in the 19.0 release modify the changelog text of a previous release?

dennisklein (changes requested):

rbx:
For the next release I would add the warning as Dennis suggets:
```
message(DEPRECATION "The FairRoot::DataMatch library is deprecated and will be removed without replacement in a future major release of FairRoot")
```
And in the future major release just remove the DataMatch library.

As for the cmake switch, add it now and keep it on, in the spirit of "It still fully works as expected, the user doesn't have to change anything". DataMatch was always built before, right? So the deprecation step should build it by default. Actually why is this switch needed? Deprecate now via msg and delete in next major release is enough. Unless someone wants the switch now for whatever reason.

dennisklein:
> Actually why is this switch needed?

Yes, talked a bit more with Christian, in principle the switch is not so much related to the deprecation/removal. We should have different motivation for it.

> For the next release I would add the warning as Dennis suggets:
> ```
> message(DEPRECATION "The FairRoot::DataMatch library is deprecated and will be removed without replacement in a future > major release of FairRoot")
> ```

Another idea we had was to put a `#warning "DataMatch is deprecated, do not use it any more."` to all headers that belong to it. That way, really only users of this library would be bothered.

karabowi:
> Another idea we had was to put a `#warning "DataMatch is deprecated, do not use it any more."` to all headers that belong to it. That way, really only users of this library would be bothered.

This idea has an undesirable effect, because (without disabling datamatch) it prints hundreds of warnings when building FairRoot itself.

dennisklein:
> > Another idea we had was to put a `#warning "DataMatch is deprecated, do not use it any more."` to all headers that belong to it. That way, really only users of this library would be bothered.
> 
> This idea has an undesirable effect, because (without disabling datamatch) it prints hundreds of warnings when building FairRoot itself.

Well, you probably want to add a switch for this warning anyways, so a user, who still wanted to use it, can silence the deprecation warnings as well, and then use this switch during compilation of the library itself? e.g.

```diff
--- a/datamatch/CMakeLists.txt
+++ b/datamatch/CMakeLists.txt
@@ -39,6 +39,10 @@ target_link_libraries(${target} PUBLIC
   ROOT::Core
 )

+target_compile_definitions(${target} PRIVATE
+  FAIR_SILENCE_DATAMATCH_DEPRECATION
+)
+
 fairroot_target_root_dictionary(${target}
   HEADERS ${headers}
   LINKDEF LinkDef.h
--- a/datamatch/FairMCDataCrawler.h
+++ b/datamatch/FairMCDataCrawler.h
@@ -15,6 +15,10 @@
 #ifndef FAIRMCDATACRAWLER_H_
 #define FAIRMCDATACRAWLER_H_

+#ifndef FAIR_SILENCE_DATAMATCH_DEPRECATION
+#warning "FairRoot::DataMatch is deprecated and will be removed in a future major release"
+#endif
+
 #include "FairMultiLinkedData.h"   // for FairMultiLinkedData

 #include <Rtypes.h>    // for Int_t, Bool_t, etc
```

dennisklein (approved):

PR #1289, opened 2022-11-21
clang-tidy: enable many bugprone-*

ChristianTackeGSI:
clang-tidy's bugprone-* checks flag common issues that can cause real bugs.


dennisklein (approved):
Fine with me.

PR #1290, opened 2022-11-22
Add Fair{Run,EventManager}::GetRootManager()

ChristianTackeGSI:
FairRun, and FairEventManager have a FairRootManager linked from them. Expose this, so it can be used.

Start to use it in some few places.

---

Checklist:

### DIFF
--- a/base/steer/FairTSBufferFunctional.cxx
+++ b/base/steer/FairTSBufferFunctional.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ * Copyright (C) 2014-2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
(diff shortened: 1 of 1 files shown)


base/steer/FairRun.h
```diff
@@ -199,6 +199,7 @@ class FairRun : public TNamed
     virtual void SetSource(FairSource* tempSource);
     /** Return non-owning pointer to source **/
     FairSource* GetSource() { return fSource.get(); }
+    FairRootManager* GetRootManager() const { return fRootManager; }
```
dennisklein, line 202:
```suggestion
    FairRootManager& GetRootManager() const { return fRootManager; }
```

I guess, we can make sure to always return one, right?

ChristianTackeGSI: Yes, we assume that anyway everywhere. And we're quiet sure also.

That said, most of the calling code expects a pointer. That said, let's see, whether we can rewrite them to use references!

ChristianTackeGSI: done.

dennisklein (approved):
Looks good to me.

One thought that comes up still, is "`const` by default". Have you checked if those three variables you touched, could actually be `const`?

ChristianTackeGSI:
done, added const to local variables.

PR #1299, opened 2022-11-28
Drop TestNonDefaultContext from parbase Usage

ChristianTackeGSI:
All setups of FairContainer in FairRoot and its examples add an extra / non default "context" to FairContainer named "TestNonDefaultContext". It doesn't look like it's being used anywhere. So drop this.

---

Checklist:


karabowi (approved):
We will also have to trigger experiments to remove these lines (12 times in cbmroot and 79 times in pandaroot), but this PR does not break experiment code.

ChristianTackeGSI:
> We will also have to trigger experiments to remove these lines […]

Should we add a notice to the changelog?

karabowi:
I think it's up to the experiments to clean their code, let's just point that out during the next meeting.

PR #1305, opened 2022-12-06
Add FairRun::GetEvtHeaderRunId and use it

ChristianTackeGSI:
Classes derived from FairRun often call GetRunId() on fEvtHeader. So let's have a shortcut.


karabowi (approved):

ChristianTackeGSI:
Thanks!

Do we want to have this in 18.8.x as well? If so, should I open a backport PR?
Or only in 19.0?

karabowi:
I'd add it to 18.8.1 as well.

Issue #1301, opened 2022-12-01
Return value check of GetSource()->InitUnpackers()


inkdot7:
The return value of `GetSource()->InitUnpackers()` is not checked in

https://github.com/FairRootGroup/FairRoot/blob/95b6b43de7ff33308989473c2cf90132effc3fa8/online/steer/FairRunOnline.cxx#L192

This leads to inconsistent results in r3broot, see https://github.com/R3BRootGroup/R3BRoot/issues/705

Should the return value be checked in `FairRunOnline::Init()` ?

karabowi:
@inkdot7 , could you test the PR #1307 and leave us feedback on the performance?

inkdot7:
Thanks for merging the fix!

PR #1307, opened 2022-12-09
Check the return value `source->InitUnpackers`

karabowi:
Return values of the calls to `source->InitUnpackers()` and `source->ReinitUnpackers()` in `FairRunOnline` are now checked. Should the value be `false`,
the fatal message will be printed.

Addresses issue #1301.

---

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,7 +65,7 @@ file an issue, so that we can see how to handle this.
   in `FairRunOnline`. Stop run if `false` returned.
 
 
-## 18.8 (UNRELEASED) - 2022-11-XX
+## 18.8.0 - 2022-12-16
 
 ### Breaking Changes
 * Move online related code into the new Online library
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,13 @@ file an issue, so that we can see how to handle this.
 * https://github.com/R3BRootGroup/R3BRoot/pull/413
 
 
+## 18.8.1 (UNRELEASED) - 2023-01-XX
+
+### Bug fixes
+* Check the return value of `source->InitUnpackers()`/`source->ReinitUnpackers()`
+  in `FairRunOnline`. Stop run if `false` returned.
+
+
 ## 18.8 (UNRELEASED) - 2022-11-XX
 
 ### Breaking Changes
@@ -111,7 +118,6 @@ file an issue, so that we can see how to handle this.
   ```
   There is also a convenience header `<FairBoostSerializationSupport.h>` which includes all
   class-based support headers currently offered.
-
 
 ### Deprecations
 
--- a/online/steer/FairRunOnline.cxx
+++ b/online/steer/FairRunOnline.cxx
@@ -43,7 +43,10 @@ using std::endl;
 
 FairRunOnline* FairRunOnline::fgRinstance = nullptr;
 
-FairRunOnline* FairRunOnline::Instance() { return fgRinstance; }
+FairRunOnline* FairRunOnline::Instance()
+{
+    return fgRinstance;
+}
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -111,6 +111,8 @@ file an issue, so that we can see how to handle this.
   ```
   There is also a convenience header `<FairBoostSerializationSupport.h>` which includes all
   class-based support headers currently offered.
+* Check the return value of `source->InitUnpackers()`/`source->ReinitUnpackers()`
+  in `FairRunOnline`. Stop run if `false` returned.
```
ChristianTackeGSI, line 115: As 18.8 was released already, either
* move this into the 19.0 section (if it only targets that release),
* or create a new 18.8.1 section and add it there (if you target 18.8.x)

online/steer/FairRunOnline.cxx
```diff
@@ -189,7 +195,10 @@ void FairRunOnline::Init()
     fRootManager->Register("EventHeader.", "Event", fEvtHeader, (nullptr != GetSink()));
     fEvtHeader->SetRunId(fRunId);
 
-    GetSource()->InitUnpackers();
+    if (GetSource()->InitUnpackers() == false) {
```
ChristianTackeGSI, line 198:
```suggestion
    if (!GetSource()->InitUnpackers()) {
```

to avoid the warning from clang-tidy.

online/steer/FairRunOnline.cxx
```diff
@@ -250,7 +259,10 @@ Int_t FairRunOnline::EventLoop()
         LOG(info) << "FairRunOnline::EventLoop() Call Reinit due to changed RunID (from " << fRunId << " to " << tmpId;
         fRunId = tmpId;
         Reinit(fRunId);
-        GetSource()->ReInitUnpackers();
+        if (GetSource()->ReInitUnpackers() == false) {
```
ChristianTackeGSI, line 262:
```suggestion
        if (!GetSource()->ReInitUnpackers()) {
```

ChristianTackeGSI (changes requested):

CHANGELOG.md
```diff
@@ -58,6 +58,13 @@ file an issue, so that we can see how to handle this.
 * https://github.com/R3BRootGroup/R3BRoot/pull/413
 
 
+## 18.8.1 (UNRELEASED) - 2023-01-XX
+
+### Breaking Changes
+* Check the return value of `source->InitUnpackers()`/`source->ReinitUnpackers()`
+  in `FairRunOnline`. Stop run if `false` returned.
```
ChristianTackeGSI, line 65: Would you really call this a breaking change? The return value wasn't used before and is now used. It could even be considered a bugfix?

In any case it will only break one use case: People return `false` and want things to continue. But why did they return `false` in the first place?

CHANGELOG.md
```diff
@@ -58,6 +58,13 @@ file an issue, so that we can see how to handle this.
...
 
 
+## 18.8.1 (UNRELEASED) - 2023-01-XX
+
+### Bug fixes
+* Check the return value of `source->InitUnpackers()`/`source->ReinitUnpackers()`
+  in `FairRunOnline`. Stop run if `false` returned.
+
+
 ## 18.8 (UNRELEASED) - 2022-11-XX
```
ChristianTackeGSI, line 68: While you're here?

```suggestion
## 18.8.0 - 2022-12-16
```

ChristianTackeGSI (dismissed):

CHANGELOG.md
karabowi: Good idea.

ChristianTackeGSI:
Please remember to backport to 18.8.

PR #1309, opened 2022-12-15
Fix the FairRoot targets

### LINKED ISSUE #1288: Clean target dependencies
* Make sure private dependencies are private

fuhlig1:
Remove unnecessary dependencies.
Up to now all dependencies where defined as PUBLIC ones. With this commit many where converted to PRIVATE dependencies such that they are not exposed to the consumers.

This PR closes #1288.

---

Checklist:

* [ x] Rebased against `dev` branch
* [x ] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -98,24 +98,26 @@ target_include_directories(${target} PUBLIC
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 
-target_link_libraries(${target} PUBLIC
+target_link_libraries(${target}
+  PUBLIC
   FairRoot::Alignment
   FairRoot::Tools
   FairRoot::ParBase
   FairRoot::GeoBase
 
   ROOT::Core
-  ROOT::EG
   ROOT::Geom
   ROOT::GeomPainter # TGeoTrack
   ROOT::Hist
   ROOT::MathCore
   ROOT::Matrix
-  ROOT::Net # TSocket
   ROOT::Physics
   ROOT::RIO
   ROOT::Tree
   ${VMCLIB}
+
+  PRIVATE
+  ROOT::EG
   $<$<BOOL:${ROOT_gdml_FOUND}>:ROOT::Gdml>
 )
 
@@ -126,7 +128,7 @@ target_compile_definitions(${target}
 )
 
 if(BUILD_PROOF_SUPPORT)
-  target_link_libraries(${target} PUBLIC ROOT::Proof)
+  target_link_libraries(${target} PRIVATE ROOT::Proof)
   target_compile_definitions(${target} PRIVATE BUILD_PROOF_SUPPORT)
 endif()
 
--- a/eventdisplay/CMakeLists.txt
+++ b/eventdisplay/CMakeLists.txt
@@ -54,23 +54,28 @@ target_include_directories(${target} PUBLIC
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 
-target_link_libraries(${target} PUBLIC
-  FairRoot::Tools
+target_link_libraries(${target}
+  PUBLIC
   FairRoot::Base # FairRootManager, FairRunAna, FairTSBufferFunctional, FairTimeStamp, FairEventManager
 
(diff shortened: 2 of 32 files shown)


base/CMakeLists.txt
```diff
@@ -98,24 +98,27 @@ target_include_directories(${target} PUBLIC
...
   ROOT::Matrix
-  ROOT::Net # TSocket
   ROOT::Physics
   ROOT::RIO
   ROOT::Tree
   ${VMCLIB}
+
+  PRIVATE
+  ROOT::EG
+  ROOT::Proof
```
ChristianTackeGSI, line 121: Are you sure about this unconditional ROOT::Proof?

We're trying to deprecate it. And I think, we have some `BUILD_PROOF_SUPPORT` option in CMake for this.

eventdisplay/CMakeLists.txt
```diff
@@ -54,23 +54,28 @@ target_include_directories(${target} PUBLIC
...
 )
 
-target_link_libraries(${target} PUBLIC
-  FairRoot::Tools
+target_link_libraries(${target}
+  PUBLIC
   FairRoot::Base # FairRootManager, FairRunAna, FairTSBufferFunctional, FairTimeStamp, FairEventManager
 
+  FairLogger::FairLogger
+
```
ChristianTackeGSI, line 62: Hmm, I don't see any `#include.*[lL]ogger` in `eventdisplay/*.h`?

ChristianTackeGSI (commented):
> Checklist:
> 
>     * [ x] Rebased against `dev` branch
> 
>     * [ x] My name is in the resp. CONTRIBUTORS/AUTHORS file
> 
>     * [x ] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

As a core maintainer, this is probably useless, just remove that part next time?
```markdown
* [X] Or format it like this
```

base/CMakeLists.txt
fuhlig1: I am not sure. I only saw that we use a Proof header file. That's why I have added it. If Proof is build only conditional than the dependency should be also conditional for sure.

eventdisplay/CMakeLists.txt
fuhlig1: If I do

`find . -name "*.h" -exec grep -H include {} \; | cut -f2 -d' ' | sort | uniq
`
I see 

`<fairlogger/Logger.h>`

`find . -name "*.h" -exec grep -H fairlogger {} \;`

gives 

```
./datasource/FairDataSourceI.h:#include <fairlogger/Logger.h>
./datasource/FairTimebasedDataHandlerT.h:#include <fairlogger/Logger.h>
```

Si in my opinion the dependency is correct.

base/CMakeLists.txt
fuhlig1: @ChristianTackeGSI,

would this be the right syntax?

`$<$<BOOL:${BUILD_PROOF_SUPPORT>:ROOT::Proof>`

eventdisplay/CMakeLists.txt
ChristianTackeGSI: Okay, you're right. I missed that for whatever reason.

Marking resolved.

base/CMakeLists.txt
ChristianTackeGSI: I think, that should work.

Alternatively, this is very explicit, but should work as well:
```cmake
if(BUILD_PROOF_SUPPORT)
  target_link_libraries(${target} PRIVATE ROOT::Proof)
endif()
```

Looking again, I think, this is _already_ in `base/CMakeLists.txt` at around line 130?

https://github.com/FairRootGroup/FairRoot/blob/a0ba5f20373308e895614f78b6464e71af5516f4/base/CMakeLists.txt#L128-L131

You might need to change the PUBLIC over to PRIVATE there?

fuhlig1: @ChristianTackeGSI,

thanks for pointing out the existing code block. I completely overlooked it.
The issue should be fixed with the latest commit.

ChristianTackeGSI (approved):

karabowi (approved):

PR #1244, opened 2022-10-20
Improve geometry state after alignment

PALoizeau:
1. Refresh the geometry once done (needed in any case)
2. Fix for GeoAssembly bounding box having 0 dx/dy after alignment (#1243)

Fix could be removed once proper fix is done in Root, should not hurt if done twice

Should be backported to the next v18.6.x if accepted as probably needed for next CbmRoot release.

---

Checklist:

* [x] Rebased against `dev` branch
* [x] My name is in the resp. CONTRIBUTORS/AUTHORS file
* [x] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

PR #1310, opened 2022-12-15
Cleanup FairAlignmentHandler

ChristianTackeGSI:
* Apply some clang-tidy cleanups
* Use TRangeDynCast in RecomputePhysicalAssmbBbox

@PALoizeau as you authored #1244, could you take a look at this?

### DIFF
--- a/alignment/FairAlignmentHandler.cxx
+++ b/alignment/FairAlignmentHandler.cxx
@@ -1,18 +1,14 @@
 #include "FairAlignmentHandler.h"
 
-#include "FairLogger.h"
-
+#include <TCollection.h>
 #include <TGeoManager.h>
 #include <TGeoPhysicalNode.h>
 #include <TGeoShapeAssembly.h>
-
-FairAlignmentHandler::FairAlignmentHandler() {}
-
-FairAlignmentHandler::~FairAlignmentHandler() {}
+#include <fairlogger/Logger.h>
 
 void FairAlignmentHandler::AlignGeometry() const
 {
-    if (fAlignmentMatrices.size() > 0) {
+    if (!fAlignmentMatrices.empty()) {
         LOG(info) << "aligning the geometry...";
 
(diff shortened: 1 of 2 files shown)


alignment/FairAlignmentHandler.cxx
```diff
@@ -106,14 +102,9 @@ void FairAlignmentHandler::RecomputePhysicalAssmbBbox() const
 {
     TObjArray* pPhysNodesArr = gGeoManager->GetListOfPhysicalNodes();
 
-    TGeoPhysicalNode* pPhysNode = nullptr;
-    TGeoShapeAssembly* pShapeAsb = nullptr;
-
-    Int_t iNbNodes = pPhysNodesArr->GetEntriesFast();
-    for (Int_t iInd = 0; iInd < iNbNodes; ++iInd) {
-        pPhysNode = dynamic_cast<TGeoPhysicalNode*>(pPhysNodesArr->At(iInd));
+    for (auto pPhysNode : TRangeDynCast<TGeoPhysicalNode>(pPhysNodesArr)) {
```
fuhlig1, line 105: Looks good. Did not know this.

PALoizeau: Same for me, nice to learn about this functionality, it may be useful later (even we use less ROOT style containers/collection with direct access recently).

PALoizeau:
I am not the original author of most of these blocks, but I think this looks good and most of the changes would be what I would do if asked to cleanup/refresh a class. So :+1:  from my side

ChristianTackeGSI:
@karabowi Do we want this in 19.0? Then you or @fuhlig1 might just press the merge button?

If we want it in 19.2, then we should set the milestone and wait until after the release of 19.0?

fuhlig1:
In my opinion it would be good to have it even in 18.8.

fuhlig1 (approved):

Issue #1286, opened 2022-11-19
 `ex_MQ_serialization1` should be stricter


dennisklein:
Logs: https://cdash.gsi.de/testDetails.php?test=12812982&build=356755

```
Error in <TExMap::Remove>: key 1949127801 not found at 286
Warning in <TMessage::CheckObject>: reference to object of unavailable class TObject, offset=1949127801 pointer will be 0

 *** Break *** segmentation violation
 ```

Another line that looks suspicious is:
```
[#0] WARNING:InputArguments -- The parameter 'fSigmaT' with range [-1e+30, 1e+30] of the RooGaussian 'fGaussT' exceeds the safe range of (0, inf). Advise to limit its range.
```

```
 ===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================

Thread 6 (Thread 0x7f517f478640 (LWP 35583) "ex-serializatio"):
#0  0x00007f51949259d9 in __futex_abstimed_wait_common () from /lib64/libc.so.6
#1  0x00007f51949287c7 in pthread_cond_clockwait
GLIBC_2.30 () from /lib64/libc.so.6
#2  0x00007f51977d74ea in std::__condvar::wait_until (__abs_time=..., __clock=1, __m=..., this=<optimized out>) at /usr/include/c++/12/bits/std_mutex.h:123
#3  std::condition_variable::__wait_until_impl<std::chrono::duration<long, std::ratio<1l, 1000000000l> > > (__lock=..., __atime=..., this=<optimized out>) at /usr/include/c++/12/condition_variable:200
#4  std::condition_variable::wait_until<std::chrono::duration<long, std::ratio<1l, 1000000000l> > > (__atime=..., __lock=..., this=<optimized out>) at /usr/include/c++/12/condition_variable:110
#5  std::condition_variable::wait_until<std::chrono::_V2::steady_clock, std::chrono::duration<long int, std::ratio<1, 1000000000> >, fair::mq::StateMachine::WaitForPendingStateFor(int) const::<lambda()> > (__p=..., __atime=..., __lock=..., this=<optimized out>) at /usr/include/c++/12/condition_variable:151
#6  std::condition_variable::wait_for<long int, std::ratio<1, 1000>, fair::mq::StateMachine::WaitForPendingStateFor(int) const::<lambda()> > (__rtime=..., __p=..., __lock=..., this=<optimized out>) at /usr/include/c++/12/condition_variable:174
[...]

PR #1311, opened 2022-12-24
fix(Ex1Processor): Handle nullptr, cleanup

ChristianTackeGSI:
fix part of #1286: The result of ROOT's deserializer should be checked.


karabowi (approved):

ChristianTackeGSI:
Thanks!

As this fixes part of #1286 (which targets 18.8.x), should I open a backport PR?

karabowi:
Yeah, go for it.

PR #1315, opened 2023-01-10
Start to enable two bugprone-* checks

ChristianTackeGSI:
Enable two `bugprone-*` checks in clang-tidy.

Fix the issues found and improve related code:

* Fix sizeof in Serializer

* Improve/Fix the Constructors of FairOnlineSink and utilize the improved copy constructor.

---

Checklist:

### DIFF
--- a/examples/MQ/pixelDetector/src/FairOnlineSink.cxx
+++ b/examples/MQ/pixelDetector/src/FairOnlineSink.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *         GNU Lesser General Public Licence version 3 (LGPL) version 3,        *
@@ -18,16 +18,6 @@
 #include "FairRootManager.h"
 
 #include <fairlogger/Logger.h>
-
-FairOnlineSink::FairOnlineSink()
-    : FairSink()
-    , fMQRunDevice(nullptr)
-{}
-
-FairOnlineSink::FairOnlineSink(const FairOnlineSink&)
-    : FairSink()
-    , fMQRunDevice(nullptr)
-{}
 
 //_____________________________________________________________________________
 void FairOnlineSink::RegisterImpl(const char*, const char*, void*) { return; }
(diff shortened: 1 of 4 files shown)


dennisklein (commented):
The second commit "fix bugprone-copy-constructor-init" confuses me, it seems to change several more things than stated in the commit?!

ChristianTackeGSI:
Thanks! Rearranged the commits somewhat.

dennisklein (approved):

examples/MQ/pixelDetector/src/FairOnlineSink.cxx
```diff
@@ -19,15 +19,9 @@
 
 #include <fairlogger/Logger.h>
 
-FairOnlineSink::FairOnlineSink()
-    : FairSink()
-    , fMQRunDevice(nullptr)
-{}
+FairOnlineSink::FairOnlineSink() = default;
```
dennisklein, line 22: I think the `= default` needs to go to the header file?

dennisklein (changes requested):

examples/MQ/pixelDetector/src/FairOnlineSink.h
```diff
@@ -24,14 +24,13 @@
...
 class FairMQRunDevice;
-class TObject;
-class TTree;
 
 class FairOnlineSink : public FairSink
 {
   public:
     FairOnlineSink();
     ~FairOnlineSink() override = default;
+    FairOnlineSink& operator=(const FairOnlineSink&) = delete;
```
dennisklein, line 33: What speaks against `= default` here?

examples/MQ/pixelDetector/src/FairOnlineSink.cxx
ChristianTackeGSI: I think, both is allowed. But yes, putting it in the header allows more optimizations. So that's a good thing.

Done.

examples/MQ/pixelDetector/src/FairOnlineSink.h
ChristianTackeGSI: The current copy constructor is private only. So for the public API, this class is not copyable in any way.

I guess FairSinks are meant as "resources" like a mutex.

CloneSink is the only internal user of the copy constructor. CloneSink is meant for multithreading and expected to do strange things (which is another topic in itself).

ChristianTackeGSI: I added a comment to the copy constructor:
```cpp
/// internal helper function for CloneSink()
```

I hope this makes it more obvious?

dennisklein (approved):

PR #1316, opened 2023-01-12
Add `CONTRIBUTING.md`

ChristianTackeGSI:
Initial work done by @dennisklein in #1277.
(Took his commit directly, just rebased.)

Added the notes from the discussion in the original PR.

Rendered: https://github.com/ChristianTackeGSI/FairRoot/blob/docs-contributing/CONTRIBUTING.md

---

Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,22 +12,12 @@ Create a [Github Pull Request](https://github.com/FairRootGroup/FairRoot/compare
    branch and base your PR against it.
    * In rare cases (e.g. backports, some hotfixes) base against the appropriate
      branch.
-2. Follow [the seven rules of a great Git commit message](https://cbea.ms/git-commit/#seven-rules)!
-   * Use a meaningful commit granularity, e.g. do not mix
-     unrelated changes in a single commit and vice verse squash related commits
-     into one.
-   * Expect that a reviewer will ask you for restructuring your commits! This
-     usually happens towards the end of the lifetime of a PR when it matured
-     enough for merging.
-   * Utilize [the commit message body to explain *what* and *why* and not *how*](https://cbea.ms/git-commit/#why-not-how).
-     Aim your message body content to be meaningful (concise, but complete
-     enough) to a reader in the future!
-   * Utilize [reference keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)
-     both in commit messages as well as in PR comments if applicable.
-3. If you are a first time contributor, add a separate commit in your PR which
+2. If you are a first time contributor, add a separate commit in your PR which
    adds your name to the [`CONTRIBUTORS`](CONTRIBUTORS) file.
-4. Follow our [Coding Guidelines](#coding-guidelines).
-5. Address failed CI checks.
+3. Follow our [Coding Guidelines](#coding-guidelines).
+4. Expect that a reviewer will ask you for restructuring your commits! This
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,109 @@
...
+
+## D - Documentation
+
+### D.1 Use [doxygen-style comments](https://www.doxygen.nl/manual/commands.html)
+
+## S - Style
+
+### S.1 Our CI enforces formatting with [`clang-format`](https://clang.llvm.org/docs/ClangFormat.html)
+
+**Config**: [`.clang-format`](.clang-format)
```
fuhlig1, line 75: You should exactly define which version is used in the CI and probably also fix  a version in the CI.
Unfortunately the changes produced by clang-format sometimes changes slightly from version to version.
We had this problem with CbmRoot in the beginning and where only able to solve it by providing a static version of clang-format for download which can be used practically on any Linux system. We also have a static binary for download for macosx .

fuhlig1 (changes requested):

CONTRIBUTING.md
ChristianTackeGSI: We had this exactly twice till now. And it boiled down to non precise declarations in `.clang-format`. Being more explicit in there fixed it for those cases.

See: ab58c0136e9f2a671365ebbb6a329b9c477c25ed, 3d7b969d6fbed0dabcf76ce17f8c9107bba3d8f1

That said, I guess, we haven't officially settled on something with this. Maybe the answer currently is: The clang-format that comes with the latest ubuntu (any, not only LTS) release. Or more generally: The CI is the authoritative version.

fuhlig1: It is fine to define clang-format to be the latests version. I checked for Ubuntu 22.10 the version is 15. How do I use it as Debian 10 or 11 user? On this systems the version is 7 respectively 11.

fuhlig1: I think all the discussions boil down to the question how we deal with failing clang-format tests. If we require that they must be fixed by the user we need to have a way that they can do this easily independent on their system.

ChristianTackeGSI: I see a few points here:

* Usually the current `.clang-format` produces the exact same result on Debian-11's clang-format (I am using that on my Desktop). And any differences–as already noted–usually boil down to imprecise declarations in `.clang-format` and should be fixed.

  That said, the differences are usually very minor and can easily be fixed by looking at the CI output and applying the changes manually. (If they're big, you could even cut'n'paste them into `patch`.)

* If you actually think, that this should be *very* easy for users, our CI could open a "reverse" Pull Request on the submitters branch and "suggest" the needed changes. But really, the changes are quite minor currently.

dennisklein: As a side note: In the FairMQ we opted to **not require** the format check. But this only works, if reviewers stay on top of this check and *care* about formatting.

> if you actually think, that this should be very easy for users, our CI could open a "reverse" Pull Request on the submitters branch and "suggest" the needed changes. But really, the changes are quite minor currently.

While it is fancy, I disliked such a feature so far (whenever I encountered it in other projects) because it likely leads to worse git history with tons of bot-authored, auto-generated "fix format" commits.

ChristianTackeGSI: I totally agree with trying to avoid bot authored commits in the final history. Reviewers should request a squash in that case.

And really, I currently really doubt that the investment into such a bot would pay off (as already explained).

ChristianTackeGSI:
As @fuhlig1 resolved his one request for changes by now: Can we merge this sometime soon? Maybe as a start for more Pull Requests to refine it?

karabowi (approved):

PR #1317, opened 2023-01-12
Event Display - moving xml configuration to separate classes.

DanielWielanek:
Hi,
I propose to remove configuration of geometry and colors from FairEventManager - now they are managed by separate classes. This slightly simplify the FEM and allow e.g. to set different track color scheme for different tasks (what is shown in tutorial with propagator that use different color for Geo and FairTracks).
I also think that in future it will be much more easy to add some stuff that is configured by XML files.

---

Checklist:

* [ ] Rebased against `dev` branch

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -4,7 +4,7 @@
   "name": "FairRoot",
   "description": "<p>A simulation, reconstruction and analysis framework that is based on the ROOT system. The user can create simulated data and/or perform analysis with the same framework. Geant3 and Geant4 transport engines are supported, however the user code that creates simulated data do not depend on a particular monte carlo engine. The framework delivers base classes which enable the users to construct their detectors and /or analysis tasks in a simple way, it also delivers some general functionality like track visualization. Moreover an interface for reading magnetic field maps is also implemented.</p>",
   "license": "https://spdx.org/licenses/LGPL-3.0-only",
-  "version": "v18.2.1",
+  "softwareVersion": "19.0.0",
   "datePublished": "2012-05-25",
   "keywords": [
     "geant4",
@@ -16,9 +16,10 @@
     "analysis",
     "simulation"
   ],
-  "contributor": [
+  "author": [
     {
       "@type": "Person",
+      "@id": "https://orcid.org/0000-0002-8071-4497",
       "familyName": "Al-Turany",
       "givenName": "Mohammad"
     },
@@ -29,11 +30,13 @@
     },
     {
       "@type": "Person",
+      "@id": "https://orcid.org/0000-0003-3787-1910",
       "familyName": "Klein",
       "givenName": "Dennis"
     },
     {
       "@type": "Person",
+      "@id": "https://orcid.org/0000-0002-7537-2875",
       "familyName": "Kresan",
       "givenName": "Dmytro"
     },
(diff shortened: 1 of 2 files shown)


ChristianTackeGSI (commented):
Hello!

First of all: Thanks for this Pull Request! It is highly welcome!

Could you try to split this into more bite-sized parts?
I have tried to move the formatting fixes into #1318 

Maybe updating the examples could be in a follow up PR? That way we can make sure, that your update keeps some API compatibility with the previous version?

If you want to remove some parts please add a `[[deprecated]]` attribute and also a `\deprecated` doxygen comment.

All that said, I have added some comments here and there. Those are not yet final, just a start.

eventdisplay/xml/FairXMLDetectorConf.h
```diff
@@ -0,0 +1,52 @@
+/********************************************************************************
+ *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
```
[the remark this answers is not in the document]
ChristianTackeGSI: Well, this was meant for all files, not just this one.

eventdisplay/xml/FairXMLDetectorConf.h
```diff
@@ -0,0 +1,52 @@
...
+#define CBMROOT_CORE_EVENTDISPLAY_XML_FAIRXMLDETECTORCONF_H_
+
+#include "FairXMLEveConf.h"
+
+#include <Rtypes.h>
+#include <RtypesCore.h>
+#include <TString.h>
+#include <vector>
+
+class TGeoNode;
```
ChristianTackeGSI, line 27: Please include the proper include files for ROOT types.

```suggestion
#include <Rtypes.h>
#include <RtypesCore.h>
#include <TGeoNode.h>
#include <TString.h>
#include <vector>
```

eventdisplay/xml/FairXMLDetectorConf.cxx
```diff
@@ -0,0 +1,104 @@
...
+                    // node configured dont need anymore the daughter
+                    fChild.erase(std::next(fChild.begin(), j--));
+                    break;
+                }
+            }
+        }
+    }
+}
+
+void FairXMLDetectorConf::ChangeNode(TGeoNode* node) const
```
ChristianTackeGSI, line 96: It looks like this method expects `node` to be non-NULL? Why not use a reference then?

```suggestion
void FairXMLDetectorConf::ChangeNode(TGeoNode& node) const
```

eventdisplay/xml/FairXMLDetectorConf.cxx
```diff
@@ -0,0 +1,104 @@
...
+        return;
+    ChangeNode(node);
+    if (depth > 0) {
+        for (int i = 0; i < node->GetNdaughters(); i++) {
+            SetupSubNode(node->GetDaughter(i), depth - 1);
+        }
+    }
+}
+
+void FairXMLDetectorConf::SetupNode(TGeoNode* node)
```
ChristianTackeGSI, line 72: It looks like this method expects node to be non-NULL? Why not use a reference then?
```suggestion
void FairXMLDetectorConf::SetupNode(TGeoNode& node)
```

eventdisplay/xml/FairXMLDetectorConf.h
```diff
@@ -0,0 +1,52 @@
...
+
+  public:
+    FairXMLDetectorConf(FairXMLNode* node = nullptr);
+    void SetupNode(TGeoNode* node);
+    void AddChild(FairXMLDetectorConf& set) { fChild.push_back(set); };
+    Int_t GetNChildren() const { return fChild.size(); }
+    TString GetDetectorName() const { return fName; }
+    const FairXMLDetectorConf& GetChild(Int_t n) { return fChild[n]; };
+    virtual ~FairXMLDetectorConf(){};
+    ClassDef(FairXMLDetectorConf, 1);
```
ChristianTackeGSI, line 49: Please use more override.

Are you expecting `FairXMLDetectorConf` to be used in Streamers? If not, please set the version to 0.

ChristianTackeGSI (commented):
I added a few more comments.

All of them are meant as hints. Even if there are suggested code changes, they usually will not work alone. They need changes in other parts of the code.
And they're meant to possibly improve other parts of the code.

So please take a look at the compiler errors and warnings.

ChristianTackeGSI:
Hi @DanielWielanek,

Could you please take a look at our Coding Guidelines, especially [G.3](https://github.com/FairRootGroup/FairRoot/blob/dev/CONTRIBUTING.md#g3-write-a-good-git-history) and rebase and rewrite the git history?

I would also propose to put 8986f59059f0871cf3d31bfedb4404b6b6ac875f into its own Pull Request, because it does something unrelated to the XML Config changes you're proposing here.

ChristianTackeGSI:
Hi @DanielWielanek,

cool to have you take on this again!

I have to apologize a bit, because we have changed eventdisplay somewhat by now and rebasing your changes onto that isn't easy (I tried myself, and it wasn't fun).

I also have to admit, that the amount of changes is too much for *me* to review (maybe somebody else would be able to…).

Just an idea: Maybe split this up in smaller Pull Requests? Maybe one adding the FairXMLPdgColor class? and using it in some places already?

PR #1329, opened 2023-01-23
FairRunSim: Improve memory handling

ChristianTackeGSI:
* Do not leak `fLoaderName`
* Use `unique_ptr` for `fApp`, `fGen,` `fMCEvHead`
* Drop unneeded `fMapName`

---

Checklist:


karabowi (approved):
Tested with PandaRoot. Requires some changes in PndMasterRunSim, as currently it assumes fGen to be protected pointer to FairPrimaryGenerator.
Requested review from Florian (CBM) and Dima (R3B).

ChristianTackeGSI:
Would it help to keep the new `std::unique_ptr` as a protected member variable instead of a private one?

karabowi:
Replaced by PR #1331.

ChristianTackeGSI:
#1331 has parts of what was here. Thanks for already merging that.

The core parts of memory handling (and the question about what to do with `fGen` (keep it as protected?) is also still open.

I have rebased it.

ChristianTackeGSI:
Okay, re-opening isn't possible, if one force-pushes before re-opening.

So created a new one: #1332

PR #1332, opened 2023-01-26
FairRunSim: Do not leak fLoaderName, drop/deprecate things

ChristianTackeGSI:
* Do not leak `fLoaderName`
* Deprecate `fGen`: Use Setter/Getter.
* Drop unneeded `fMapName`

Targetting 18.8.x, so that the deprecation is in there and the memory leak gets fixed.

Re-created, because re-opening #1329 wasn't possible. any longer.

---

Checklist:


ChristianTackeGSI:
I don't think, I deprecate the constructor?

```cpp
FairRunSim::FairRunSim(Bool_t isMaster)
```
Do you mean this one?

The pragma disables deprecation warnings for the constructor. The constructor initializes `fGet` to `nullptr`. So it access `fGen`! All the places inside FairRunSim that currently need to access fGen would throw a deprecation warning. We know that, we expect it. So we can calm it.

Once fGen becomes a private unique_ptr, we should remove all the pragmas again.

Does that explain it?

karabowi:
Thanks, @ChristianTackeGSI , for the explanation.

karabowi (approved):

PR #1336, opened 2023-02-01
Combine serialization examples into one

rbx:
Both examples differ only in the number of messages/parts transferred & serializer type, most of the other things are duplicated.

Here I combine them into one, use multiple serializers and simplify the structure a bit.


examples/MQ/serialization/data/MyDigi.h
```diff
@@ -32,9 +32,21 @@ class access;
...
 {
   public:
-    MyDigi();
-    MyDigi(Int_t x, Int_t y, Int_t z, Double_t timeStamp, Double_t timeStampError = 0.0);
-    ~MyDigi() override;
+    MyDigi()
+        : FairTimeStamp()
+        , fX(0)
+        , fY(0)
+        , fZ(0)
```
ChristianTackeGSI, line 39: Could you move those into default member initializers?

With those in place, this constructor should be able to be `= default;` as well?

examples/MQ/serialization/data/MyHit.h
```diff
@@ -33,20 +33,25 @@ class TVector3;
 class MyHit : public FairHit
 {
   public:
-    /** Default constructor **/
-    MyHit();
+    MyHit() : FairHit() {}
```
ChristianTackeGSI: I think, this should work?

```suggestion
    MyHit() = default;
```

examples/MQ/serialization/data/MyHit.cxx
```diff
@@ -1,40 +0,0 @@
...
-             Double_t timeStampError)
-    : FairHit(detID, pos, dpos, mcindex)   //
-{
-    SetTimeStamp(timeStamp);
-    SetTimeStampError(timeStampError);
-}
-
-MyHit::~MyHit() {}
-
-ClassImp(MyHit);
```
ChristianTackeGSI, line 40: Interesting to see that the `ClassImp` isn't needed any longer… or I don't see it in the `.h` now?

I probably should read up on when we need ClassImp and when not…

ChristianTackeGSI (commented):
A few comments… despite Github tells me, that they might be outdated.

examples/MQ/serialization/data/MyHit.cxx
dennisklein: Both `ClassImp` and `ClassDef` should not be needed as far as I understand.

examples/MQ/serialization/sampler.cxx
```diff
@@ -32,7 +41,7 @@ class Ex2Sampler : public fair::mq::Device
     void Init() override
     {
         fFileName = fConfig->GetValue<std::string>("input-file");
-        fInputFile = TFile::Open(fFileName.c_str(), "READ");
+        fInputFile = std::make_unique<TFile>(fFileName.c_str(), "READ");
```
ChristianTackeGSI, line 44:
```suggestion
        fInputFile.reset(TFile::Open(fFileName.c_str(), "READ"));
```

I think, TFile::Open is recommended over instantiating the TFile directly, because TFile::Open could return a specialized subclass (like for network access).

examples/MQ/serialization/sampler.cxx
```diff
@@ -6,24 +6,33 @@
...
+namespace bpo = boost::program_options;
+
+class Sampler : public fair::mq::Device
 {
   public:
-    Ex2Sampler()
+    Sampler()
         : fInput(nullptr)
         , fTree(nullptr)
         , fInputFile(nullptr)
```
ChristianTackeGSI, line 38:
```suggestion
```

A unique_ptr should be a good state by itself, right?

examples/MQ/serialization/data/MyDigi.h
rbx: Done.

examples/MQ/serialization/ExHeader.h
```diff
@@ -0,0 +1,27 @@
+#ifndef EXHEADER_H
+#define EXHEADER_H
```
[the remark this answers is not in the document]
rbx: I updated the copyright header everywhere.

examples/MQ/serialization/data/MyHit.h
rbx: Done.

examples/MQ/serialization/data/MyHit.cxx
rbx: @ChristianTackeGSI I don't understand it either. I consulted Dennis and Radek and both told me it should not be needed. I also tested thoroughly after the change, including the I/O that is done with this class in the example.

.gitignore
```diff
@@ -51,3 +51,6 @@ html/
 
 # VS Code directory
 .vscode
+
+# install directory
+install/
```
dennisklein, line 56: Btw, you can also add those directory on a per-user config, either for all your git repos e.g.
```
git config --global core.excludesfile $HOME/.gitignore
```
or per-repo e.g.
```
git config core.excludesfile .git/.user_gitignore
```

examples/MQ/serialization/sampler.cxx
rbx: Didn't know about that, I'll apply your suggestion then.

.gitignore
rbx: Cool, thx.

ChristianTackeGSI: There's also:

* `$XDG_CONFIG_HOME/git/ignore` (`~/.config/git/ignore` on most systems) for per-user
* `.git/info/exclude` for per-repo

examples/MQ/serialization/processor.cxx
```diff
@@ -1,56 +1,63 @@
...
-class Ex1Processor : public fair::mq::Device
+namespace bpo = boost::program_options;
+
+class Processor : public fair::mq::Device
 {
   public:
-    Ex1Processor() = default;
-    ~Ex1Processor() override = default;
+    Processor() = default;
+    ~Processor() override = default;
```
dennisklein, line 32:
```suggestion
```

see [CPPCG::C20](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c20-if-you-can-avoid-defining-default-operations-do)

examples/MQ/serialization/processor.cxx
```diff
@@ -1,56 +1,63 @@
...
+
+// std
 #include <memory>
 
-class Ex1Processor : public fair::mq::Device
+namespace bpo = boost::program_options;
+
+class Processor : public fair::mq::Device
 {
   public:
```
dennisklein, line 30:
```suggestion
struct Processor : fair::mq::Device
{
```

If you like `class` for consistency, fine with me, if `struct` is not objected by anyone, this could be shorter :P

rbx: +1 for struct

rbx:
Fixed some typos in the readme in the latest push.

examples/MQ/serialization/data/MyHit.cxx
ChristianTackeGSI: Looks like `ClassImp` got lately deprecated in ROOT:
* [Doxygen docs for ClassImp in master](https://root.cern/doc/master/Rtypes_8h.html#a3f9bf160bbbb44209e95556960836b96)
* [Commit adding the deprecation doxygen label](https://github.com/root-project/root/commit/52f133b62a04ef2b05eacf4e88c5d6730688e336#diff-7fb16dd1fa5a219aef72fcc8a7536b215e7c0df6d403f612fbd97adf0c7d4db8R374)

examples/MQ/serialization/sink.cxx
```diff
@@ -1,37 +1,44 @@
...
+
+struct Sink : fair::mq::Device
 {
-  public:
-    Ex2Sink()
+    Sink()
         : fInput(nullptr)
-        , fOutFile(nullptr)
-        , fTree("SerializationEx1", "Test output")
+        , fOutputFile(nullptr)
```
ChristianTackeGSI, line 32:
```suggestion
```

examples/MQ/serialization/sink.cxx
```diff
@@ -1,37 +1,44 @@
...
+        , fTree("SerializationEx", "Test output")
         , fNumMsgs(0)
     {}
 
     void Init() override
     {
         fNumMsgs = fConfig->GetValue<int>("num-msgs");
         fFileName = fConfig->GetValue<std::string>("output-file");
-        fOutFile = TFile::Open(fFileName.c_str(), "RECREATE");
+        fOutputFile = std::make_unique<TFile>(fFileName.c_str(), "RECREATE");
```
ChristianTackeGSI, line 41:
```suggestion
        fOutputFile.reset(TFile::Open(fFileName.c_str(), "RECREATE"));
```

ChristianTackeGSI (commented):
Some last bits…

examples/MQ/serialization/sink.cxx
rbx: Done.

examples/MQ/serialization/sink.cxx
rbx: Done.

.gitignore
ChristianTackeGSI: Is `install/` _the_ preferred prefix for installations?
In `FairRoot_build_test.cmake` we use `build/install/` for installing.

examples/MQ/serialization/ExHeader.h
```diff
@@ -0,0 +1,35 @@
...
+#include "FairMQ.h"   // for fair::mq::Message
+
+namespace boost::serialization {
+class access;
+}
+
+struct ExHeader
+{
+    int eventNumber = 0;
+    int detectorId = 0;
```
ChristianTackeGSI, line 21: I don't care much, but didn't we have some "member variables start with `f`" thing?

ChristianTackeGSI (approved):

examples/MQ/serialization/ExHeader.h
rbx: I sometimes make exceptions for the tiny structs. Can't explain why...
I'll follow the rule if there is one :)

dennisklein: > but didn't we have some "member variables start with f" thing?

There is no required hungarian notation AFAIK. FairRoot loosely adopted naming rules from ROOT, but there are already violating cases in various places.

That said, I personally dislike this hungarian notation and also don't see a technical reason to have it, so if we would put it up for a vote, I'd vote against having it.

> Can't explain why...

If we voted for having the f-prefix required, I would want to propose an exception for public member variable names which I think is roughly the case when Alexey would intuitively decide to use plain variable names.

rbx: > public member variable names

Sounds like a good condition for an exception to me :)

.gitignore
dennisklein: > Is install/ the preferred prefix for installations?

No. That is completely user defined.

One could argue this does not belong here. but we also have `build/` in it, which then consequently also shouldn't be there I guess.

dennisklein: I guess we put things in here in the past due to the lack of knowledge about how to add it on a per-user config level.

examples/MQ/serialization/ExHeader.h
ChristianTackeGSI: :+1: for the public member variable exception.

I'll resolve this with exactly that exception!

rbx:
Removed duplicate comment.

.gitignore
ChristianTackeGSI: > One could argue this does not belong here. but we also have `build/` in it, which then consequently also shouldn't be there I guess.

For `build/` one could argue, that our CI tools use that. So those are not chosen by the developer.

ChristianTackeGSI: Proposal: Let's add something like this at the top of the `.gitignore` and leave the rest as is for now?

```
#
# Ignore patterns specific to this project
#
# If you need personal additions, consider
# - `$XDG_CONFIG_HOME/git/ignore` (`~/.config/git/ignore` on most systems)
#   for per-user patterns
# - `.git/info/exclude` for this local repository
# Also see "man gitignore", and `core.excludesFile` in "man git-config"
#
```

ChristianTackeGSI (approved):

Issue #1295, opened 2022-11-25
`vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)` in `ex_MQ_serialization2`


dennisklein:
https://cdash.gsi.de/testDetails.php?test=12862953&build=357940

```
[ex-serialization2-processor][07:27:20.692320][INFO][Device.cxx:757:LogSocketRates] data1[0]: in: 1.4881 (0.196571 MB) out: 0 (0 MB)
[ex-serialization2-processor][07:27:20.692370][INFO][Device.cxx:757:LogSocketRates] data2[0]: in: 0 (0 MB) out: 0 (0 MB)
[ex-serialization2-processor][07:27:20.711638][STATE][StateMachine.cxx:101:on_exit] Exiting FairMQ state machine
[ex-serialization2-processor][07:27:20.712433][ERROR][runDevice.h:53:main] Uncaught exception reached the top of main: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
 ```

ChristianTackeGSI:
Does this only happen with 18.4?

I had really hoped for exceptions to have stack traces on them… but it looks like they don't.

I have no idea, what's going on. And I agree with @rbx that this looks possibly related to #1286.

I wonder, whether the `~TFile` error (in the same cdash) only happens because of the unclean exit, or whether it's another issue.

karabowi:
Closing as the examples had been rewritten in #1341 .
Many of the MQ test failures are probably connected to running tests in parallel,
which is no good for hardcoded IP addresses.

PR #1339, opened 2023-02-06
Serialization Examples: Diagnose Amount of Messages

ChristianTackeGSI:
To assist debugging, check the number of received messages.

This might help in debugging #1295

---

Checklist:

### DIFF
--- a/examples/MQ/serialization/processor.cxx
+++ b/examples/MQ/serialization/processor.cxx
@@ -36,11 +36,15 @@ struct Processor : fair::mq::Device
             fair::mq::Parts partsIn;
 
             if (Receive(partsIn, "data1") > 0) {
+                if (partsIn.Size() != 2) {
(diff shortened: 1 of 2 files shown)


examples/MQ/serialization/processor.cxx
```diff
@@ -36,9 +36,13 @@ struct Processor : fair::mq::Device
             fair::mq::Parts partsIn;
 
             if (Receive(partsIn, "data1") > 0) {
+                if (partsIn.Size() != 2) {
+                    LOG(warn) << "partsIn does not have 2 messages; skipping";
```
ChristianTackeGSI, line 40: Maybe we should use `error` here instead?

ChristianTackeGSI: done. Resolving.

rbx (approved):

PR #1341, opened 2023-02-07
Serialization example: Add data check on sender side

rbx:
As suggested in https://github.com/FairRootGroup/FairRoot/issues/1286#issuecomment-1415734862, adding a data check on sender side.

May not need to merge this, if we manage to reproduce something with the PR checks.

@dennisklein How frequent are https://github.com/FairRootGroup/FairRoot/issues/1286 and https://github.com/FairRootGroup/FairRoot/issues/1295?


rbx:
Formatter asks for:
```diff
-#include "MyHit.h" // temporary for the data check (see below)
+#include "MyHit.h"   // temporary for the data check (see below)
```
Is this an intentional rule to have extra spaces there?

dennisklein:
> Formatter asks for:
> 
> ```diff
> -#include "MyHit.h" // temporary for the data check (see below)
> +#include "MyHit.h"   // temporary for the data check (see below)
> ```
> 
> Is this an intentional rule to have extra spaces there?

Certainly not intentional as in "we want it", but clang-format does it as long as I know it.

ChristianTackeGSI:
Pardon me… How does FindHits do a "check"? I don't understand the "check". Could you please elaborate on this? (maybe by adding appropriate comments in the code?)

Having more checks is generally a good thing!

ChristianTackeGSI:
> Formatter asks for:
> 
> ```diff
> -#include "MyHit.h" // temporary for the data check (see below)
> +#include "MyHit.h"   // temporary for the data check (see below)
> ```
> 
> Is this an intentional rule to have extra spaces there?

I guess, it is this config setting?

https://github.com/FairRootGroup/FairRoot/blob/d7d342d356b1fcb9324b765c969e73ecfa7b0163/.clang-format#L38

which was introduced in 0c11f1329ce5aa42fe03b8de5d384fdc53a5b28c

rbx:
> Pardon me… How does FindHits do a "check"? I don't understand the "check". Could you please elaborate on this? (maybe by adding appropriate comments in the code?)
> 

In https://github.com/FairRootGroup/FairRoot/issues/1286 the processor crashes in FindHits, which could mean problem either in serialization or in the generated data. If Sampler would also crash in the same method, it would rule out serialization.

rbx:
> I guess, it is this config setting?
> 
> https://github.com/FairRootGroup/FairRoot/blob/d7d342d356b1fcb9324b765c969e73ecfa7b0163/.clang-format#L38
> 
> which was introduced in [0c11f13](https://github.com/FairRootGroup/FairRoot/commit/0c11f1329ce5aa42fe03b8de5d384fdc53a5b28c)

Oh, I see x)
Then I would add a commit and change it to `SpacesBeforeTrailingComments: 1`, if that's okay with you.

examples/MQ/serialization/sampler.cxx
```diff
@@ -64,6 +66,9 @@ struct Sampler : fair::mq::Device
             ExHeader header;
             header.eventNumber = idx;
 
+            // temporary data check before transfer
```
ChristianTackeGSI, line 69: Maybe change this a bit?

```suggestion
            // Temporarily apply data processing steps here to check that they work out.
```

ChristianTackeGSI (dismissed):

examples/MQ/serialization/sampler.cxx
rbx: Alright. Modified to something like you suggest.

ChristianTackeGSI:
> Oh, I see x) Then I would add a commit and change it to `SpacesBeforeTrailingComments: 1`, if that's okay with you.

I *personally* by now have a slight preference (got accustomed by now, probably) for the three spaces. But then I don't care too much. If you want more comments, start a fresh PR with only that commit?

examples/MQ/serialization/sampler.cxx
```diff
@@ -64,6 +66,9 @@ struct Sampler : fair::mq::Device
             ExHeader header;
             header.eventNumber = idx;
 
+            // temporarily apply data processing steps to check that they work before & after serialization
+            TClonesArray hits = FindHits(*fInput);
```
ChristianTackeGSI, line 70: I hope the optimizer is not removing this completely… Is there some construct in C++ to let the compiler think that this local variable is used?

ChristianTackeGSI (approved):

examples/MQ/serialization/sampler.cxx
rbx: Not sure. I could add a log message with the number of hits found.

dennisklein: I remember https://github.com/google/benchmark/blob/main/docs/user_guide.md#preventing-optimization, maybe the implementation of this helper reveals something useful.

rbx: How about `volatile`?

rbx: Looks like the benchmark uses `volatile`: https://github.com/google/benchmark/blob/main/include/benchmark/benchmark.h#L444
But what exactly the rest does I don't understand yet.

ChristianTackeGSI: I think, the magic is in the `asm`. AFAIK the compiler does not look at the assembler code at all. It just knows that the developer wants to do *something*. (We know, it's nothing, because we can look at the `""`.)

Guessing: the `"+m,r"(value)` might tell the compiler, that the assembler code needs `value` as input for read and/or write access and should give the assembler code access to it in some register or something. That way the compiler has to provide `value` (because it doesn't know what the assembler code actually does).

dennisklein:
> @dennisklein How frequent are https://github.com/FairRootGroup/FairRoot/issues/1286 and https://github.com/FairRootGroup/FairRoot/issues/1295?

I cannot easily query exact frequencies of these issues, but we see them mostly in the FairSoft tests. After a FairSoft legacy build on the various OSes we run a FairRoot 18.[4,6,8] compilation plus full FairRoot test suite against it. But because it is a nested CMake run within a CMake test, we don't have convenient *per-FairRoot-test-case* browsability on cdash. One has to look for failed Fair**Soft** tests and then read the full log what actually failed, e.g.

https://cdash.gsi.de/index.php?project=FairSoft&date=2022-12-06 has a couple of legacy build with failed tests, e.g. here on rockylinux it failed for all FairRoot releases: https://cdash.gsi.de/viewTest.php?onlyfailed&buildid=359802. In the output of the FairRoot 18.8 test e.g. you can then see that compilation worked, but the serialization test failed: https://cdash.gsi.de/testDetails.php?test=12944880&build=359802 (Note, we have the CTest retry count set to 5, but the test fails all 5 tries in this case) with the error:

```
Error in <TExMap::Remove>: key 1949127801 not found at 286
Warning in <TMessage::CheckObject>: reference to object of unavailable class TObject, offset=1949127801 pointer will be 0

 *** Break *** segmentation violation
 Generating stack trace...
[ex-serialization1-sink][19:10:28.869261][INFO][Device.cxx:757:LogSocketRates] data2[0]: in: 0 (0 MB) out: 0 (0 MB)
[ex-serialization1-sampler][19:10:29.008785][INFO][Device.cxx:757:LogSocketRates] data1[0]: in: 0 (0 MB) out: 0 (0 MB)
 0x00007f5f6a3f0439 in fair::mq::Device::RunWrapper() at /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/Source/fairmq/fairmq/Device.cxx:486 (discriminator 6) from /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/install/lib/libFairMQ.so.1.4
 0x00007f5f6a3f2389 in <unknown> from /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/install/lib/libFairMQ.so.1.4
 0x00007f5f69348e9d in boost::detail::function::void_function_obj_invoker1<std::function<void (fair::mq::State)>, void, fair::mq::State>::invoke(boost::detail::function::function_buffer&, fair::mq::State) at /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/install/include/boost/function/function_template.hpp:159 from /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/install/lib/libFairMQStateMachine.so.1.4
 0x00007f5f693516b9 in boost::signals2::detail::signal_impl<void (fair::mq::State), boost::signals2::optional_last_value<void>, int, std::less<int>, boost::function<void (fair::mq::State)>, boost::function<void (boost::signals2::connection const&, fair::mq::State)>, boost::signals2::mutex>::operator()(fair::mq::State) at /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/install/include/boost/optional/optional.hpp:348 from /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/install/lib/libFairMQStateMachine.so.1.4
 0x00007f5f69358358 in fair::mq::fsm::Machine_::ProcessWork() at /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/install/include/boost/signals2/detail/signal_template.hpp:726 from /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/install/lib/libFairMQStateMachine.so.1.4
 0x00007f5f69347668 in fair::mq::StateMachine::ProcessWork() at /usr/include/c++/8/bits/std_function.h:256 from /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/install/lib/libFairMQStateMachine.so.1.4
 0x00007f5f6a3fd6cd in fair::mq::DeviceRunner::Run() at /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/Source/fairmq/fairmq/PluginManager.h:86 from /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/install/lib/libFairMQ.so.1.4
 0x000000000041863b in <unknown> from /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/FairRoot_v18.8_patches/build/examples/MQ/serialization/ex-serialization1-processor
 0x00007f5f624c8d85 in __libc_start_main + 0xe5 from /lib64/libc.so.6
 0x00000000004186ae in <unknown> from /tmp/jenkins-FairRootGroup-FairSoft-PR-489-1.6UB/build/FairRoot_v18.8_patches/build/examples/MQ/serialization/ex-serialization1-processor
 ```
 
 and another suspicious line during data generation is
 
 ```
[#0] WARNING:InputArguments -- The parameter 'fSigmaT' with range [-1e+30, 1e+30] of the RooGaussian 'fGaussT' exceeds the safe range of (0, inf). Advise to limit its range.
 ```

So, without knowing the exact frequencies, from memory, I see one of serialization examples fail rather often, maybe 1 out of 5 legacy builds (but not on all OSes, however it does not seem to be OS-specific, I have seen it on all)

Within the FairRoot CDash data from just 2023 I found only one failed serialization test: https://cdash.gsi.de/testDetails.php?test=13255085&build=366583 which does not contain any error message, but just fails to terminate.

**Comment by rbx** [comment]
I've looked at https://cdash.gsi.de/testDetails.php?test=12944878&build=359802
there the serialization example times out. In the logs I see **some** transfers:
processor:
```
[ex-serialization2-processor][19:11:59.044850][INFO][Device.cxx:757:LogSocketRates] data1[0]: in: 20 (2.64208 MB) out: 0 (0 MB)
[ex-serialization2-processor][19:11:59.044896][INFO][Device.cxx:757:LogSocketRates] data2[0]: in: 0 (0 MB) out: 20 (2.31132 MB)
```
sink:
```
[ex-serialization2-sink][19:11:59.124947][INFO][Device.cxx:757:LogSocketRates] data2[0]: in: 20 (2.31132 MB) out: 0 (0 MB)
```
While sampler reports no transfers at all.

Do these tests perhaps run in parallel with the same tests from parallel builds? The ports in the example are hardcoded, so maybe some data is routed incorrectly. Like the processor here receiving data from a sampler from a different testsuite of another build.

That still does not explain the crashes, but it would explain the failure I linked above.

**Comment by rbx** [comment]
I'll switch the log severity in the example to debug, so we can see more connection info.

**Comment by rbx** [comment]
Possible fix would be to switch the example from TCP to IPC, then the "socket" file would be in the workdir of the build and will not conflict with parallel builds. Or even with the same execution of the test, if the socket filename is randomized.

**Comment by ChristianTackeGSI** [comment]
We currently run one testsuite per host. So unless different tests from FairRoot conflict with each other, this shouldn't be a problem?

**Comment by rbx** [comment]
While we still had two serialization tests (until couple of days ago when I changed them to one) they were running on the same ports. That must've been the issue.

I could also reproduce the timeout now locally by running two tests in parallel.

**Comment by rbx** [comment]
I added a commit, randomizing the connection string for the serialization test.

**Comment by dennisklein** [comment]
> I added a commit, randomizing the connection string for the serialization test.

Thx, looks good.

Just a thought triggered by this: A completementary approach to reduce such kind of conflicts could also be to start using `--net` and `--ipc` options of singularity. That wouldn't have helped for segmenting two ctests within the same suit, but it can help guarding against name conflicts with multiple parallel builds on the same host.

**Comment by rbx** [comment]
After the randomozation of the connection strings, I ran the same test in parallel.
And I frequently got this failure in the sampler:
```
19: Warning in <TFile::Init>: file /home/orybalch/dev/FairRoot/build/examples/MQ/serialization/data_io/testinput.root probably not closed, trying to recover
19: Info in <TFile::Recover>: /home/orybalch/dev/FairRoot/build/examples/MQ/serialization/data_io/testinput.root, recovered key TList:BranchList at address 360
19: Warning in <TFile::Init>: successfully recovered 1 keys
...
19: [ex-serialization-sampler][11:44:36.972442][INFO][Device.cxx:441:RunWrapper] fair::mq::Device running...
19: 
19:  *** Break *** segmentation violation
...
19: Thread 1 (Thread 0x7f0b999a50c0 (LWP 527698) "ex-serializatio"):
19: #0  0x00007f0b9b55449f in __GI___wait4 (pid=527757, stat_loc=stat_loc
19: entry=0x7ffd383a3a78, options=options
19: entry=0, usage=usage
19: entry=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
19: #1  0x00007f0b9b55441b in __GI___waitpid (pid=<optimized out>, stat_loc=stat_loc
19: entry=0x7ffd383a3a78, options=options
19: entry=0) at ./posix/waitpid.c:38
19: #2  0x00007f0b9b4babcb in do_system (line=<optimized out>) at ../sysdeps/posix/system.c:171
19: #3  0x00007f0b9bbf7974 in TUnixSystem::Exec (shellcmd=<optimized out>, this=0x558267b74800) at /home/orybalch/dev/FairSoftFull/build/Source/root/core/unix/src/TUnixSystem.cxx:2104
19: #4  TUnixSystem::StackTrace (this=0x558267b74800) at /home/orybalch/dev/FairSoftFull/build/Source/root/core/unix/src/TUnixSystem.cxx:2395
19: #5  0x00007f0b9bbf4d25 in TUnixSystem::DispatchSignals (this=0x558267b74800, sig=kSigSegmentationViolation) at /home/orybalch/dev/FairSoftFull/build/Source/root/core/unix/src/TUnixSystem.cxx:3615
19: #6  <signal handler called>
19: #7  Sampler::Run (this=0x558267c2af00) at /home/orybalch/dev/FairRoot/examples/MQ/serialization/sampler.cxx:60
19: #8  0x00007f0b9c8ff429 in fair::mq::Device::RunWrapper (this=0x558267c2af00) at /home/orybalch/dev/FairSoftFull/build/Source/fairmq/fairmq/Device.cxx:486
```

After randomizing also the filename of the input/output file I no longer see these types of crashes.

Not sure if it is related to the failures seen in https://github.com/FairRootGroup/FairRoot/issues/1286 and https://github.com/FairRootGroup/FairRoot/issues/1295 however.

**Comment by dennisklein** [comment]
> After randomizing also the filename of the input/output file I no longer see these types of crashes.

I suggest to randomize the working directory instead of individual files.

**Comment by rbx** [comment]
> > After randomizing also the filename of the input/output file I no longer see these types of crashes.
> 
> I suggest to randomize the working directory instead of individual files.

Do we have an example of how to do this somewhere?
I'm guessing:
```cmake
string(RANDOM LENGTH 8 TEST_SESSION1)
file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${TEST_SESSION1})
set_tests_properties(ex_MQ_serialization PROPERTIES
  TIMEOUT "30"
  PASS_REGULAR_EXPRESSION "Received 100 messages!"
  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${TEST_SESSION1}
)
```

rbx:
Nvm, that's only done once per cmake run, not per ttest execution...

dennisklein:
Just `cd` into some directory in the controller bash script? (And make sure the files are emitted to some location relative to the working dir)

dennisklein:
> > > After randomizing also the filename of the input/output file I no longer see these types of crashes.
> > 
> > 
> > I suggest to randomize the working directory instead of individual files.
> 
> Do we have an example of how to do this somewhere? I'm guessing:
> 
> ```cmake
> string(RANDOM LENGTH 8 TEST_SESSION1)
> file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${TEST_SESSION1})
> set_tests_properties(ex_MQ_serialization PROPERTIES
>   TIMEOUT "30"
>   PASS_REGULAR_EXPRESSION "Received 100 messages!"
>   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${TEST_SESSION1}
> )
> ```

That should be fine as well, then you can remove the randomized address again, just use ipc name relative to workdir. Or, if address needs to see it, use the same random input both for working dir and address.

rbx:
Updated with the randomized working dir.

ChristianTackeGSI (approved):
LGTM

PR #1344, opened 2023-02-08
format: set SpacesBeforeTrailingComments to 1

rbx:
Set clang-format's `SpacesBeforeTrailingComments` to `1`.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -35,7 +35,7 @@ IncludeBlocks: Regroup
 NamespaceIndentation: None
 PointerAlignment: Left
 SortIncludes: true
-SpacesBeforeTrailingComments: 3
+SpacesBeforeTrailingComments: 1
 Standard: c++17
 UseTab: Never
 ...


ChristianTackeGSI:
I finally have an idea, why I have a slight preference for the current style: Three spaces are a very good separator between code and comment. One space can occur anywhere in normal code.

That said, this is highly personal and in no way a veto. So if others like one space better, go ahead.

karabowi:
Like @ChristianTackeGSI, I also prefer larger space between code and comments. Should we organize some voting?

rbx:
When introducing the format style, the `3` got in there by accident, copied from somewhere else unintentionally.
I think it is useless and "good separator" argument I don't really understand :)
But if you like it that's enough of an argument to keep it. I also don't care much about this. So feel free to close this.

dennisklein:
> Should we organize some voting?

I am neutral on this :P

PR #1349, opened 2023-02-10
Tutorial3 cleanup

rbx:
- Added tests.
- Moved relevant files from `basemq` to the example directory. These have very limited reusability value and are more suitable as an example. I added these to the breaking changes. The only known user is some old code in Panda. Confirmed with Tobias that it is not actually needed and are OK to remove.
- Reduced number of launch scripts to 1 - others produce more noise than value. Also simplified the scripts.
- Removed example with msgpack. It was used for some tests long ago, but now adds more noise than value.
- Simplified structure - shortened class/file names, simpler template structures.
- Removed ack channel. The idea behind it was to make sure all data arrives, but this can be done simpler (as it is done in the serialization example test).

The example classes could use further modernization (namespaces, memory management for (Fair)ROOT objects), but I'll leave that for future work.


examples/advanced/Tutorial3/MQ/README.md
```diff
@@ -3,3 +3,27 @@
 This part of Tutorial3 demonstrates how to transport data of Tutorial3 (FairTestDetector) via FairMQ.
 
 - \subpage ex_tutorial3_mq_dataformat
+
+## Data Format {#ex_tutorial3_mq_dataformat}
```
ChristianTackeGSI, line 7: Please remove this line:
```
- \subpage ex_tutorial3_mq_dataformat
```

It links to a full file with that "tag". As this file is now contained in the same file as subsection, it's not needed any longer.

If there are not other references to ` {#ex_tutorial3_mq_dataformat}`, you can remove that as well.

examples/advanced/Tutorial3/MQ/fileSink/FileSink.h
```diff
@@ -76,12 +72,12 @@ class FairTestDetectorFileSink : public fair::mq::Device
     void InitTask() override;
 
   private:
-    std::unique_ptr<TClonesArray> fOutput;
+    TClonesArray* fOutput;
```
ChristianTackeGSI, line 74: Why are you turning this back from a unique_ptr to a raw pointer?

One could even think of using a value member instead?

```cpp
TClonesArray fOutput{FairTestDetectorHit::Class()};
```

The `Branch()` then should look like this:
```cpp
fTree.Branch("Output", &fOutput, 64000, 99);
```

rbx: in FileSinkTMessage.h there is `fTree.SetBranchAddress("Output", &fOutput);`, which wants to have a pointer to a pointer....
Apparently it was just broken before, so I "repare" it here.
I don't like the solution either.
But I can't get value member or unique_ptr to work directly... maybe i'm still missing something x)

ChristianTackeGSI: Hmmm, very strange.

I would have expected [this overload of SetBranchAddress](https://root.cern/doc/master/classTTree.html#a518a8303cac58607bca55672537742b4) to be used.

rbx: If I change it back to unique_ptr, and:
```
fTree.SetBranchAddress("Output", fOutput.get());
```
, which seems to be the overload you suggest,
Then I get this at runtime:
```
Error in <TTree::SetBranchAddress>: The address for "Output" should be the address of a pointer!

 *** Break *** segmentation violation
```

ChristianTackeGSI: I have the feeling, that this is a bug. I actually have an idea. I am going to post on the root forum and ask for some help in understanding their code.

Another question: Do we actually need the additional `tree.SetBranchAddress` when we already set the address in the `tree.Branch()` call?

rbx: > Another question: Do we actually need the additional `tree.SetBranchAddress` when we already set the address in the `tree.Branch()` call?

I don't know. Maybe @karabowi can give a hint?

ChristianTackeGSI: Okay, just for the reference, here's my root forum post where I got stuck reading their code about the idea I had: https://root-forum.cern.ch/t/please-explain-isptr-usage-in-ttree-setbranchaddress/53571

karabowi: The only solution I got after some lookaround is:
```
        auto array = fOutput.get();
        fTree.SetBranchAddress("Output", &array);
```
But that is no better than plain raw pointer fOutput, I guess.

dennisklein: The problem with `array` is that it is not clear how what lifetime TTree expects here. All this discussion here is years (maybe decades) old question, see https://github.com/FairRootGroup/FairRoot/issues/790. This broken TTree interface pops up again and again and it seems like people are "discovering" its awkwardness over and over -.-

rbx: > The only solution I got after some lookaround is:
> 
> ```
>         auto array = fOutput.get();
>         fTree.SetBranchAddress("Output", &array);
> ```
> 
> But that is no better than plain raw pointer fOutput, I guess.

Yeah, what happens when `array` goes out of scope? Does it need to stay alive or not? For how long? 

@dennisklein mentioned that Alice suggesting using a static pointer here. But then this is called in a loop, so it gets ugly quick.

ChristianTackeGSI: `array` has to be alive as long as TTree needs to access the TClonesArray! (Yes, it uses double pointer derefencing.)

A static pointer sounds like a global. I am definitely against a global in what ever variant, unless there is a clear need for it.

What about this one then?

```suggestion
    TClonesArray fOutputStorage{FairTestDetectorHit::Class()};
    TClonesArray* fOutput{&fOutputStorage};
```

No need for `new`/`delete`, no need to set fOutput anywhere else. (Disclaimer: Haven't tried it. Just an idea!)

rbx: Personally I would not go to such lengths to avoid new/delete. But if ya'll think that is the way I can do it like that.

ChristianTackeGSI: Well, it's your turn. If you don't like it. Keep the new / delete! It just was a suggestion, not a "You're required to do it that way to get a :+1:".

My *personal* thing isn't really about avoiding new/delete, but about avoiding boilerplate in constructors/destructors.

dennisklein: > No need for `new`/`delete`, no need to set fOutput anywhere else. (Disclaimer: Haven't tried it. Just an idea!)

Hmm, are we sure TTree will never create a new TCA? If it does, will it try to call delete on the existing one and even if it does not, we would then have two different TCAs. Otherwise, I am supporting value members for owned containers!

ChristianTackeGSI: We're 98% sure. See https://root.cern.ch/doc/v628/classTBranchElement.html#a7b19271c3bf833e6738bc18a43d6352f
* If `fOutput` would ever be a `nullptr`, then TTree/TBranch would do a new and put that in fOutput.

  > If addr is not zero and the pointer addr points at is also not zero, then the caller has allocated a branch object and is asking us to use it. The caller owns it and must delete it when it is no longer needed.

* We may NOT set "AutoDelete"
  > If AutoDelete is on (see [TBranch::SetAutoDelete](https://root.cern.ch/doc/master/classTBranch.html#a12047e8cff2e5e570bc0915ab48362df)), the top level objet will be deleted and recreate each time an entry is read, whether or not the [TTree](https://root.cern.ch/doc/master/classTTree.html) owns the object.

rbx: It is not obvious to me what the best solution for this issue will be, accounting both for `SetBranchAddress` and the serializers.

For this PR I prefer to leave it at raw ptr + new/delete, as it is also done in:
https://github.com/FairRootGroup/FairRoot/blob/dev/examples/MQ/Lmd/runMBSSink.cxx#L75
and
https://github.com/FairRootGroup/FairRoot/blob/dev/examples/MQ/serialization/sink.cxx#L83

The change from unique_ptr to raw ptr in this PR is necessary to get it to work at all - it was entirely broken before, which the added tests reveal.

ChristianTackeGSI: One thing I wondered: If the serializers usually allocate the TCA, do we actually need to `new` one in the constructor?

rbx: Looks like it is not needed to be in the ctor. Probably it doesn't have to be a member at all. But rather be created and destroyed for each message.

ChristianTackeGSI: Having it as a member is good. Because TBranch needs its address.

Well, you're right: Let's keep it for this PR and clean this up in another go.

ChristianTackeGSI (approved):
Okay, I wasn't able to follow each detail, but generally this starts to look good for me.

Some final things I noticed.

examples/advanced/Tutorial3/MQ/samplerTask/DigiLoaderBin.h
```diff
@@ -5,31 +5,28 @@
...
 
     for (int i = 0; i < numEntries; ++i) {
         FairTestDetectorDigi* digi = static_cast<FairTestDetectorDigi*>(fInput->At(i));
         if (!digi) {
             continue;
         }
-        new (&digiPayload[i]) TestDetectorPayload::Digi();
-        digiPayload[i] = TestDetectorPayload::Digi();
+        new (&digiPayload[i]) Payload::Digi();
+        digiPayload[i] = Payload::Digi();
```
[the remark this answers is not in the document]
rbx: Yeah, second line seems unnecessary. I'll remove it.

examples/advanced/Tutorial3/MQ/fileSink/FileSink.h
ChristianTackeGSI: Having it as a member is good. Because TBranch needs its address.

Well, you're right: Let's keep it for this PR and clean this up in another go.

karabowi (approved):

PR #1352, opened 2023-02-14
build: Introduce `/source/`

ChristianTackeGSI:
Instead of cluttering the main directory with all the public components of FairRoot, introduce a new `src/` that will host them.

Also each component (target) should have its dedicated directory there. No components inside other directories (like MCConfigurator was a subdirectory of fairtools).

---

Checklist:


karabowi:
I do not like the name of the directory. Mainly because of specific meaning of `src` directory in the `ROOT` realm.
I would prefer to reorganize the `fairtools` directory for things like this.
But I am open to discussions.

ChristianTackeGSI:
`source/`, `lib/`, `library/`?

The idea is to put all the "c++ code that provides the public API of FairRoot" there in a nice `foo/<COMPONENT>/` way. Where COMPONENT is anything like "mcconfigurator", "fairtools", "base", "parbase", "online", etc. And then we have nice structures there. Like `foo/COMPONENT/LinkDef.h` etc.

I don't care too much for the actual name of that directory.
* I would like to declutter the main directory (codemeta and friends will add at least two more files soonosh).
* I want to get rid of `COMPONENT1/something/COMPONENT2/` structures, like we have it for `base/sim/fastim/`. This is confusing.

karabowi:
I like `tools` but it's already kind of taken by fairtools...

ChristianTackeGSI:
I already said, I don't care much…

`tools/` sounds like directly usable things to me. Like a hammer.

Would you call FairDetector a tool that can be directly used?

rbx:
> I do not like the name of the directory. Mainly because of specific meaning of `src` directory in the `ROOT` realm.

What is the special meaning of `src`?

ChristianTackeGSI:
> Not really, but are you planning to have `FairDetector` in this directory as well?

Defintely yes. By moving the `base` COMPONENT from `base/` to `src/base/`. (But first moving fastsim out of base.)

I just started with mcconfigurator and datamatch because … no good reason at all.

rbx:
> In ROOT the `src` directories always contain source files, while `inc` directories contain header files. Quite annoying, but still, seeing `src` directory in any project I would expect it to contain only source files.

I see. Alice also separates source and include files and it is **horrible** for navigation in my opinion.

Keeping both headers and sources in `/src` and all the other meta stuff in `/` seems the best to me.

In FairMQ we have meta in `/` and sources+headers in `/fairmq`.
In ODC I also have meta in `/` and sources+headers in `/odc`.

But `/src` is just as good.

ChristianTackeGSI:
We even could call it `/fairroot/`. Not really my favourite, but acceptable.

Note: I would not move examples and templates there. Just all the remaining cxx/h.

dennisklein:
> In FairMQ we have meta in / and sources+headers in /fairmq.
> In ODC I also have meta in / and sources+headers in /odc.

There is actually semantics behind that, in both projects we install a potentially nested and prefixed include directory that follows this directory structure in the source tree. In CMake terms, FairMQ and ODC just have
```cmake
target_include_directories(${target}
    PUBLIC
    $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>
    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
  )
```
This enables that one can use the project's own headers always like this
```cpp
#include <fairmq/MyHeader.h>
```
no matter if its from a using code or from own code. Pretty consistent, that is why I chose it mainly.

In FairRoot we have a mess:
* in the source tree we scatter header files in a nested directory structure
* but in the install tree we collect them and put them all flat into `<prefix>/include/` (not even into `<prefix>/include/fairroot` - ok, we prefix all our installed headers with `Fair`, so no directory needed, right?! ... NOT)
* we `#include "SomeFairRootHeader.h"`, but the `#include ""` syntax means include this file from the same directory, but since headers are scattered everywhere, this usually does not work and falls back to the `#include <>` logic anyways, which means search the header in the configured search path (`-I` which we set in CMake via `target_include_directories`)

So, once our users consume FairRoot targets, we are free to clean the actual install tree if we wanted to.

> What is the special meaning of src?

I see the following semantics here:
Having a source tree top-level structure
```
/docs
/src/libA
/src/libB
/tests/testA
/tests/testB
/examples/exA
/examples/exB
LICENSE
README
```
e.g. having `examples` and `libA` on the same directory level feels off to me, because they are of a different type - single libdir vs container dir for examples.

The issue whether one has a seperate `/include` (either nested per lib or one global) kind of directory structure, I see as a separate concern from this.

But to avoid the confusion (although `src/` is quite a common choice for this), what about `lib` or `libs`? Then the source tree could cleanly be
```
/<type>/<instance-of-type>
/<metafiles> that have to stay on top-level
```

e.g. more or less:

```
/cmake  <-- this is common convention, but i could also see it in /scripts/cmake
/docs  <-- renamed from doxygen
/examples/<category>/<example>  <-- do we really need the category directory level?
/libs/<lib>  <-- all libs flat on this level, no nested stuff any more
/scripts  <-- move all the scripts from / here as well
/templates/<templates>
/tests  <-- since its basically empty / unused (and private anyways), lets rename it to plural for consistency?
/CMakeLists.txt
/Jenkinsfile
/AUTHORS
/CONTRIBUTORS
/README
/LICENSE
```

karabowi:
Now I understand what is behind this move. Looking around, I opt for the name `source`.
One of the reasons behind I mentioned. The other is, if `src` then why `tests` and not `tst`,
why `examples` and not `exe`?

ChristianTackeGSI:
Okay, renamed to `/source/`.

I also moved `*mq` (basemq, fairmq, parmq) into the new directory.

You can take a look at the result [here](https://github.com/ChristianTackeGSI/FairRoot/tree/pr/cmake/source).

karabowi (dismissed):

rbx:
> This enables that one can use the project's own headers always like this
> 
> ```c++
> #include <fairmq/MyHeader.h>
> ```
> 
> no matter if its from a using code or from own code. Pretty consistent, that is why I chose it mainly.

I think this is a really nice reason!
Then in FairRoot we could have, for example:
```
fairroot/datamatch/FairMCMatch.h
```
in code:
```c++
#include <fairroot/datamatch/FairMCMatch.h> // no matter if it is from FairRoot code or Panda/CBM/* code)
```
and once namespaces eventually are used:
```c++
#include <fairroot/datamatch/MCMatch.h>
```
with include guards such as
```c++
#ifndef FAIRROOT_DATAMATCH_MCMATCH
```

Although `fairroot` feels a little bit verbose in code to me. How about just `fair`:
```
fair/datamatch/MCMatch.h
```
```c++
#include <fair/datamatch/MCMatch.h>
```
with include guards such as
```c++
#ifndef FAIR_DATAMATCH_MCMATCH
#define FAIR_DATAMATCH_MCMATCH

namespace fair::datamatch {
class MCMatch : public TNamed {
```
There is some `fair::` namespace in FairRoot already: https://github.com/FairRootGroup/FairRoot/blob/master/basemq/policies/Serialization/BoostSerializer.h#L37-L39

ChristianTackeGSI:
I would suggest that we go step by step:

Let's first consolidate the build structure under `/source/`.
And then think again about improving the include-methodology.
Maybe open a midterm issue to keep track of discussions about refactoring the include-structure?

karabowi:
I've added the topic to [wekan](https://wekan.gsi.de/b/oLHPT3gvG75oKK3qA/sde-group-weekly-meeting).

rbx:
We agreed in the meeting to have the directory name be `fairroot`.

ChristianTackeGSI:
Okay, renamed.

dennisklein (commented):
```diff
- Move fairmq, basemq, and parmq also into `/fairroot/`.
+ Move fairmq, basemq, and parmq into `/fairroot/`.
```

*also* is confusing to me as it refers to some context which is not obvious to discover. Maybe, replace it with:

```diff
- Move fairmq, basemq, and parmq also into `/fairroot/`.
+ Move fairmq, basemq, and parmq into `/fairroot/`. See #1352 for more details.
```

karabowi (approved):

Issue #1355, opened 2023-02-20
Tutorial3 Sampler crashes when trying to remove a file


rbx:
https://cdash.gsi.de/testDetails.php?test=13503471&build=371764

After some investigation it looks like Sampler is trying to remove an intermediate file that it creates:
```
SysError in <TFile::TFile>: could not delete /shared/alfaci/jenkins/workspace/FairRootGroup_FairRoot_PR-1352@6/build/examples/advanced/Tutorial3/macro/data/testdigi_TGeant3.root.out.root (errno: 2) (No such file or directory)
[tut3-sampler][11:27:55.987090][FATAL][FairRootFileSink.cxx:66:FairRootFileSink] Error opening file /shared/alfaci/jenkins/workspace/FairRootGroup_FairRoot_PR-1352@6/build/examples/advanced/Tutorial3/macro/data/testdigi_TGeant3.root.out.root
```
Which is somewhere in:
https://github.com/FairRootGroup/FairRoot/blob/dev/examples/advanced/Tutorial3/MQ/sampler.cxx#L101-L103
And because the filename is hardcoded and this code is executed by different tests in parallel, they enter a race trying to create/delete this file.

Logically, the Sampler should only read an input file and send out data, it shouldn't have to create any intermediary files.
After consulting with @karabowi it seems that the reason for this output file is that `FairRunAna` needs to have a sink, otherwise it will also crash.

A quick/stupid fix would be to parameterize the intermediary filename with a session ID, then parallel tests will not crash each other when trying to recreate this file.

Better fix would be to make it possible for `FairRunAna` to work without a Sink.

dennisklein:
> Better fix would be to make it possible for FairRunAna to work without a Sink.

+1

> A quick/stupid fix would be to parameterize the intermediary filename with a session ID, then parallel tests will not crash each other when trying to recreate this file.

Perhaps less stupid, but certainly also less quick would be to incorporate the session ID in a working directory name instead of propagating it down to each individual file.

rbx:
> Perhaps less stupid, but certainly also less quick would be to incorporate the session ID in a working directory name instead of propagating it down to each individual file.

True, however the working directory for these tests is already parameterized. But FairRootFileSink (or some other FairRoot component) seems to follow its own rules when it comes to working directories and in this case it equals the directory of the input file (which is not parameterized, since the input file is reused for several tests).

ChristianTackeGSI:
[...]

PR #1357, opened 2023-02-21
Allow running without sink

karabowi:
Replaced `LOG(fatal)` with `LOG(error)`
in `FairRootManager::RegisterImpl()`.
Now running without `sink` will not crash.
Note that in this mode also persistent branches would not be stored anywhere.

Addresses issue #1355.

---

Checklist:

### DIFF
--- a/base/steer/FairRootManager.cxx
+++ b/base/steer/FairRootManager.cxx
@@ -148,8 +148,10 @@ Bool_t FairRootManager::InitSink()
 {
     if (fSink) {
         fSink->InitSink();
+        return true;
     }
-    return kTRUE;
+    LOG(info) << "The output sink is not set. No branches will be stored.";
+    return false;
 }
 
 template<typename T>
@@ -173,7 +175,7 @@ void FairRootManager::RegisterImpl(const char* name, const char* folderName, T* 
         if (fSink) {
             fSink->RegisterImpl(name, folderName, obj);
         } else {
-            LOG(fatal) << "The sink does not exist to store persistent branches.";
+            LOG(warning) << "The sink does not exist to store persistent branches (" << name << ").";
         }
     }
     AddMemoryBranch(name, obj);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@ In the MT mode of Geant4 the folder names changed from 'folderName_1' and 'folde
 ### Bug fixes
 * Check the return value of `source->InitUnpackers()`/`source->ReinitUnpackers()`
   in `FairRunOnline`. Stop run if `false` returned.
+* Remove sink from Tutorial3/MQ/sampler.cxx.
 
 ### Other Notable Changes
 * Allow running without output sink. In this case even persistent branches would not be stored anywhere.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,9 @@ In the MT mode of Geant4 the folder names changed from 'folderName_1' and 'folde
 * Check the return value of `source->InitUnpackers()`/`source->ReinitUnpackers()`
   in `FairRunOnline`. Stop run if `false` returned.
 
+### Other Notable Changes
+* Allow running without output sink. In this case even persistent branches would not be stored anywhere.
+
 ## 18.8.0 - 2022-12-16
 
 ### Breaking Changes
--- a/base/steer/FairRunAna.cxx
+++ b/base/steer/FairRunAna.cxx
@@ -213,7 +213,7 @@ void FairRunAna::Init()
         fRunId = GetEvtHeaderRunId();
 
         // Copy the Event Header Info to Output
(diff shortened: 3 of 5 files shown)


ChristianTackeGSI:
Could you add a "test" by changing sampler in Tutorial3 to not use a sink? That shoukd probably fix #1355?

karabowi:
Added new commit with removal of sink from Tutorial3/MQ/sampler.cxx

base/steer/FairRootManager.cxx
```diff
@@ -173,7 +173,7 @@ void FairRootManager::RegisterImpl(const char* name, const char* folderName, T*
         if (fSink) {
             fSink->RegisterImpl(name, folderName, obj);
         } else {
-            LOG(fatal) << "The sink does not exist to store persistent branches.";
+            LOG(error) << "The sink does not exist to store persistent branches.";
```
ChristianTackeGSI, line 176: Please excuse for noticing now.

The tutorial3 "test" helped!

```
[tut3-sampler][14:46:22.509645][ERROR][FairRootManager.cxx:176:RegisterImpl] The sink does not exist to store persistent branches.
```

Please consider outputting the `name` and/or `folderName` here as well (and that toFile is `true`)? That way people might be able to not call `Register` with `toFile=true`.

ChristianTackeGSI (approved):

rbx:
if it is a valid use case, then it should not be with your severity. debug or maybe warn.

ChristianTackeGSI:
> if it is a valid use case, then it should not be with error severity. debug or maybe warn.

:+1:

"warn" maybe. People can set `toFile=false`, if they don't have a sink. (We should take a look, where the Register in Tut3 is and fix it.)

karabowi:
This Register comes for sure from the EventHeader branch.

karabowi:
Changed the main commit message to:

Allow running without sink

Replaced `LOG(fatal)` with `LOG(debug)` in `FairRootManager::RegisterImpl()`.
Note that in this mode also persistent branches would not be stored anywhere.
Changed `FairRootManager::InitSink()` to print a `warning` and `return false`.

karabowi:
I still believe a warning should be printed, as in most cases running without a `sink` is user's mistake rather than user's chosen option.

ChristianTackeGSI:
> I still believe a warning should be printed, as in most cases running without a `sink` is user's mistake rather than user's chosen option.

Yes, warning, please. We should warn users, if they do something non-sensible.

And in relevant `Register` calls we should check for sink being set.

base/event/FairEventHeader.cxx
```diff
@@ -16,7 +16,8 @@
 
 void FairEventHeader::Register(Bool_t Persistence)
 {
-    FairRootManager::Instance()->Register("EventHeader.", "EvtHeader", this, Persistence);
+    FairRootManager::Instance()->Register(
+        "EventHeader.", "EvtHeader", this, (FairRootManager::Instance()->GetSink() ? Persistence : false));
```
ChristianTackeGSI, line 20: I am not so happy about FairRootManager::Instance being used more. We possibly should look at the callers and implement the check there.

That said, I am okay with this for now.

base/event/FairMCEventHeader.cxx
```diff
@@ -82,7 +82,8 @@ void FairMCEventHeader::Reset()
 void FairMCEventHeader::Register()
 {
     // Dot at the end of the name is needed for splitting!!
-    FairRootManager::Instance()->Register("MCEventHeader.", "Event", this, kTRUE);
+    FairRootManager::Instance()->Register(
+        "MCEventHeader.", "Event", this, (FairRootManager::Instance()->GetSink() ? true : false));
```
ChristianTackeGSI, line 86: clang-tidy

> redundant boolean literal in ternary expression result

```suggestion
    FairRootManager::Instance()->Register(
        "MCEventHeader.", "Event", this, FairRootManager::Instance()->GetSink());
```

ChristianTackeGSI (commented):
Please fix the clang-tidy issue.

karabowi:
Print `info` in the initialization, if sink not present.
Print `warning` if registering persistent branch without sink.
Register `(MC)EventHeader` only if sink present.

base/event/FairEventHeader.cxx
karabowi: So would you prefer to apply the check `GetSink` in the places where we actually `Register` the branches, i.e. in `FairRunAna*` and `FairMCApplication`?

ChristianTackeGSI: Yes, I would prefer that. It would make things more straight IMHO.

And for FairMCApplication (that's FairRunSim, right?) I would then ask: Is running a simulation without a sink a supported use case? If not, then not checking the sink and letting RegisterImpl throw a warning would actually be fine (as we would warn the user about a not really supported use case).

ChristianTackeGSI (approved):

PR #1358, opened 2023-02-21
Fix memory leak in PixelDigi*Source

ChristianTackeGSI:
Use valued members instead of heap objects.

FairRootManager::GetObject uses FairSource::ActivateObject as a backend. Let's start and fix the memory handling there.

---

Checklist:

### DIFF
--- a/examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
+++ b/examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
@@ -15,21 +15,16 @@
 //
 #include "PixelDigiBinSource.h"
 
-#include "FairLogger.h"
 #include "FairRootManager.h"
 #include "PixelDigi.h"
-#include "PixelEventHeader.h"
 
-#include <TClonesArray.h>
-#include <TObject.h>
-#include <TString.h>
 #include <cstring>
+#include <fairlogger/Logger.h>
 #include <string>
 
 PixelDigiBinSource::PixelDigiBinSource(TString inputFileName)
     : FairSource()
-    , fEventHeader(nullptr)
-    , fDigis(nullptr)
+    , fDigis(PixelDigi::Class(), 10000)
     , fNDigis(0)
     , fTNofEvents(0)
     , fTNofDigis(0)
@@ -43,8 +38,6 @@ PixelDigiBinSource::PixelDigiBinSource(TString inputFileName)
     LOG(debug) << "PixelDigiBinSource created------------";
 }
 
-PixelDigiBinSource::~PixelDigiBinSource() {}
-
 Bool_t PixelDigiBinSource::Init()
 {
     // Get input array
@@ -55,12 +48,10 @@ Bool_t PixelDigiBinSource::Init()
         LOG(fatal) << "No FairRootManager";
 
     // Register output array StsDigi
-    fDigis = new TClonesArray("PixelDigi", 10000);
-    ioman->Register("PixelDigis", "Pixel", fDigis, kFALSE);
+    ioman->Register("PixelDigis", "Pixel", &fDigis, kFALSE);
 
-    fEventHeader = new PixelEventHeader();
-    fEventHeader->SetName("EventHeader.");
-    ioman->Register("EventHeader.", "EvtHeader", fEventHeader, kFALSE);
+    fEventHeader.SetName("EventHeader.");
+    ioman->Register("EventHeader.", "EvtHeader", &fEventHeader, kFALSE);
 
     fInputFile.open(fInputFileName.Data(), std::fstream::in | std::fstream::binary);
 
@@ -74,7 +65,7 @@ Bool_t PixelDigiBinSource::Init()
 
 Int_t PixelDigiBinSource::ReadEvent(UInt_t i)
 {
-    fDigis->Clear();
+    fDigis.Clear();
     fNDigis = 0;
 
     if (!fInputFile) {
@@ -89,7 +80,7 @@ Int_t PixelDigiBinSource::ReadEvent(UInt_t i)
     fCurrentEntryNo = i;
 
     std::string buffer;
-    LOG(debug) << "PixelDigiBinSource::ReadEvent() Begin of (" << fDigis->GetEntries() << ")";
+    LOG(debug) << "PixelDigiBinSource::ReadEvent() Begin of (" << fDigis.GetEntries() << ")";
 
     Int_t head[4];   // runId, MCEntryNo, PartNo, NofDigis
     fInputFile.read((char*)head, sizeof(head));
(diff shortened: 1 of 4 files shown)


examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
```diff
@@ -108,22 +99,22 @@ Int_t PixelDigiBinSource::ReadEvent(UInt_t i)
...
                    << " 0.";
-        new ((*fDigis)[fNDigis]) PixelDigi(-1,
-                                           (Int_t)dataCont[idata * dataSize + 0],
-                                           (Int_t)dataCont[idata * dataSize + 1],
-                                           (Int_t)dataCont[idata * dataSize + 2],
-                                           (Int_t)dataCont[idata * dataSize + 3],
-                                           0.,
-                                           0.);
+        new (fDigis[fNDigis]) PixelDigi(-1,
+                                        (Int_t)dataCont[idata * dataSize + 0],
```
fuhlig1, line 112: I would propose C++ cast instead of C-style casts when the code is cleaned up anyway.

examples/MQ/pixelDetector/src/PixelDigiBinSource.h
```diff
@@ -62,8 +59,8 @@ class PixelDigiBinSource : public FairSource
     Bool_t ActivateObject(TObject** obj, const char* BrName) override;
 
   private:
-    PixelEventHeader* fEventHeader;
-    TClonesArray* fDigis; /** Output array of PixelDigi **/
+    PixelEventHeader fEventHeader{};
+    TClonesArray fDigis; /** Output array of PixelDigi **/
```
fuhlig1, line 63: Are you sure that this will work in cases where the TClonesArray needs to be resized?

ChristianTackeGSI: https://root.cern/doc/master/classTClonesArray.html#a2bb70b2ac885b9120dcc304fde2cfd50

> The second argument, s, indicates an approximate number of objects that will be entered in the array. If more than s objects are entered, the array will be automatically expanded.

How does a `new`/`delete`d object change that?

examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
ChristianTackeGSI: Thanks for noticing!

I don't think, a cast is needed at all. This is just a short int, that becomes an int. That should not even envolve any losses.

fuhlig1: Ok. I did not check the data types but upcasting should be fine. Downcasting is a problem and would end in a compiler error.
So the solution would be to simply remove the casts.

Is there a clang-tidy option to warn about C-style casts?

examples/MQ/pixelDetector/src/PixelDigiBinSource.h
fuhlig1: If ROOT can enlarge the TClonesArray in place there is probably no problem. Bu in general case ROOT needs to create anew and larger TClonesArray somewhere in memory, copy the old data, change the pointer to the new location and deletes the old TClonesArray.

dennisklein: Hm, but why would it ever need to create a new TCA? It would need to point the [`TClonesArray::fCont` member](https://root.cern/doc/master/classTObjArray.html#a1f358720e4cf05c1031d48e20824729e) to the new storage location and perhaps update `TClonesArray::fLast/fLowerBound`, would it not?

dennisklein: and second question: how would ROOT in general know what to do with the old TCAs? Does ROOT own all TCAs then?

ChristianTackeGSI: The TCA object is always 80 bytes in size (on x86-64), no matter how many elements are in there. The element storage is somewhere else. So it shouldn't make a difference, where the TCA lives.

fuhlig1: Right. I did not check the code properly. You are right, ROOT is handling the memory itself. So forget my comment.
It look like the TCA is implemented in the same way as e.g. the vector.

examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
ChristianTackeGSI: I removed the casts.

> Is there a clang-tidy option to warn about C-style casts?

Yes, there is: [cppcoreguidelines-pro-type-cstyle-cast](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-cstyle-cast.html)

But enabling it would spam us with hundreds of warnings. Nobody will fix them in the near future. If they're in headers, then they will be even more often.

Our current clang-tidy setup has issues with too many warnings (it misses sometimes some, sometimes even strange new ones appear). This is basically, because we compile in parallel and the warnings interleaf.

fuhlig1: I wouldn't switch them on by default. But maybe one should run clang-tidy before the final pull request.

examples/MQ/pixelDetector/src/PixelDigiSource.cxx
```diff
@@ -134,10 +126,10 @@ Int_t PixelDigiSource::ReadEvent(UInt_t i)
...
 Bool_t PixelDigiSource::ActivateObject(TObject** obj, const char* BrName)
 {
-    if (strcmp(BrName, "PixelDigis"))
-        *obj = (TObject*)fDigis;
-    else if (strcmp(BrName, "EventHeader."))
-        *obj = (TObject*)fEventHeader;
+    if (strcmp(BrName, "PixelDigis") == 0)
+        *obj = &fDigis;
+    else if (strcmp(BrName, "EventHeader.") == 0)
+        *obj = &fEventHeader;
```
ChristianTackeGSI, line 132: The previous `strcmp` usage here looks like a bug to me! I wonder, why this worked at all before? Or is it not called at all?

karabowi: Probably a bug, we are not testing this anywhere.

ChristianTackeGSI: Just a wild guess:

The `Register` that happens in ::Init seeds FairRootManager enough, so that GetObject will not try to Activate?
https://github.com/FairRootGroup/FairRoot/blob/cfdc0e708247bf545f904f7b0c510e693bcaf978/examples/MQ/pixelDetector/src/PixelDigiSource.cxx#L59

karabowi (approved):

PR #1363, opened 2023-02-25
Move ActivateObjectAnyImpl to FairFileSourceBase

ChristianTackeGSI:
ActivateObjectAnyImpl is really specific to TTree, TFile and friends.  Move it into FairFileSourceBase and remove the template part.

---

Checklist:

### DIFF
--- a/base/source/FairFileSourceBase.cxx
+++ b/base/source/FairFileSourceBase.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- *    Copyright (C) 2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ * Copyright (C) 2022-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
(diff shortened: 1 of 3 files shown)


karabowi (approved):

YanzhaoW:
Hi,

Could we add the `ActivateObjectAnyImpl` back to `FairSource.h`? We have our own implementation of [FileSource](https://github.com/R3BRootGroup/R3BRoot/blob/dev/r3bbase/R3BFileSource2.h), which doesn't need `FairFileSourceBase` at all. This also introduces a breaking change for the code based on v18.8.0.

ChristianTackeGSI:
Hi @YanzhaoW,

`FairFileSourceBase::ActivateObjectAnyImpl` is a static method, so you could `#include <FairFileSourceBase.h> and use it from anywhere. JUST: It's currently protected.

I would like to keep `FairSource.h` as generic as possible (if this is possible at all it another story, I don't know!).

Idea: Let's make it public, so it's usable from everywhere. I am happy with a PR for this. Please try to include some Doxygen comments! We can even backport this to 18.8. :-)

PR #1384, opened 2023-03-26
fix multiple leaks in FairMCApplication

ChristianTackeGSI:
Turned multiple member variables from raw pointers to unique_ptrs so that the allocated objects get destructed at destruction of FMCA.

Allocation of FairTrajFilter also moved from FairRunSim to FMCA and that way avoids the ::Instance methods.

Also turned one member variable into a valued member to simplify some code.

---

Checklist:

### DIFF
--- a/fairroot/base/sim/FairMCApplication.cxx
+++ b/fairroot/base/sim/FairMCApplication.cxx
@@ -55,24 +55,21 @@
 #include <TObject.h>            // for TObject
 #include <TParticlePDG.h>       // for TParticlePDG
 #include <TROOT.h>              // for TROOT, gROOT
-#include <TRefArray.h>          // for TRefArray
 #include <TSystem.h>            // for TSystem, gSystem
 #include <TVirtualMC.h>         // for TVirtualMC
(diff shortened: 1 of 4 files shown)


fairroot/base/sim/FairMCApplication.cxx
```diff
@@ -55,24 +55,21 @@
 #include <TObject.h>            // for TObject
 #include <TParticlePDG.h>       // for TParticlePDG
 #include <TROOT.h>              // for TROOT, gROOT
-#include <TRefArray.h>          // for TRefArray
 #include <TSystem.h>            // for TSystem, gSystem
 #include <TVirtualMC.h>         // for TVirtualMC
 #include <TVirtualMCStack.h>    // for TVirtualMCStack
-class TParticle;
+#include <float.h>              // for DBL_MAX
```
fuhlig1, line 61: Shouldn't we use the C++ headers <cfloat> and <cstdlib> instead of the ones from C?

ChristianTackeGSI: Thanks for finding this one!

Fixed.

karabowi (approved):

PR #1385, opened 2023-03-27
Geant4MT: Fix closing sink in MT mode

karabowi:
In the MT mode FairRunSim cannot close the sinks,
only FairMCApplication can.

---

Checklist:

### DIFF
--- a/fairroot/base/sim/FairMCApplication.cxx
+++ b/fairroot/base/sim/FairMCApplication.cxx
@@ -509,6 +509,10 @@ void FairMCApplication::FinishRunOnWorker()
     LOG(debug) << "FairMCApplication::FinishRunOnWorker: ";
 
     FinishRun();
+
+    // Close the sink on workers so that the execution of main FairRunSim->Run()
+    // finishes with the output files properly closed.
+    fRootManager->CloseSink();
 }
 
 //_____________________________________________________________________________


fairroot/base/sim/FairMCApplication.cxx
```diff
@@ -391,6 +391,7 @@ void FairMCApplication::FinishRun()
 
     if (!fRadGridMan && fRootManager) {
         fRootManager->Write();
+        fRootManager->CloseSink();
     }
```
dennisklein, line 395: hm, so `!fRadGridMan && fRootManager` means MT mode?

how does it relate to `fMC->IsMT()`?

ChristianTackeGSI:
What I also don't understand: In MT mode we have a "cloned" FairMCApplication that owns a "cloned" FairRunSim, and that owns the Sink. So once the simulation ends, this chain should get destructed. So the destructor of the FairSink should close it (I think, it does).

So the Sink needs to be closed earlier?

karabowi:
@ChristianTackeGSI , you are right, moved the CloseSink to FinishRunOnWorker.

ChristianTackeGSI (approved):
> In the MT mode FairRunSim cannot close the sinks,
only FairMCApplication can.

I still don't understand, why the destructor chain of FairMCApplication (on the worker) → FairRunSim → FairRun → FairSink (via FairRun::fSink) can't handle this?

karabowi:
> > In the MT mode FairRunSim cannot close the sinks,
> > only FairMCApplication can.
> 
> I still don't understand, why the destructor chain of FairMCApplication (on the worker) → FairRunSim (via fWorkerRunSim) → FairRun → FairSink (via FairRun::fSink) can't handle this?

Yes, it works, however (*).

To describe the problem, in the current `dev` the `ex_sim_tutorial1_Geant4MT` ends dirty with: 
```
MT mode, reading two files: t1 and t2
Warning in <TFile::Init>: file .//tutorial1_TGeant4_pions.mc_p2.000000_t0.000000_n20_t1.root probably not closed, trying to recover
Info in <TFile::Recover>: .//tutorial1_TGeant4_pions.mc_p2.000000_t0.000000_n20_t1.root, recovered key TFolder:cbmroot at address 318
Info in <TFile::Recover>: .//tutorial1_TGeant4_pions.mc_p2.000000_t0.000000_n20_t1.root, recovered key TList:BranchList at address 676
Info in <TFile::Recover>: .//tutorial1_TGeant4_pions.mc_p2.000000_t0.000000_n20_t1.root, recovered key TList:TimeBasedBranchList at address 885
Info in <TFile::Recover>: .//tutorial1_TGeant4_pions.mc_p2.000000_t0.000000_n20_t1.root, recovered key TTree:cbmsim at address 977
Warning in <TFile::Init>: successfully recovered 4 keys
Output chain has 20 events, 34 tracks, 40 points in 3 branches, 46 leaves.
Simulation successful.
```

When the `CloseSink` is added in the `FairMCApplication`, the warnings disappear.

Alternatively, I could fix it in the macro with `run.reset();` which closes the output files, and the warnings disappear.

(*) However, I have also spotted one other artifact. 

When I run the macro in current `dev` and I try to stop it with `ctrl+c`:
a/ in the single threaded mode, the execution stops and finishes gracefully,
b/ in the MT mode, nothing happens, the execution continues till the end.

I would like the execution to stop also in the MT mode.
The solution `run.reset();` in the macro does not stop execution when `ctrl+c`.
The line `fRootManager->CloseSink();` in the `FairMCApplication` allows the user
to stop the execution with `ctrl+c`. I know it is not related, just describing the behavior.

karabowi:
OK, had tested this `ctrl+c` more often. It looks quite random.

karabowi:
Changed the commit with the comment:

```
Geant4MT: Fix Geant4MT macro

Close the run before checking ouptut.
```

examples/simulation/Tutorial1/macros/run_tutorial1.C
```diff
@@ -160,8 +160,11 @@ int run_tutorial1(int nEvents = 10, string mcEngine = "TGeant3", bool isMT = tru
 
     // ------------------------------------------------------------------------
 
+    auto wasMT = run->WasMT();
+    run.reset();
```
ChristianTackeGSI, line 164: So good to see unique_ptr working for you!

```suggestion
    auto wasMT = run->WasMT();

    // Destruct and de-allocate the FairRunSim.
    run.reset();
```

I was actually looking for a `reset` method on FairRunSim and then realized that `run` is a unique_ptr… so maybe put a comment here?

ChristianTackeGSI (approved):
So if I understand that all correctly by now:

Not destructing the main FairRunSim (which your commit now fixes for this example) does not destruct the worker-FairMCApplication either? So even though the "Run" is over, the thread is either still alive, or at least the management stuff of that thread (the worker-RunSim and the worker-FairMCApplication) are still alife.

As we can't quickly fix that (it's geant4-vmc's job to destruct worker-FMCA (which would destruct worker-RunSim)), I think, your first way of fixing this issue has its own appeal:

It can be seen as a good style that after a `FairRunSim::Run()` all files are in a good state. So either I would suggest to put back your CloseSink thing (with a comment and/or commit message explaining that the worker-RunSim/FMCA live longer then expected), or maybe we should take a look at [TFile::Flush()](https://root.cern/doc/master/classTFile.html#ab8f4a2c2b79f89e3255d2310a53a75b5) and use that with nearly the same explanation. But closing now sounds quite reasonable to me. I think, we have some CloseSink in the main thread also.

P.S.: I guess, the Ctrl-C thing is now really unrelated?

ChristianTackeGSI (approved):

PR #1386, opened 2023-03-28
New way to setup Event Display by XML files.

DanielWielanek:
Hi,
again I'm trying to use XML to remove configuration of tracks/detectors from FairRoot Event Display

Mention any issue this PR resolves or is related to.

---

Checklist:

### DIFF
--- a/fairroot/eventdisplay/xml/FairXMLPdgColor.h
+++ b/fairroot/eventdisplay/xml/FairXMLPdgColor.h
@@ -1,0 +1,37 @@
+/********************************************************************************
+ * Copyright (C) 2020-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ *                                                                              *
+ *              This software is distributed under the terms of the             *
+ *              GNU Lesser General Public Licence (LGPL) version 3,             *
+ *                  copied verbatim in the file "LICENSE"                       *
+ ********************************************************************************/
+/*
+ * FairXMLPdgColor.h
+ *
+ *  Created on: 21 paź 2021
+ *      Author: Daniel Wielanek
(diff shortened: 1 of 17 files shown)


fairroot/eventdisplay/xml/FairXMLPdgColor.h
```diff
@@ -0,0 +1,37 @@
...
+
+class FairXMLPdgColor : public FairXMLEveConf
+{
+    std::map<int, int> fPDGToColor;
+    void SetDefColor();
+
+  public:
+    FairXMLPdgColor(FairXMLNode* node = nullptr);
+    Int_t GetColor(Int_t pdg) const;
+    ~FairXMLPdgColor() override;
```
ChristianTackeGSI, line 33:
```suggestion
    explicit FairXMLPdgColor(FairXMLNode* node = nullptr);
    Int_t GetColor(Int_t pdg) const;
    ~FairXMLPdgColor() override = default;
```

We haven't official adopted [CPPCG::C.46](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-explicit), but I think, it would be a good idea.

I think, the destructor is completely empty in the .cxx, so let's go for a `= default` one here, maybe?

ChristianTackeGSI (commented):
Hi!

As I already noted, this changeset is too big for me, to really review it quickly. Some ideas to split this up into digestable pieces:

* `FairEveTracks.*` looks like a nice, general cleanup. Easy for a dedicated PR.
* FairXMLPdgColor and its base class also look quite digestable. It could be used to perform a temporary refactoring of FairEventManager in a non-API breaking manor?

Issue #1243, opened 2022-10-20
Zero size bounding box for some TGeoVolumeAssembly nodes after strong alignment


PALoizeau:
**Describe the bug**
After alignment is applied, the BoundingBox of TGeoVolumeAssembly objects may end up with 0 size (GetDX, GetDY and GetDZ returning 0) if the offset exceed the original boundaries. This is not expected as it means there are other effects on the Volume aside from the translation/rotation.
We (me and @fuhlig1) suspect that this may partially be recursive (affecting the parent or daughter of the aligned volume)

**To Reproduce**
Steps to reproduce the behavior with the attached files (includes example both for problem and for macro-level fix):
1. Compile a copy of FairSoft apr21p2 and FairRoot v18.6.7 (would probably also work with any older versions of each and with recent version of FairRoot)
2. Download [BboxAlignBug.zip](https://github.com/FairRootGroup/FairRoot/files/9830426/BboxAlignBug.zip) and unzip the macro and the two root files in a folder
3. Go to this folder, load your FairSoft-FairRoot environment
4. Run 
    ```
    root -l -b -q 'BboxAlignBug.C("mcbm_beam_2022_05_23_nickel.geo.root", "AlignmentMatrices_mcbm_beam_2022_05_23_nickel.root")'
    ```

**Expected behavior**
After alignment of the geometry, all volumes/nodes are only offset/rotated without changes to their other properties.
This can be cured by forcing the recomputation of the BoundingBox of the affected Volume.

A temp fix for now (until this is hopefully fixed in ROOT) is to simply  do the recompute for all TGeoVolumeAssembly in the geometry tree, which seems to be fast enough that we do not have to try to track affected Volumes to be more selective.

**System information:**
 - OS: Debian 8, Debian 10, Ubuntu 20.04
 - Compiler: GCC > 8
 - Environment: FairSoft apr21 (and patch releases)

**Additional context**
- Corresponding pull request with temp fix on FairRoot side coming soon (waiting until issue done to have proper linking)
- Probably of interest to @TobiasStockmanns and @RalfKliemt  to check that our proposed fix does not break something for PandaRoot

PALoizeau:
I switch the PR back to draft mode for the following reasons:
1. @fuhlig1 has some doubt about the universal loop, as it creates a physical node for all AssemblyVolumes, including those who did not have one to start, which will increase the memory
[...]

PR #1389, opened 2023-04-06
GeoAssembly bounding box after alignment: use ROOT expert patch

PALoizeau:
**Related to:**
- cf discussion in https://github.com/root-project/root/issues/12242
- Improvement of PR #1244
- see also Issue #1243
- Solves new sub-case of Cbmroot issue https://redmine.cbm.gsi.de/issues/2620 (behind CBM login due to GSI policy)

**Summary:**
Replace original fix with more generic solution also catching edge case where only Nodes inside assembly are aligned but not assembly itself.

**Problem description** 
(valid for original patch in Fairroot v18.8.0 and v18.8.1):
- previous re-computation patch is only applied to Assemblies which are/have a PhysicalNode (first check if PhysicalNode, then if Assembly)
- by default, Assemblies do not have/are not PhysicalNodes (not expert on Geometries myself, so maybe bad description)
- the alignment procedure forces a PhysicalNode creation/conversion but only on the nodes being aligned

=> In case the user provides alignment only for leaves/daughters of the Assembly for which the BoundingBox is used, the later was never recomputed. They fixed it by introducing a Unity matrix alignment but this potentially lead to confusion over what was moved during the alignment.
=> @agheata patch handles both the original case and this new one without any penalty from what I could test (and anyway this all happens at Init)

**Proposed backports**:
- Patch release of the `v18.8.x` family

---

Checklist:

### DIFF
--- a/fairroot/alignment/FairAlignmentHandler.cxx
+++ b/fairroot/alignment/FairAlignmentHandler.cxx
@@ -4,6 +4,7 @@
 #include <TGeoManager.h>
 #include <TGeoPhysicalNode.h>
 #include <TGeoShapeAssembly.h>
+#include <TGeoVoxelFinder.h>
 #include <fairlogger/Logger.h>
 
 void FairAlignmentHandler::AlignGeometry() const
@@ -18,8 +19,9 @@ void FairAlignmentHandler::AlignGeometry() const
             AlignGeometryByFullPath();
         }
 
-        // --- Force BoundingBox recomputation for AssemblyVolumes as they may have been corrupted by alignment
-        // FIXME: will hopefully be fixed in Root in near future, temp fix in meantime
+        // --- Force BoundingBox recomputation for volumes/nodes needing it after alignment
+        // FIXME: May be added to RefreshPhysicalNode step in Root in near future, temp fix needed here in meantime
+        ///       cf https://github.com/root-project/root/issues/12242
         RecomputePhysicalAssmbBbox();
 
         LOG(info) << "Refreshing geometry...";
@@ -100,15 +102,23 @@ void FairAlignmentHandler::AddAlignmentMatrices(const std::map<std::string, TGeo
 
 void FairAlignmentHandler::RecomputePhysicalAssmbBbox() const
 {
-    TObjArray* pPhysNodesArr = gGeoManager->GetListOfPhysicalNodes();
-
-    for (auto pPhysNode : TRangeDynCast<TGeoPhysicalNode>(pPhysNodesArr)) {
-        if (pPhysNode) {
-            auto pShapeAsb = dynamic_cast<TGeoShapeAssembly*>(pPhysNode->GetShape());
-            if (pShapeAsb) {
-                // Should reach here only if the original node was a TGeoShapeAssembly
-                pShapeAsb->ComputeBBox();
-            }
+    /// After alignment, forces the recomputation of the bounding box of all volumesAssemblies and the rebuilding of
+    /// Nodes requiring it.
+    /// => Not done after each change to geometry in ROOT for perf reason
+    /// => to be called once all changes done
+    /// => original snippet proposed in https://github.com/root-project/root/issues/12242 by A. Gheata
+    TObjArray* volumes = gGeoManager->GetListOfVolumes();
+    for (auto vol : TRangeDynCast<TGeoVolume>(volumes)) {
+        if (!vol) {
+            continue;
+        }
+        if (vol->IsAssembly()) {
+            vol->GetShape()->ComputeBBox();
+        }
+        auto finder = vol->GetVoxels();
+        if (finder && finder->NeedRebuild()) {
+            finder->Voxelize();
+            vol->FindOverlaps();
         }
     }
 }


fairroot/alignment/FairAlignmentHandler.cxx
```diff
@@ -100,15 +101,15 @@ void FairAlignmentHandler::AddAlignmentMatrices(const std::map<std::string, TGeo
...
-    for (auto pPhysNode : TRangeDynCast<TGeoPhysicalNode>(pPhysNodesArr)) {
-        if (pPhysNode) {
-            auto pShapeAsb = dynamic_cast<TGeoShapeAssembly*>(pPhysNode->GetShape());
-            if (pShapeAsb) {
-                // Should reach here only if the original node was a TGeoShapeAssembly
-                pShapeAsb->ComputeBBox();
-            }
+    TIter nextv(gGeoManager->GetListOfVolumes());
+    TGeoVolume* vol;
+    while ((vol = dynamic_cast<TGeoVolume*>(nextv()))) {
```
ChristianTackeGSI, line 106: Please use [TRangeDynCast](https://root.cern.ch/doc/v628/TCollection_8h.html#ab36279cd943b06d94ecec8a2a10110f7).

```suggestion
    TObjArray* volumes = gGeoManager->GetListOfVolumes();

    for (auto vol : TRangeDynCast<TGeoVolume>(volumes)) {
        if (!vol) {
            continue;
        }
```

fairroot/alignment/FairAlignmentHandler.cxx
```diff
@@ -100,15 +101,15 @@ void FairAlignmentHandler::AddAlignmentMatrices(const std::map<std::string, TGeo
...
+        if (vol->IsAssembly())
+            vol->GetShape()->ComputeBBox();
```
ChristianTackeGSI, line 108: I would prefer braces even around single statements. But that's really personal by now.

```suggestion
        if (vol->IsAssembly()) {
            vol->GetShape()->ComputeBBox();
        }
```

ChristianTackeGSI (commented):
I can't (yet) judge on the real stuff behind this, but I can add some stylistic things.

BTW: If you have an ORCID, please consider adding it to the `CONTRIBUTORS` file (take a look at `AUTHORS` for the format).

fairroot/alignment/FairAlignmentHandler.cxx
```diff
@@ -100,15 +101,15 @@ void FairAlignmentHandler::AddAlignmentMatrices(const std::map<std::string, TGeo
 
 void FairAlignmentHandler::RecomputePhysicalAssmbBbox() const
 {
```
ChristianTackeGSI, line 103: Could you put some sort of comment here, so that we know in the future, why this all is happening?
* Consider linking to the ROOT issue, so that we can check in the future, if that issue has been resolved and in which version of ROOT.

fairroot/alignment/FairAlignmentHandler.cxx
PALoizeau: Done, I also prefer it so and just applied too quickly the proposal from the CI test.

fairroot/alignment/FairAlignmentHandler.cxx
PALoizeau: Done

fairroot/alignment/FairAlignmentHandler.cxx
PALoizeau: I added both some descriptive information in the method itself and a reference to the ROOT issue in the existing `FIXME` to make it clear that this may be included on the ROOT side in the `Refresh` which we are anyway calling two line later.
Just let me know in case I should also add a line on why the Assemblies need a recompute at all.

ChristianTackeGSI: I asked about this, as we will likely keep these fixes for somethign like three years or more (assuming ROOT fixes it just now), because we will support "older" ROOT releases for a while.

@karabowi @fuhlig1: Are the current comments good enough for you?

ChristianTackeGSI (approved):
For me this is now good enough. But I don't fully understand the underlying issues, so I would like @karabowi or @fuhlig1 to take a second look (and press the Merge button).

fuhlig1 (approved):

fairroot/alignment/FairAlignmentHandler.cxx
```diff
@@ -100,15 +102,23 @@ void FairAlignmentHandler::AddAlignmentMatrices(const std::map<std::string, TGeo
...
+    for (auto vol : TRangeDynCast<TGeoVolume>(volumes)) {
+        if (!vol) {
+            continue;
+        }
+        if (vol->IsAssembly()) {
+            vol->GetShape()->ComputeBBox();
+        }
+        auto finder = vol->GetVoxels();
+        if (finder && finder->NeedRebuild()) {
+            finder->Voxelize();
```
dennisklein, line 120: I was curious what this `Voxelize()` function does and it looks to me as if [it computes the bounding box a second time](https://github.com/root-project/root/blob/454a6dff73228a35d97bbfe84a29a6a08d53de84/geom/geom/src/TGeoVoxelFinder.cxx#L2229) in that case? Is this intended?

PALoizeau: Indeed. I must confess I did not check carefully what was done under the hood in the ROOT calls as it fixed our new case.

From what we had in CBM, I would suspect that the two lines before this block in this PR are catching all Assemblies including those for which the `NeedRebuild`/`NeedRefresh` flag was not set, while the `Voxelize` is applied only to those where it was set but also having an affect on non-Assembly volumes.

But the best is probably to ask this from @agheata in the [ROOT issue 12242](https://github.com/root-project/root/issues/12242), as he is the author of both the `Voxelize` block and the patch applied here, plus the fact that the answer may be interesting also for the broader ROOT audience.

dennisklein: Side note, there is also a `TGeoVolume::Voxelize()` which seems to abort if the volume does not have a `TGeoVoxelFinder` (the thing returned by `GetVoxels()`) ... To me, the ROOT APIs look very confusing here.

fuhlig1:
@ChristianTackeGSI,

I don't see find the change in the v18.8 patches branch. Could you please check if it was really ported.

PR #1392, opened 2023-04-13
chore(eventdisplay): Refactor PDG To Color

ChristianTackeGSI:
This is part of a bigger patchset by @DanielWielanek.

Move the color configuration from FairEventManager to a new FairXMLPdgColor.

FairXMLPdgColor can also read config from an xml file, which is not yet used.

Co-Authored: Christian Tacke @ChristianTackeGSI

---

Checklist:

### DIFF
--- a/examples/advanced/propagator/src/FairEveRecoTracksExample.cxx
+++ b/examples/advanced/propagator/src/FairEveRecoTracksExample.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- *    Copyright (C) 2020 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ * Copyright (C) 2020-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -18,6 +18,7 @@
 
 #include "FairEveRecoTrack.h"
 #include "FairEveTrack.h"
+#include "FairEventManager.h"
 #include "FairField.h"
 #include "FairHit.h"
 #include "FairMCTrack.h"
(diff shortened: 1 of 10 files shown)


ChristianTackeGSI:
Hi @DanielWielanek,

thanks for that good idea!

I marked FairEventManager::StringToColor as deprecated (and fixed all callers).

I don't want to deprecate FairEventManager::Color in this Pull Request / commit, because there is no simple alternative for users of it.

We would need to make fPDGColor accessible otherwise (either by making it protected/public or by providing an accessor method). Maybe we should do that? I don't know.

ChristianTackeGSI:
Thanks @dennisklein! Fixed.

@DanielWielanek could you take a second look, if things are fine now with you?

DanielWielanek:
@ChristianTackeGSI Hi, I found a bug and code crash. I will try to fix it.

DanielWielanek:
@ChristianTackeGSI btw. during fixing this bug I noticed that ED sometimes crashes (also without your changes) if I use FairRoot ver 22. Did you see such behavior? When I tried rotate detectors (by mouse) ED crash, what is interesting is that with FairSoft ver 21 it works fine (doesn't matter if I use ED with/witout this new feature) that looks like a bug in ROOT.

ChristianTackeGSI:
@DanielWielanek

wow, interesting!

* If I understand you correctly, the changes in this PR (1392) do not in anyway interact with the bug you have found? So we could merge this one? Or are there other issues with this PR?
* If you think, that this is a bug in ROOT, could you first file a fresh issue on FairSoft about this. Especially, as it changes with version 22? Then we can try to track things down maybe? Maybe even the newest ROOT releases (there are some) also fixes this again? Anyway, it would be good to have an issue on FairSoft to track this?

> during fixing this bug 

Have you been able to fix it (or "avoid") the bug?

ChristianTackeGSI:
@karabowi and it does not crash without the changes in this branch?! Very, very strange!

DanielWielanek:
The Init is a bug that I fixed (but I cannot push changes to this PR - don't know why).
Then I run macro with two FairSoft version and discovered another which is realted with displaying of geometry (not tracks or hits) becuase it happens also without this new classes with XML colors.

ChristianTackeGSI:
@DanielWielanek 

> The Init is a bug that I fixed (but I cannot push changes to this PR - don't know why).

I see a few options:
* If your fix is not related to the Color changes, then I'd suggest to create a fresh branch and a fresh PR just for that.
* If it's a fix for the Color stuff, then maybe push your brnach (including the base Color change) to your FairRoot repo and create a fresh PR for this (I'll close my PR then, as your new PR would include those changes anyway).

DanielWielanek:
Actually its a trivial fix - I just call FairEveTracks::Init at the begin of derived classes Init (not at the end).

karabowi:
> @karabowi and it does not crash without the changes in this branch?! Very, very strange!

@ChristianTackeGSI When I remove the last commit from this branch, it still crashes. So it was something introduced earlier.

ChristianTackeGSI:
@DanielWielanek 

> Actually its a trivial fix - I just call FairEveTracks::Init at the begin of derived classes Init (not at the end).

This is a real bug fix, because those functions use GetEventManager(), which needs the FairEveTracks::Init first.

So please create a fresh PR, so you'll get all the credit for finding this one.

ChristianTackeGSI:
Fixes by @DanielWielanek added in <https://github.com/ChristianTackeGSI/FairRoot/pull/2>. I'll fix them up for this PR and create a separate one with only the fix for 18.8.

karabowi (approved):

DanielWielanek:
@ChristianTackeGSI it works, for both version of FairRoot (21 & 22) on my laptop, it seems that it crash only on my PC.

PR #1394, opened 2023-04-18
Switch ex_tutorial3_mq_* to Geant4

ChristianTackeGSI:
Also use more unique_ptr.

---

Checklist:

### DIFF
--- a/examples/advanced/Tutorial3/macro/run_sim.C
+++ b/examples/advanced/Tutorial3/macro/run_sim.C
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2022 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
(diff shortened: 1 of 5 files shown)


examples/advanced/Tutorial3/macro/run_sim.C
```diff
@@ -28,7 +32,7 @@ void run_sim(Int_t nEvents = 100, TString mcEngine = "TGeant3")
     gSystem->Setenv("CONFIG_DIR", tut_configdir.Data());
 
     // create Instance of Run Manager class
-    FairRunSim *fRun = new FairRunSim();
+    auto fRun = std::make_unique<FairRunSim>();
```
dennisklein, line 35: * How does this not conflict with the fact that [`FairRunSim` wants to be a singleton](https://github.com/FairRootGroup/FairRoot/blob/b0482233554eb7cc8ecc81bea588bfcb0e5a6572/fairroot/base/steer/FairRunSim.cxx#L68-L72)?
* And if it would no longer be a singleton, why not just put it on the stack?

dennisklein: Side note: If we touch that line, let's also rename the variable (probably in a separate commit). The `f` prefix suggests a member variable which it is clearly not.

ChristianTackeGSI: Well, you know, it's not a classic singleton. It's one that can be deallocated (by the owner). There's just a global pointer to it, for all those, who don't have a pointer to it.

All that said, I didn't want to touch every single line referencing it… but still deallocate it.

But you're right: Let's turn it into a stack variable (and as I need to touch every line anyway, let's a a rename in the same commit).

rbx (approved):
The code changes looks fine to me. But what the implications of switching from geant3 to geant4 are I don't know.  @karabowi should confirm.
MQ tests with sampler->processor->sink seem to behave same as before.

Issue #1346, opened 2023-02-09
Ownership semantics of `FairRootManager::GetObject`


ChristianTackeGSI:
Started by this discussion: https://github.com/FairRootGroup/FairRoot/pull/1340#discussion_r1099872737

`FairRootManager::GetObject` returns a `TObject*`. It is not obvious, whether this pointer is an owning or non-owning pointer.

Some code assumes it to be owning and consequently `delete`s the object at the end. Others don't.

The current implementation of this member function highly suggests that it returns non-owning pointers, because it usually returns copies of pointers (for example stored in `fObj2`–see little cleanup in #1343). @rbx came to this conclusion in the above mentioned discussion and I fully agree.

First the intended semantics should be clearly documented.

If it should return a non-owning pointer, then all callers should be analyzed to no longer `delete` the received pointer and FairRootManager should clean them up at the end (since it is owning the pointer).
If it should return an owning pointer, we should have an API that returns a unique_ptr (possibly already casted to the correct type).

ChristianTackeGSI:
An interesting question from this morning:

Should callers of GetObject be able to modify the object they get this way? If not, then GetObject maybe should return a `const TObject*`?

rbx:
Ideally it should be a `const TObject &`, or? Then it is clear that the manager maintains the ownership.

ChristianTackeGSI:
I guess GetObject could return a nullptr, if the thing isn't found?

What we need is something like a unique_ptr, just with no ownership semantics. In theory we could use a shared_ptr. But I don't know, if this really is the way to go, or whether it's just the biggest hammer.

karabowi:
> An interesting question from this morning:
> 
> Should callers of GetObject be able to modify the object they get this way? If not, then GetObject maybe should return a `const TObject*`?

That is a good question. On one hand there may be cases, where the data modification is good, but generally I would say the data should not be modified.

karabowi:
However. Simple example of the time based reconstruction and `FairTSBufferFunctional`.
The `fInputArray` is obtained via `FRM::GetObject()`:
https://github.com/FairRootGroup/FairRoot/blob/master/base/steer/FairTSBufferFunctional.cxx#L42
and here it is absorbed into `fOutputArray`:
https://github.com/FairRootGroup/FairRoot/blob/master/base/steer/FairTSBufferFunctional.cxx#L54
But looking at the description of the `AbsorbObjects` function(s):
[...]

PR #1396, opened 2023-04-28
Remove unused FairRunOnline::InitContainers() function

karabowi:
The function is not being invoked, and was not used.

Addresses issue #1346.

---

Checklist:

### DIFF
--- a/fairroot/online/steer/FairRunOnline.cxx
+++ b/fairroot/online/steer/FairRunOnline.cxx
@@ -215,16 +215,20 @@ Int_t FairRunOnline::EventLoop()
 
     FillEventHeader();
     auto const tmpId = GetEvtHeaderRunId();
-
     if (tmpId != fRunId) {
-        LOG(info) << "FairRunOnline::EventLoop() Call Reinit due to changed RunID (from " << fRunId << " to " << tmpId;
+        LOG(info) << "FairRunOnline::EventLoop() Detected changed RunID from " << fRunId << " to " << tmpId;
         fRunId = tmpId;
-        Reinit(fRunId);
-        if (!GetSource()->ReInitUnpackers()) {
-            LOG(fatal) << "FairRunOnline->EventLoop() ReInitUnpackers() failed!";
-            return 1;
+        if (!fStatic) {
+            LOG(info) << "FairRunOnline::EventLoop() Call Reinit.";
+            Reinit(fRunId);
+            if (!GetSource()->ReInitUnpackers()) {
+                LOG(fatal) << "FairRunOnline->EventLoop() ReInitUnpackers() failed!";
+                return 1;
+            }
+            fTask->ReInitTask();
+        } else {
+            LOG(info) << "FairRunOnline::EventLoop() ReInit not called because initialisation is static.";
         }
-        fTask->ReInitTask();
     }
 
     fRootManager->StoreWriteoutBufferData(fRootManager->GetEventTime());
--- a/fairroot/online/steer/FairRunOnline.cxx
+++ b/fairroot/online/steer/FairRunOnline.cxx
@@ -185,13 +185,12 @@ void FairRunOnline::Init()
     fTask->SetParTask();
     fRtdb->initContainers(fRunId);
 
-    //  InitContainers();
     // --- Get event header from Run
     if (!fEvtHeader) {
-        LOG(fatal) << "FairRunOnline::InitContainers:No event header in run!";
+        LOG(fatal) << "FairRunOnline::Init() No event header in run!";
         return;
     }
-    LOG(info) << "FairRunOnline::InitContainers: event header at " << fEvtHeader;
+    LOG(info) << "FairRunOnline::Init() Event header at " << fEvtHeader;
     fRootManager->Register("EventHeader.", "Event", fEvtHeader, (nullptr != GetSink()));
     fEvtHeader->SetRunId(fRunId);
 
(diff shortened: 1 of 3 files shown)


fairroot/online/steer/FairRunOnline.cxx
```diff
@@ -185,7 +185,6 @@ void FairRunOnline::Init()
     fTask->SetParTask();
     fRtdb->initContainers(fRunId);
 
-    //  InitContainers();
     // --- Get event header from Run
     if (!fEvtHeader) {
         LOG(fatal) << "FairRunOnline::InitContainers:No event header in run!";
```
ChristianTackeGSI, line 190: Can you please fix those LOG messages? They reference InitContainers.
(There's another one a few lines down.)

fairroot/online/steer/FairRunOnline.h
```diff
@@ -55,8 +55,6 @@ class FairRunOnline : public FairRun
      *  is not checked anymore after initialization
      */
 
-    /** Init containers executed on PROOF, which is part of Init when running locally*/
-    void InitContainers();
     void SetContainerStatic(Bool_t tempBool = kTRUE);
     Bool_t GetContainerStatic() { return fStatic; };
```
ChristianTackeGSI, line 59: Can you take a look at `SetContainerStatic`/`GetContainerStatic` as well? They look unused as well?

As InitContainers (and the *Static ones) are public, could you consider adding a (breaking) notice to the ChangeLog?

ChristianTackeGSI (commented):
Very good catch!

I have a few notes.

fairroot/online/steer/FairRunOnline.cxx
karabowi: You are right, I have changed it and added ChangeLog notice.

fairroot/online/steer/FairRunOnline.h
kresan: @karabowi In R3B scenario one always had to update parameters when RunID was changing.

But for generic use case, I agree with you, that we should check for `! fStatic`.

karabowi: I have discussed the issue with both @kresan and @fuhlig1 , and we agreed that `fStatic` should be used in a proper way in `FairRunOnline`. Appropriate commit will be attached to this PR later today.

fairroot/online/steer/FairRunOnline.cxx
```diff
@@ -217,12 +217,18 @@ Int_t FairRunOnline::EventLoop()
...
 
     if (tmpId != fRunId) {
-        LOG(info) << "FairRunOnline::EventLoop() Call Reinit due to changed RunID (from " << fRunId << " to " << tmpId;
         fRunId = tmpId;
-        Reinit(fRunId);
-        if (!GetSource()->ReInitUnpackers()) {
-            LOG(fatal) << "FairRunOnline->EventLoop() ReInitUnpackers() failed!";
-            return 1;
+        if (!fStatic) {
+            LOG(info) << "FairRunOnline::EventLoop() Call Reinit due to changed RunID (from " << fRunId << " to "
```
ChristianTackeGSI, line 222: The `fRunId = tmpId;` in line 221/220 will make both the same.
So the LOG will only show the new RunID.

Maybe move the LOG above the assignment and make a second LOG in both if branches?

fairroot/online/steer/FairRunOnline.cxx
```diff
@@ -217,12 +217,18 @@ Int_t FairRunOnline::EventLoop()
...
+            Reinit(fRunId);
+            if (!GetSource()->ReInitUnpackers()) {
+                LOG(fatal) << "FairRunOnline->EventLoop() ReInitUnpackers() failed!";
+                return 1;
+            }
+        } else {
+            LOG(info) << "FairRunOnline::EventLoop() Changed RunID (from " << fRunId << " to " << tmpId
+                      << "). ReInit not called because initialisation is static.";
         }
         fTask->ReInitTask();
```
ChristianTackeGSI, line 233: FairRunAna has the `fTask->ReInitTask();` inside the `if(!fStatic)`?

See for example:

https://github.com/FairRootGroup/FairRoot/blob/76fc0d676477f188843af073e74fa2ffd789db09/fairroot/base/steer/FairRunAna.cxx#L348-L351

So I wonder, why this is not consistant here?

If there is a reason, please put a comment about that reason here.

ChristianTackeGSI (commented):
I also wonder, whether we should refactor this a bit into some common FairRun feature?

Or should FairRunAna and FairRunOnline have common helper base class?

Just pondering! I haven't thought this through, really!

fairroot/online/steer/FairRunOnline.cxx
karabowi: Fixed the commit to include `ReInitTask()` in the `!fStatic` case.
Now the comments should print correct values.

fairroot/online/steer/FairRunOnline.cxx
```diff
@@ -209,60 +208,24 @@ void FairRunOnline::Init()
...
+            LOG(info) << "FairRunOnline::EventLoop() Call Reinit due to changed RunID (from " << fRunId << " to "
+                      << tmpId << ")";
+            Reinit(tmpId);
+            fTask->ReInitTask();
+        } else {
+            LOG(info) << "FairRunOnline::EventLoop() Changed RunID (from " << fRunId << " to " << tmpId
+                      << "). ReInit not called because initialisation is static.";
         }
-        fTask->ReInitTask();
+        fRunId = tmpId;
```
ChristianTackeGSI, line 228: `fRunId is set quite late in your change. So for example `fTask->ReInitTask` might see the old fRunId?

FairRunAna sets it earlier:

https://github.com/FairRootGroup/FairRoot/blob/76fc0d676477f188843af073e74fa2ffd789db09/fairroot/base/steer/FairRunAna.cxx#L347-L351

fairroot/online/steer/FairRunOnline.cxx
```diff
@@ -209,60 +208,24 @@ void FairRunOnline::Init()
...
-        fRunId = tmpId;
-        Reinit(fRunId);
-        if (!GetSource()->ReInitUnpackers()) {
-            LOG(fatal) << "FairRunOnline->EventLoop() ReInitUnpackers() failed!";
-            return 1;
+        if (!fStatic) {
+            LOG(info) << "FairRunOnline::EventLoop() Call Reinit due to changed RunID (from " << fRunId << " to "
+                      << tmpId << ")";
+            Reinit(tmpId);
+            fTask->ReInitTask();
```
ChristianTackeGSI, line 223: Where did `GetSource()->ReInitUnpackers()`? I am not seeing it?

karabowi: Good point again, fixed hopefully finally.

ChristianTackeGSI (approved):

Issue #1397, opened 2023-05-10
FairRunAna::Run() unclear description


karabowi:
**Is your feature request related to a problem? Please describe.**
Current implementation of `FairRunAna::Run()` does not match the description.
What we have now in `FairRunAna` is:
```
    /**Run from event number NStart to event number NStop */
    void Run(Int_t NStart = 0, Int_t NStop = 0) override;
    /**Run for the given single entry*/
    void Run(Long64_t entry);
```
Only the first function with NStart < NStop works as described.
The second function is never invoked, i.e. it is impossible to run single event this way.

Unexpected behavior:
Run(-1) - breaks in FairRunInfo::WriteInfo()
Run(0) - runs all input (100 events)
Run(15) - runs first 15 events

Run(-1,any int) - breaks in FairRunInfo::WriteInfo()
Run(15,-1) - runs from event 15 till the end
Run(15,0) - runs first 15 events
Run(15,15) - breaks in FairRunInfo::WriteInfo()
Run(15,16) - runs event number 15

**Describe the solution you'd like**
Fix the comments to match the behavior.

Change `Run(Long64_t entry)` to `RunSingle(Long64_t entry)`

Break properly in `FairRunAna::Run()` when first `int` parameter is `negative`.

Change the behavior of:
Run(15,0) - runs first 15 events
to:
Run(15,0) - run from event 15 till the end

Run(15,15) - breaks in FairRunInfo::WriteInfo()
to 
Run(15,15) - run event number 15

Run(15,16) - runs event number 15
to
Run(15,16) - run event 15 and 16

**Describe alternatives you've considered**
Maybe `RunEvent(Long64_t entry)` is better than `RunSingle(Long64_t entry)`?

Maybe we should keep: Run(15,0) - run first 15 events, I really would like to hear your opinion.

**Additional context**
---

rbx:
There is also:
```C++
    /**Run over the whole input file with timpe window delta_t as unit (entry)*/
    void Run(Double_t delta_t);
```
Although it is a little bit harder to confuse because it is using a double, it's still error-prone. E.g. if the value is taken from some config container with `auto`.

PR #1405, opened 2023-05-23
Create `FairBaseTypes.h` with `alias using EventID = UInt_t`

karabowi:
Basically moved the `alias EventID` from `FairMCEventHeader`
to namespace `fr` in `FairBaseTypes.h`.

Bunch of files affected.

Addresses issue #1397.

---

Checklist:


dennisklein:
> Removed the huge commit and left only the basic one. Modified the message.
> 
> @dennisklein . Yes, you are right. However after some investigation, and a short discussion with @ChristianTackeGSI, I have decided to stay by the name `EventID`, which was already implemented in `FairMCEventHeader`.

Even more confused

fairroot/tools/FairBaseTypes.h
```diff
@@ -0,0 +1,28 @@
...
+ ********************************************************************************/
+/*
+ * FairBaseTypes.h
+ *
+ *  Created on: May 16, 2023
+ *      Author: R.Karabowicz
+ */
+
+#ifndef BASE_FAIRTYPES_H_
+#define BASE_FAIRTYPES_H_
```
ChristianTackeGSI, line 16: I don't know really, but I guess, I would prefer an identifier that looks more like the fulename.

```suggestion
#ifndef FAIRBASETYPES_H_
#define FAIRBASETYPES_H_
```

fairroot/tools/FairBaseTypes.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include <Rtypes.h>
+
+namespace fr {
```
ChristianTackeGSI, line 20:
```suggestion
namespace fairroot {
```

What about a more descriptive namespace?
Yes, this feels like more typing. But really, it's 2023 and we have good editors.

fairroot/tools/CMakeLists.txt
```diff
@@ -20,6 +20,11 @@ set(sources
 
 fair_change_extensions_if_exists(.cxx .h FILES "${sources}" OUTVAR headers)
 
+set(headers
+  ${headers}
+  FairBaseTypes.h
+)
```
ChristianTackeGSI, line 26: I hope this works:

```suggestion
list(APPEND headers
  FairBaseTypes.h
)
```

ChristianTackeGSI (commented):
First some detailed comments.

fairroot/tools/FairBaseTypes.h
```diff
@@ -0,0 +1,28 @@
...
+    /// Alias name for event id type
+    using EventID = UInt_t;
+
+    /// Sentinel value for the event id, when none set
+    static constexpr EventID NoEventID = std::numeric_limits<EventID>::max();
```
ChristianTackeGSI, line 25:
```
/shared/alfaci/jenkins/workspace/FairRootGroup_FairRoot_PR-1405@13/tools/FairBaseTypes.h:25:47: error: no member named 'numeric_limits' in namespace 'std' [clang-diagnostic-error]
    static constexpr EventID NoEventID = std::numeric_limits<EventID>::max();
                                         ~~~~~^
```

Please add a `#include <limits>` to this header.

ChristianTackeGSI (changes requested):

ChristianTackeGSI:
About the confusion:

I am not sure either about all of this.

IF FairRoot has some fields (as in: It's stored in the "datasets" by FairRoot) and the experiment is responsible for this, then yes, we should have a datatype (and corresponding type alias) for it.

And this specific one actually looks quite like it? Especially as for simulation (MC…Header) it looks like the entries are actually events, right?

BTW: I think, we should consider [scoped enumerations](https://en.cppreference.com/w/cpp/language/enum#Scoped_enumerations) for these types, so that a conversion from say an eventid to an entryno can not be done accidentally. One then has to think before doing it.

PR #1409, opened 2023-06-11
chore(examples): More includes in macros

ChristianTackeGSI:
Even so macros do not need `#include`s, they're a nice way of documenting what's needed. And they help IDEs to find the right symbols/etc.

The FairRoot includes themselves still have issues with this, so we only include ROOT and other includes for now.

---

Checklist:


dennisklein:
> Even so macros do not need `#include`s, they're a nice way
of documenting what's needed. And they help IDEs to find
the right symbols/etc.

:+1:

An additional argument for having `#include`s in macros would be [ACLiC support](https://root.cern.ch/root/htmldoc/guides/users-guide/Cling.html#aclic-compiling-scripts-into-libraries).

dennisklein (approved):

PR #1410, opened 2023-06-13
Create `FairBaseTypes.h`

karabowi:
Created `namespace FairRoot`
with `struct EntryID {size_t fId;}`.

Should replace PR #1405.

---

Checklist:

### DIFF
--- a/fairroot/base/steer/FairRunAna.h
+++ b/fairroot/base/steer/FairRunAna.h
@@ -35,25 +35,34 @@ class FairRunAna : public FairRun
     FairRunAna();
     /**initialize the run manager*/
     void Init() override;
-    /**Run from event number NStart to event number NStop */
-    void Run(Int_t NStart = 0, Int_t NStop = 0) override;
+    /**Run analysis:
+     * for the whole input if NEntry == NStop == 0,
+     * for NEntry entries if NStop == 0,
+     * from NEntry until NStop entry is reached
+     * from NEntry to the end of input, if NStop = -1 */
+    void Run(FairRoot::EntryID NEntry = 0, int NStop = 0) override;
     /**Run over the whole input file with timpe window delta_t as unit (entry)*/
     void Run(Double_t delta_t);
+    /** \deprecated Deprecated in v19, will be removed in v20. */
+    [[deprecated("Replaced by RunSingle(), which has an unambiguous name")]] void Run(Long64_t entry)
+    {
+        RunSingle(entry);
+    }
     /**Run for the given single entry*/
-    void Run(Long64_t entry);
-    /**Run event reconstruction from event number NStart to event number NStop */
-    void RunEventReco(Int_t NStart, Int_t NStop);
+    void RunSingle(FairRoot::EntryID entry);
+    /**Run event reconstruction from entry number NEntry to entry number NStop */
+    void RunEventReco(FairRoot::EntryID NEntry, Int_t NStop);
     /**Run over all TSBuffers until the data is processed*/
     void RunTSBuffers();
     /** the dummy run does not check the evt header or the parameters!! */
-    void DummyRun(Int_t NStart, Int_t NStop);
+    void DummyRun(FairRoot::EntryID NStart, FairRoot::EntryID NStop);
     /** This methode is only needed and used with ZeroMQ
-     * it read a certain event and call the task exec, but no output is written
+     * it read a certain entry and call the task exec, but no output is written
      * @param entry : entry number in the tree
      */
-    void RunMQ(Long64_t entry);
+    void RunMQ(FairRoot::EntryID entry);
     /** Run on a list of lmd files*/
-    void RunOnLmdFiles(UInt_t NStart = 0, UInt_t NStop = 0);
+    void RunOnLmdFiles(FairRoot::EntryID NStart = 0, FairRoot::EntryID NStop = 0);
 
     void RunOnTBData();
     /** finish tasks, write output*/
--- a/fairroot/base/steer/FairRun.cxx
+++ b/fairroot/base/steer/FairRun.cxx
@@ -32,7 +32,10 @@
 
 TMCThreadLocal FairRun* FairRun::fRunInstance = nullptr;
 
-FairRun* FairRun::Instance() { return fRunInstance; }
+FairRun* FairRun::Instance()
+{
+    return fRunInstance;
+}
 
 FairRun::FairRun(Bool_t isMaster)
     : TNamed()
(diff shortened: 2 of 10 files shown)


fairroot/tools/CMakeLists.txt
```diff
@@ -20,6 +20,10 @@ set(sources
 
 fair_change_extensions_if_exists(.cxx .h FILES "${sources}" OUTVAR headers)
 
+list(APPEND headers
+  FairBaseTypes.h
```
dennisklein, line 24: * How about just `FairTypes.h`?
* I would put such a header into the base lib instead of tools (which I interpret more as a collection of helpers/utilities)?

ChristianTackeGSI (commented):
I like this!

Here is a first set of comments to improve some parts.

fairroot/tools/FairBaseTypes.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+#ifndef FAIR_BASETYPES_H_
+#define FAIR_BASETYPES_H_
+
+#include <iostream>
+#include <limits>
+
+namespace FairRoot {
+    struct EntryID {
```
ChristianTackeGSI, line 22: Consider to add some Doxygen comments to the namespace and class?

```suggestion
/**
 * \brief Main namespace for FairRoot classes / types
 *
 * All new public classes and types of FairRoot will be in this namespace.
 * Note: Nost classes are still in no namespace.
 */
namespace FairRoot {

/**
 * \brief <What Entry is this ID meant for?>
 */
struct EntryID
{
```

ChristianTackeGSI (commented):
The first iteration of the Doxygen docs for the namespace is [here](http://lxci001.gsi.de/ci/tmp/jenkins-FairRootGroup-FairRoot-PR-1410-1/db/da7/namespaceFairRoot.html). And the link to the *current* Doxygen build for this PR can be found here: https://github.com/FairRootGroup/FairRoot/pull/1410/checks (and click on Doxygen-Preview).

fairroot/base/steer/FairTypes.h
```diff
@@ -0,0 +1,106 @@
...
+#include <iostream>
+#include <limits>
+
+/**
+ * \brief Main namespace for FairRoot classes / types
+ *
+ * All new public classes and types of FairRoot will be in this namespace.
+ * Note: Nost classes are still in no namespace.
+ */
+namespace FairRoot {
```
rbx, line 27: Didn't we agree on a lowercase name for this?
```suggestion
namespace fairroot {
```
If we didn't, I would still suggest to use lowercase. Checking the guidelines of Mozilla (upon which we base our clang-format), they seem to suggest lowercase: https://firefox-source-docs.mozilla.org/code-quality/coding-style/coding_style_cpp.html#c-namespaces

So do the Alice guidelines: https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html?showone=Namespace_Names#Namespace_Names

For me it's more a personal preference. But one objective reason could be so they are distinguishable from class names.

fairroot/base/steer/FairTypes.h
```diff
@@ -0,0 +1,106 @@
...
+#define FAIR_TYPES_H_
+
+#include <iostream>
+#include <limits>
+
+/**
+ * \brief Main namespace for FairRoot classes / types
+ *
+ * All new public classes and types of FairRoot will be in this namespace.
+ * Note: Nost classes are still in no namespace.
```
rbx, line 25:
```suggestion
 * Note: Most classes are still in no namespace.
```

dennisklein:
@rbx But I am not 100% convinced myself, if this is the best use case for a strong alias. Especially, as there is already existing code using all kinds of int, uint, size_t etc for the same thing including serialized root files that need to be read back etc. Not sure, how much will break for the good or the worse^^

fairroot/base/steer/FairRunAna.h
```diff
@@ -35,25 +35,34 @@ class FairRunAna : public FairRun
...
     void Run(Double_t delta_t);
+    /** \deprecated Deprecated in v19, will be removed in v20. */
+    [[deprecated("Replaced by RunSingle(), which has an unambiguous name")]] void Run(Long64_t entry)
+    {
+        RunSingle(entry);
+    }
     /**Run for the given single entry*/
-    void Run(Long64_t entry);
-    /**Run event reconstruction from event number NStart to event number NStop */
-    void RunEventReco(Int_t NStart, Int_t NStop);
```
dennisklein, line 45: we also should deprecated the int, int overloads first before removing them, right?

dennisklein:
If we see a potential for more "strong aliases" like this one in the near future, we could adopt the templates proposed here: https://www.foonathan.net/2016/10/strong-typedefs/

karabowi:
> > Soooo, what are the benefits of going full type here?
> 
> I guess, the idea would be to have safer-to-use and better-self-documenting function signatures (and call sites), e.g.
> 
> ```c++
> void RunReco(EntryID start, EntryID stop); // range [start, stop]
> void RunReco(EntryID start, int n);        // range [start, start + n]
> void RunReco(EntryID id);                  // single entry
> void RunReco(int n);                       // n entries (starting from some implicit cursor)
> ```
> 
> With an alias it would be less clear what is meant and the compiler couldn't help you, if you misuse one of the signatures. (some overloads wouldn't even be possible). Now imagine argument lists with even more int args, where some are timeouts in seconds, different ids etc, easy to mess up the order when calling.

Are you also proposing to ultimately replace the `Run()` functions with `RunReco()`?

dennisklein:
> > > Soooo, what are the benefits of going full type here?
> > 
> > 
> > I guess, the idea would be to have safer-to-use and better-self-documenting function signatures (and call sites), e.g.
> > ```c++
> > void RunReco(EntryID start, EntryID stop); // range [start, stop]
> > void RunReco(EntryID start, int n);        // range [start, start + n]
> > void RunReco(EntryID id);                  // single entry
> > void RunReco(int n);                       // n entries (starting from some implicit cursor)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With an alias it would be less clear what is meant and the compiler couldn't help you, if you misuse one of the signatures. (some overloads wouldn't even be possible). Now imagine argument lists with even more int args, where some are timeouts in seconds, different ids etc, easy to mess up the order when calling.
> 
> Are you also proposing to ultimately replace the `Run()` functions with `RunReco()`?

? I cannot follow

PR #1412, opened 2023-06-22
Fix root GetSeed

fuhlig1:
This PR brings first a test which checks if the set seed value is properly stored in FairBasParSet container during the simulation. Due to a breaking change in ROOT 6.24 the return value of the function TRandom3::GetSeed() doesn't return the seed value but a fixed value of 624. The newly added test should fail for FairSoft versions apr22 and nov22.
The PR will be updated with the proper fix for the problem in a second step.

---

Checklist:

* [x ] Followed the [Contributing Guidelines](https://github.com/FairRootGroup/FairRoot/blob/dev/CONTRIBUTING.md)

### DIFF
--- a/examples/simulation/Tutorial2/macros/CMakeLists.txt
+++ b/examples/simulation/Tutorial2/macros/CMakeLists.txt
@@ -12,21 +12,30 @@ GENERATE_ROOT_TEST_SCRIPT(${CMAKE_CURRENT_SOURCE_DIR}/read_digis.C)
 GENERATE_ROOT_TEST_SCRIPT(${CMAKE_CURRENT_SOURCE_DIR}/run_background.C)
 GENERATE_ROOT_TEST_SCRIPT(${CMAKE_CURRENT_SOURCE_DIR}/run_signal.C)
 GENERATE_ROOT_TEST_SCRIPT(${CMAKE_CURRENT_SOURCE_DIR}/run_tutorial2.C)
+GENERATE_ROOT_TEST_SCRIPT(${CMAKE_CURRENT_SOURCE_DIR}/compare_seed_value.C)
 
 set(maxTestTime 60)
 
 add_test(NAME ex_sim_tutorial2
-         COMMAND ${CMAKE_CURRENT_BINARY_DIR}/run_tutorial2.sh)
+         COMMAND ${CMAKE_CURRENT_BINARY_DIR}/run_tutorial2.sh  10 \"TGeant4\" true 333)
 set_tests_properties(ex_sim_tutorial2 PROPERTIES
   TIMEOUT ${maxTestTime}
   PASS_REGULAR_EXPRESSION "Macro finished successfully"
   FIXTURES_SETUP fixtures.ex_sim_tutorial2_sim
 )
 
+add_test(NAME ex_compare_seed_value
+         COMMAND ${CMAKE_CURRENT_BINARY_DIR}/compare_seed_value.sh \"tutorial2_pions.params_p2.000_t0_n10.root\" 333)
+set_tests_properties(ex_compare_seed_value PROPERTIES
+  TIMEOUT ${maxTestTime}
+  FIXTURES_REQUIRED fixtures.ex_sim_tutorial2_sim
+  FIXTURES_SETUP fixtures.ex_compare_seed_value
+)
+
 add_test(NAME ex_sim_tutorial2_create_digis
          COMMAND ${CMAKE_CURRENT_BINARY_DIR}/create_digis.sh)
 set_tests_properties(ex_sim_tutorial2_create_digis PROPERTIES
-  FIXTURES_REQUIRED fixtures.ex_sim_tutorial2_sim
+  FIXTURES_REQUIRED fixtures.ex_compare_seed_value
   TIMEOUT ${maxTestTime}
   PASS_REGULAR_EXPRESSION "Macro finished successfully"
   FIXTURES_SETUP fixtures.ex_sim_tutorial2_digi
--- a/examples/simulation/Tutorial2/macros/CMakeLists.txt
+++ b/examples/simulation/Tutorial2/macros/CMakeLists.txt
@@ -1,5 +1,5 @@
 ################################################################################
-# Copyright (C) 2014-2019 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  #
+# Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  #
 #                                                                              #
 #              This software is distributed under the terms of the             #
 #              GNU Lesser General Public Licence (LGPL) version 3,             #
@@ -22,23 +22,25 @@ set_tests_properties(ex_sim_tutorial2 PROPERTIES
   TIMEOUT ${maxTestTime}
   PASS_REGULAR_EXPRESSION "Macro finished successfully"
   FIXTURES_SETUP fixtures.ex_sim_tutorial2_sim
+  RESOURCE_LOCK ex_sim_tutorial2_paramFile
 )
 
-add_test(NAME ex_compare_seed_value
+add_test(NAME ex_sim_tutorial2_compare_seed_value
          COMMAND ${CMAKE_CURRENT_BINARY_DIR}/compare_seed_value.sh \"tutorial2_pions.params_p2.000_t0_n10.root\" 333)
-set_tests_properties(ex_compare_seed_value PROPERTIES
+set_tests_properties(ex_sim_tutorial2_compare_seed_value PROPERTIES
   TIMEOUT ${maxTestTime}
   FIXTURES_REQUIRED fixtures.ex_sim_tutorial2_sim
-  FIXTURES_SETUP fixtures.ex_compare_seed_value
+  RESOURCE_LOCK ex_sim_tutorial2_paramFile
 )
 
 add_test(NAME ex_sim_tutorial2_create_digis
          COMMAND ${CMAKE_CURRENT_BINARY_DIR}/create_digis.sh)
 set_tests_properties(ex_sim_tutorial2_create_digis PROPERTIES
-  FIXTURES_REQUIRED fixtures.ex_compare_seed_value
+  FIXTURES_REQUIRED fixtures.ex_sim_tutorial2_sim
   TIMEOUT ${maxTestTime}
   PASS_REGULAR_EXPRESSION "Macro finished successfully"
   FIXTURES_SETUP fixtures.ex_sim_tutorial2_digi
+  RESOURCE_LOCK ex_sim_tutorial2_paramFile
 )
 
 add_test(NAME ex_sim_tutorial2_read_digis
--- a/examples/simulation/Tutorial2/macros/compare_seed_value.C
+++ b/examples/simulation/Tutorial2/macros/compare_seed_value.C
@@ -1,0 +1,20 @@
+int compare_seed_value(TString filename, UInt_t initial_seed) {
+
+  // Open parameter file and get the stored random seed from the
+  // FairBaseParSet parameter container
+  // Compare the random seed with the expected one set during the
+  // transport run
+
+  std::unique_ptr<TFile> myFile( TFile::Open(filename) );
+  std::unique_ptr<FairBaseParSet> par(myFile->Get<FairBaseParSet>("FairBaseParSet"));
+  UInt_t seed = par->GetRndSeed();
+
+  if ( seed == initial_seed ) {
+    return 0;
+  }
+  else {
+    std::cout << "Expected seed value  : " << initial_seed << std::endl;
+    std::cout << "Seed value from file : " << seed << std::endl;
+    return 1;
+  }
+}
--- a/examples/simulation/Tutorial2/macros/compare_seed_value.C
+++ b/examples/simulation/Tutorial2/macros/compare_seed_value.C
@@ -1,20 +1,47 @@
-int compare_seed_value(TString filename, UInt_t initial_seed) {
+/********************************************************************************
+ *    Copyright (C) 2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ *                                                                              *
+ *              This software is distributed under the terms of the             *
+ *              GNU Lesser General Public Licence (LGPL) version 3,             *
+ *                  copied verbatim in the file "LICENSE"                       *
+ ********************************************************************************/
+#if !defined(__CLING__) || defined(__ROOTCLING__)
+#include "FairBaseParSet.h"
 
-  // Open parameter file and get the stored random seed from the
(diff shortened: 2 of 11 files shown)


examples/simulation/Tutorial2/macros/compare_seed_value.C
```diff
@@ -0,0 +1,20 @@
+int compare_seed_value(TString filename, UInt_t initial_seed) {
```
ChristianTackeGSI, line 1: Could you add a copyright header and basic set of includes, please?

```suggestion
/********************************************************************************
 *    Copyright (C) 2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
 *                                                                              *
 *              This software is distributed under the terms of the             *
 *              GNU Lesser General Public Licence (LGPL) version 3,             *
 *                  copied verbatim in the file "LICENSE"                       *
 ********************************************************************************/

#include <TFile.h>
#include <iostream>
#include <memory>

int compare_seed_value(TString filename, UInt_t initial_seed) {
```

examples/simulation/Tutorial2/macros/CMakeLists.txt
```diff
@@ -12,21 +12,30 @@ GENERATE_ROOT_TEST_SCRIPT(${CMAKE_CURRENT_SOURCE_DIR}/read_digis.C)
...
+  TIMEOUT ${maxTestTime}
+  FIXTURES_REQUIRED fixtures.ex_sim_tutorial2_sim
+  FIXTURES_SETUP fixtures.ex_compare_seed_value
+)
+
 add_test(NAME ex_sim_tutorial2_create_digis
          COMMAND ${CMAKE_CURRENT_BINARY_DIR}/create_digis.sh)
 set_tests_properties(ex_sim_tutorial2_create_digis PROPERTIES
-  FIXTURES_REQUIRED fixtures.ex_sim_tutorial2_sim
+  FIXTURES_REQUIRED fixtures.ex_compare_seed_value
```
ChristianTackeGSI, line 38: Does `ex_sim_tutorial2_create_digis` really depend on the Seed being safed correctly?

examples/simulation/Tutorial2/macros/run_tutorial2.C
```diff
@@ -12,7 +12,8 @@
 #include <TSystem.h>
 #include <memory>
 
-void run_tutorial2(Int_t nEvents = 10, TString mcEngine = "TGeant4", Bool_t isMT = true)
+void run_tutorial2(Int_t nEvents = 10, TString mcEngine = "TGeant4",
+                   Bool_t isMT = true, UInt_t initial_seed=98989)
```
ChristianTackeGSI, line 16: Please update the copyright header on this file as well.

examples/simulation/Tutorial2/macros/compare_seed_value.C
```diff
@@ -0,0 +1,20 @@
...
+
+  // Open parameter file and get the stored random seed from the
+  // FairBaseParSet parameter container
+  // Compare the random seed with the expected one set during the
+  // transport run
+
+  std::unique_ptr<TFile> myFile( TFile::Open(filename) );
```
dennisklein, line 8: Prevent potential segfault in line 9:
```suggestion
std::unique_ptr<TFile> myFile( TFile::Open(filename) );
if (!myFile) {
  std::cout << "Could not open file: " << filename << std::endl;
  return 1;
}
```

examples/simulation/Tutorial2/macros/compare_seed_value.C
```diff
@@ -0,0 +1,20 @@
...
+  std::unique_ptr<FairBaseParSet> par(myFile->Get<FairBaseParSet>("FairBaseParSet"));
```
dennisklein, line 9: Are you sure `Get` returns ownership?

examples/simulation/Tutorial2/macros/CMakeLists.txt
fuhlig1: No, but this was the easiest way to fix the execution order and avoid that create_digis.C and compare_seed_value.C will run in parallel since both of them access the same parameter files.
I changed this now by using a resource lock. That means both tests now depend on the simulation but there is an additional resource lock such that latter macros are executes sequentially.

ChristianTackeGSI: As long as they only read from the file, it shouldn't be a problem, if they run in parallel?

examples/simulation/Tutorial2/macros/run_tutorial2.C
fuhlig1: Done. I added also missing header files. Since I was working on the issue anyway I also added the missing header files in all other macros of Tutorial2.

examples/simulation/Tutorial2/macros/compare_seed_value.C
ChristianTackeGSI: Looking at the [source of `TDirectoryFile::Get`](https://root.cern.ch/doc/master/TDirectoryFile_8cxx_source.html#l00937) (and especially the first `idcur` part) it looks like this returns a non-owning pointer.

examples/simulation/Tutorial2/macros/compare_seed_value.C
fuhlig1: Done.

examples/simulation/Tutorial2/macros/compare_seed_value.C
fuhlig1: I don't know. Why do you ask?

dennisklein: Because if you store it in a `unique_ptr` as you do, it will `delete` it once it goes out of scope. So, your code makes two assumptions:
1. `Get` passes ownership to you
2. The pointer points to "`delete`able" heap memory

fuhlig1:
When adding the FairRoot header files in the macro as well as adding the needed path information to CPATH such that the FairRoot headers are found I get the following compilation error

> Processing /opt/fairroot/source/dev/examples/simulation/Tutorial2/macros/run_tutorial2.C()...
> G__Base dictionary payload:3692:23: error: cannot initialize a parameter of type 'TObject *' with an lvalue of type 'FairGenerator *'
>         fGenList->Add(generator);
>                       ^~~~~~~~~
> /opt/fairsoft/nov22p1/include/root/TObjArray.h:68:34: note: passing argument to parameter 'obj' here
>    void             Add(TObject *obj) { AddLast(obj); }
>                                  ^
> Assertion failed: (!Calls.empty() && "Missing lambda call operator!"), function getLambdaCallOperator, file DeclCXX.cpp, line 1397.

The code is in FairPrimaryGenerator and compiles without problems. Does anybody has any idea what could be the problem.

examples/simulation/Tutorial2/macros/create_digis.C
```diff
@@ -6,6 +6,7 @@
  *                  copied verbatim in the file "LICENSE"                       *
  ********************************************************************************/
 
+/*
```
ChristianTackeGSI, line 9: I used to not add our FairRoot headers in the macros for now, exactly for this reason (does not yet work).

Instead of commenting them out, what about something like this:
https://github.com/FairRootGroup/FairRoot/blob/6031156f11d95b3e4bd4470d705a76bcfce98773/examples/simulation/Tutorial1/macros/run_tutorial1_binary.C#L9

(Disclaimer: I don't know, which symbols are defined or not defined in the interpreter!)

fuhlig1: The commenting was for now to see if things work as expected. I will go with the preprocessor statement for now.

I could work around the first crash by moving some code block from the header to the source file but ran into another crash right after.

ChristianTackeGSI:
> Does anybody has any idea what could be the problem.

I looked into this a few months ago.

My conclusion at that time was:
* We have the headers baked into the distionary (where the root interpreter loads them *all* at once when loading the shared library)
* We have the headers in the filesystem, where the `#include` finds them.

At that point I tried to disable the embedding of the headers in the dictionary. But that did not work out nicely either. As you already notice, this requires the include paths to be correct.

But afterwards I had some issues with circular includes (A.h includes B.h, B.h includes A.h). With our current design these circular includes are not trivial to avoid always.

examples/simulation/Tutorial2/macros/compare_seed_value.C
fuhlig1: No I get your point. I think both assumptions are correct in the example.

examples/simulation/Tutorial2/macros/CMakeLists.txt
fuhlig1: Since we had some of such runtime problems with CbmRoot where more than one process accessed the same file I would avoid such situations.

examples/simulation/Tutorial2/macros/CMakeLists.txt
```diff
@@ -12,15 +12,25 @@ GENERATE_ROOT_TEST_SCRIPT(${CMAKE_CURRENT_SOURCE_DIR}/read_digis.C)
...
-         COMMAND ${CMAKE_CURRENT_BINARY_DIR}/run_tutorial2.sh)
+         COMMAND ${CMAKE_CURRENT_BINARY_DIR}/run_tutorial2.sh  10 \"TGeant4\" true 333)
 set_tests_properties(ex_sim_tutorial2 PROPERTIES
   TIMEOUT ${maxTestTime}
   PASS_REGULAR_EXPRESSION "Macro finished successfully"
   FIXTURES_SETUP fixtures.ex_sim_tutorial2_sim
+  RESOURCE_LOCK sim_paramFile
+)
+
+add_test(NAME ex_compare_seed_value
```
dennisklein, line 28:
```suggestion
add_test(NAME ex_sim_tutorial2_compare_seed_value
```

For consistency, all tests in this directory have the prefix `ex_sim_tutorial2`, thx!

examples/simulation/Tutorial2/macros/CMakeLists.txt
```diff
@@ -12,15 +12,25 @@ GENERATE_ROOT_TEST_SCRIPT(${CMAKE_CURRENT_SOURCE_DIR}/read_digis.C)
...
+         COMMAND ${CMAKE_CURRENT_BINARY_DIR}/compare_seed_value.sh \"tutorial2_pions.params_p2.000_t0_n10.root\" 333)
+set_tests_properties(ex_compare_seed_value PROPERTIES
+  TIMEOUT ${maxTestTime}
+  FIXTURES_REQUIRED fixtures.ex_sim_tutorial2_sim
+  RESOURCE_LOCK sim_paramFile
```
ChristianTackeGSI, line 33: Could you please namespace the resource?

```suggestion
  RESOURCE_LOCK ex_sim_tutorial2_paramFile
```

fuhlig1: Done.

examples/simulation/Tutorial2/macros/CMakeLists.txt
fuhlig1: Done.

fuhlig1:
Who should resolve the conversations when an issue was fixed? I would think this should be done by the person who opened the conversation and requested the change. Is this the correct assumption?

examples/simulation/Tutorial2/macros/CMakeLists.txt
dennisklein:
```
CMake Error at examples/simulation/Tutorial2/macros/CMakeLists.txt:30 (set_tests_properties):
  set_tests_properties Can not find test to add properties to:
  ex_sim_tutorial2_compare_seed_value

```

dennisklein: tbh, I would write
```cmake
set(test ex_sim_tutorial2_compare_seed_value)
add_test(NAME ${test} ...)
set_tests_properties(${test} ...)
```
to not have to change the same name in multiple places.

fuhlig1: > ```
> CMake Error at examples/simulation/Tutorial2/macros/CMakeLists.txt:30 (set_tests_properties):
>   set_tests_properties Can not find test to add properties to:
>   ex_sim_tutorial2_compare_seed_value
> ```

Commit before thinking. I did not even test locally and messed up.

fuhlig1: I agree. This is what I normally do, but I didn't want to change it in the existing CMakeLists.txt

ChristianTackeGSI: BTW: Please update the copyright of the CMakeLists.txt as well.

ChristianTackeGSI (commented):
Could you please fix the [formatting errors](https://alfa-ci.gsi.de/blue/organizations/jenkins/FairRootGroup%2FFairRoot/detail/PR-1412/7/pipeline/29)?

examples/simulation/Tutorial2/macros/compare_seed_value.C
```diff
@@ -17,4 +33,8 @@ int compare_seed_value(TString filename, UInt_t initial_seed) {
     std::cout << "Seed value from file : " << seed << std::endl;
     return 1;
   }
+}
+
+int main() {
+  return compare_seed_value("",1);
```
ChristianTackeGSI: Why are you adding this here, if you remove it in a later commit again?

ChristianTackeGSI (commented):
Just one other note.

examples/simulation/Tutorial2/macros/compare_seed_value.C
fuhlig1: Because it was added by accident. This was only meant for testing and should not end up in the PR.

ChristianTackeGSI: Can you please remove it from the relevant commit and possibly squash some commits together?

ChristianTackeGSI:
Things look good to me now!

Can you consider squashing some things together? For example the formatting changes could already be in the commits that actually add/change that relevant code.

examples/simulation/Tutorial2/macros/compare_seed_value.C
fuhlig1: Done.

karabowi (approved):

fuhlig1:
> Things look good to me now!
> 
> Can you consider squashing some things together? For example the formatting changes could already be in the commits that actually add/change that relevant code.

Done.

ChristianTackeGSI (approved):

ChristianTackeGSI:
(I'll press the merge button, once our CI stuff is fixed.)

fuhlig1:
@ChristianTackeGSI,

should be ported to v18.8_patches, maybe even to the v18.6_patches

PR #1413, opened 2023-06-22
Improve Documentation for internal facades

ChristianTackeGSI:
I learned about the [Facade pattern](https://en.wikipedia.org/wiki/Facade_pattern) from @dennisklein and he also suggested to improve the documentation about this.

For the current cases:

From the page:
> * improve the readability and usability of a software library by masking interaction with more complex components behind a single (and often simplified) API

simplified: Yes
Trying to improve the readability (of the callers): Yes

> * provide a context-specific interface to more generic functionality (complete with context-specific input validation)

Yes!
Well, not everything is validated. But see the `\todo`.

> * serve as a launching point for a broader refactor

Maybe.

Note: The Facade pattern talks about objects, while this here is more applied to member functions.

---

Checklist:

### DIFF
--- a/fairroot/base/steer/FairRun.cxx
+++ b/fairroot/base/steer/FairRun.cxx
@@ -227,6 +227,19 @@ TString FairRun::GetUserOutputFileName() const { return fUserOutputFileName; }
 
 void FairRun::AlignGeometry() const { fAlignmentHandler.AlignGeometry(); }
 
+/**
+ * \todo Should we check for `fEvtHeader` being set as well?
+ */
+void FairRun::FillEventHeader()
+{
+    if (!fSource) {
+        LOG(warning) << "FairRun::FillEventHeader: No Source!";
+        return;
+    }
+
+    fSource->FillEventHeader(fEvtHeader);
+}
+
 void FairRun::AddAlignmentMatrices(const std::map<std::string, TGeoHMatrix>& alignmentMatrices, bool invertMatrices)
 {
     fAlignmentHandler.AddAlignmentMatrices(alignmentMatrices, invertMatrices);
--- a/fairroot/base/steer/FairRun.h
+++ b/fairroot/base/steer/FairRun.h
@@ -244,17 +244,19 @@ class FairRun : public TNamed
     std::unique_ptr<FairSource> fSource{};   //!
 
     void AlignGeometry() const;
-    /**
-     * Call FillEventHeader on the source
-     */
-    void FillEventHeader()
-    {
-        if (fSource)
-            fSource->FillEventHeader(fEvtHeader);
-    }
 
     /**
-     * Get the RunId of the Event Header
+     * \brief Internal facade for FairSource::FillEventHeader
+     *
+     * Call \ref FairSource::FillEventHeader with the current header
+     * on the current source. Will perform some checks.
+     */
+    void FillEventHeader();
+
+    /**
+     * \brief Internal facade: Get the RunId of the Event Header
+     *
+     * \todo Should `fEvtHeader` be checked?
      */
     UInt_t GetEvtHeaderRunId() const { return fEvtHeader->GetRunId(); }
 
--- a/fairroot/base/steer/FairRunAna.cxx
+++ b/fairroot/base/steer/FairRunAna.cxx
@@ -59,7 +59,10 @@ void FRA_handler_ctrlc(int)
 //_____________________________________________________________________________
 FairRunAna* FairRunAna::fgRinstance = nullptr;
 //_____________________________________________________________________________
-FairRunAna* FairRunAna::Instance() { return fgRinstance; }
+FairRunAna* FairRunAna::Instance()
+{
+    return fgRinstance;
+}
 //_____________________________________________________________________________
 FairRunAna::FairRunAna()
     : FairRun()
--- a/fairroot/base/steer/FairRunAnaProof.cxx
+++ b/fairroot/base/steer/FairRunAnaProof.cxx
@@ -32,7 +32,10 @@
 
 FairRunAnaProof* FairRunAnaProof::fRAPInstance = nullptr;
 
-FairRunAnaProof* FairRunAnaProof::Instance() { return fRAPInstance; }
+FairRunAnaProof* FairRunAnaProof::Instance()
+{
+    return fRAPInstance;
+}
 
 FairRunAnaProof::FairRunAnaProof(const char* proofName)
     : FairRunAna()


fairroot/base/steer/FairRun.cxx
```diff
@@ -227,6 +227,18 @@ TString FairRun::GetUserOutputFileName() const { return fUserOutputFileName; }
...
+ * \todo Should we check for `fEvtHeader` being set as well?
+ */
+void FairRun::FillEventHeader()
+{
+    if (fSource) {
+        fSource->FillEventHeader(fEvtHeader);
+    } else {
+        LOG(warning) << "FairRun::FillEventHeader: No Source!";
+    }
+}
```
dennisklein, line 240: First, a pure stylistic suggestion:

```suggestion
void FairRun::FillEventHeader()
{
    if (!fSource) {
        LOG(warning) << "FairRun::FillEventHeader: No Source!";
        return;
    }

    fSource->FillEventHeader(fEvtHeader);
}
```

The idea is to not use an explicit block for the else case by default to reduce the amount of blocks. Then there is a choice what to put inside the remaining block - here I followed the idea to keep the "good"/"healthy" code path without blocks and the "exceptional"/"special cases" in the conditional block. At least to me, this style improves readability. What do you think?

---

The next comment may very well be out of scope of this PR, but I make it anyways, while at it; if you like it, we can pick it up later:

Using logs to communicate program logic errors feels off. I would like to propose to promote the warning into a "precondition check". There are two standard approaches to do this that come to mind:

Option A: `throw`
```cpp
void FairRun::FillEventHeader()
{
    if (!fSource) {
        throw std::runtime_error("fSource not set");
    }

    fSource->FillEventHeader(fEvtHeader);
}
```
This could fit, if FairRoot in general wants to make much more use of exceptions.

Option B: assert
```cpp
void FairRun::FillEventHeader()
{
    assert(fSource);

    fSource->FillEventHeader(fEvtHeader);
}
```
This would be the less invasive version, but will mostly be similar to a comment as people usually do not work with debug builds at all. At the end, this may also be subject to a project-wide style discussion on how to express function preconditions.

dennisklein (dismissed):
I leave the merging to you, in case you want to apply the suggested style changes.

fairroot/base/steer/FairRun.cxx
dennisklein: For reference, the CPP Core Guideline recommends adopting `Expect()` (and `Ensure()`), see [I.5](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i5-state-preconditions-if-any), [I.6](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i6-prefer-expects-for-expressing-preconditions), [I.7](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i7-state-postconditions), and [I.8](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i8-prefer-ensures-for-expressing-postconditions)

ChristianTackeGSI: Thanks for reminding me about reducing indented blocks! (It's actually one reason, why *I* like 80 column rules: It forces people to avoid too much indentation.)

And I like the style of "check things and exit/return early".

---

About pre-conditions, etc:

* I don't know, whether fSource not being set is actually a valid scenario. I doubt it, but I am really not sure. So before going to `LOG(error)` or `throw` I wanted to increase awareness a bit.
* Yes, *I* would prefer exceptions for this type of error. People can catch them, handle them, etc. Yes, we probably should discuss this in a broader community.
* I haven't looked deeper into `Expects`, but this sounds like a nice thing!

dennisklein: > I don't know, whether fSource not being set is actually a valid scenario. I doubt it,

This would support that it should be a classic pre-condition. I was looking it from this angle: What can the user do to get rid of this warning?! Probably mostly by a code change. Which is a design time activity, so I would want to consider design time checks/hints. Since we also interpret everything optionally, it becomes more complex what/when design time is and which type of check is the best. but anyways, out of scope of this PR. LGTM

dennisklein (approved):

fairroot/base/steer/FairRun.cxx
YanzhaoW: I have to two suggestions:

1. Always prefer exceptions than abort
   Using exception could lead to stack unwinding, releasing memories allocated in heap before the operating system kills the program. On the other hand, `abort` (or other functions that calls it in the end, such as `assert` and `std::terminate`) doesn't do any stack unwinding and leave the memory cleaning to the operating system. Some operating systems do not do the cleaning job and there could be memory leak. Of course there are many situations when we have no choice but to call abort.

2. Include MS/GSL library into FairSoft?
    I don't know when FairRoot will move to C++20 or even C++23. But some features in GSL are really useful. Contract (expect and ensure) is one of them. gsl::span, gsl::not_null, Range are also pretty useful.

YanzhaoW:
Hi,

This follows the discussion from PR #1254 

> a facade is an [object](https://en.wikipedia.org/wiki/Object_(computer_science)) that serves as a front-facing interface masking more complex underlying or structural code.

To be honest, I don't think `FillEventHeader` is a facade design. As in the quote, a facade must serve as a "front-facing" interface, which means the interface must be open to the public user. However, the member `FillEventHeader` is a protected  and therefore not entirely "front-facing".

> improve the readability and usability of a [software library](https://en.wikipedia.org/wiki/Software_library) by masking interaction with more complex components behind a single (and often simplified) [API](https://en.wikipedia.org/wiki/Application_programming_interface)

In case of FairRoot developers who need to design a derived class from `FairRun`, `FillEventHeader` could be viewed as a "front-facing" interface. But there is another problem: there already exists another "front-facing" interface, called `GetSource()`, which "front-end" developers can also use to fill the event header. Therefore, the complex components are not longer "behind a **single** API".

The signature of this function is also concerning:
```cpp
void FillEventHeader();
```
What is the purpose of this "front-facing" interface that takes nothing and outputs nothing? If everything the function does is just to fill an internal event header into an internal file source, why should a "front-end" user be concerned? Generally speaking, if the function doesn't require anything, it should be closed (declared as private) and executed internally.

ChristianTackeGSI:
> What is the purpose of this "front-facing" interface that takes nothing and outputs nothing? […] if the function doesn't require anything, it should be closed (declared as private) and executed internally.

I don't understand this, really. So you would say, that [`std::vector<>::clear()`](https://en.cppreference.com/w/cpp/container/vector/clear) should be a private member function of `std::vector` and be executed internally in the "right" places?

Anyway: Some of our "facades" (maybe let's call them "convenience helpers"?) can be considered part of a bigger refactoring (which also is listed as a point on the facade wikipedia page). Having very concise calls in the main control flow makes it a lot easier to spot refactoring options. Maybe we learn that ReadEvent is always followed by `FillEventHeader()`? Then you're right and we should put the FillEventHeader into ReadEvent. And as this is all purely internal to FairRoot, it does not concern any users ("customers") of FairRoot.

YanzhaoW:
> I don't understand this, really. So you would say, that [std::vector<>::clear()](https://en.cppreference.com/w/cpp/container/vector/clear) should be a private member function of std::vector and be executed internally in the "right" places?

Sorry, I was too general. Yes, you are right. User should be concerned about functions like `reset()` or `clear()`, which do change the internal data.

For me, I also kind of think `GetSource()->FillEventHeader(eventHeader)` contains more information and more readable than `FillEventHeader()`. Both are one line of code. But anyway, let's not waste more time on this subtle thing :D. About my PR #1254, should I change it back to `FillEventHeader()`?

ChristianTackeGSI:
> About my PR #1254, should I change it back to `FillEventHeader()`?

Yes. Please!

I think, it will also reduce the size of your patchset (because you like don't need to change so many places).

PR #1419, opened 2023-06-30
Add compare macro

fuhlig1:
Add a macro which allows to compare the cbmsim tree of two different ROOT macros. The added script does this comparison for all relevant ROOT files produced by the test suite. This allows to easily compare the results before and after a change of FairRoot to see if there are unexpected changes.
---

Checklist:


examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+        auto hist1 = (TH1F*)gPad->GetPrimitive("hist1");
+        int entries2{0};
+        float low2{0.};
+        float high2{0.};
+        int nBins2{0};
+        if (hist1) {
+            entries2 = hist1->GetEntries();
+            nBins2 = hist1->GetNbinsX();
+            low2 = hist1->GetXaxis()->GetXmin();
+            high2 = hist1->GetXaxis()->GetXmax();
```
ChristianTackeGSI, line 115: data from hist**1** ends up in variable**2**?

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+    //  std::cout << "Empty leaves:                     " << numEmptyLeaves << std::endl;
+    std::cout << "Leaves with identical entries:    " << numIdenticalEntries << std::endl;
+    std::cout << "Leaves with identical histograms: " << numIdenticalHistograms << std::endl;
+    std::cout << "Leaves with kolmo histograms:     " << numKolmogorovHistograms << std::endl;
+    std::cout << "Test passed. All leaves of all branches are exactly identical." << std::endl;
+
+    return 0;
+}
+
+Bool_t CheckEntriesIdentical(TH1* compHist, std::stringstream& outstream)
```
ChristianTackeGSI, line 211:
```suggestion
Bool_t CheckEntriesIdentical(TH1& compHist, std::stringstream& outstream)
```

You're assuming that `compHist != nullptr` anyway. So let's use a reference?

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+    } else {
+        outstream << "Entries are identical." << std::endl;
+        outstream << "**********************" << std::endl;
+        return kTRUE;
+    }
+}
+
+Bool_t CheckHistogramIdentical(TH1* origHist, TH1* newHist, std::stringstream& outstream)
+{
+    if (origHist && newHist) {
```
ChristianTackeGSI, line 225: Are there cases, where any of these two can be nullptr?

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+Bool_t CheckHistogramKolmogorov(TH1* origHist, TH1* newHist, std::stringstream& outstream)
+{
+    Double_t kolmo = origHist->KolmogorovTest(newHist);
+
+    outstream << "Result of Kolmogorov test: " << kolmo << endl;
+    if (kolmo > 0.99) {
+        return kTRUE;
+    }
+
+    return kFALSE;
```
ChristianTackeGSI, line 248: Maybe that's simpler?

```suggestion
    return (kolmo > 0.99);
```

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+                if (bla2->GetEntriesFast() == 4) {
+                    TString _branch = ((TObjString*)(bla2->At(2)))->GetString();
+                    TString _leaf = ((TObjString*)(bla2->At(3)))->GetString();
+                    ListOfLeaves.emplace_back(_branch + "." + _leaf);
+                }
+            }
+
+            if (mystring.Contains("GeoTracks")) {
+                continue;
+            }
```
ChristianTackeGSI, line 299: `continue;` at the end of a loop?

```suggestion
```

ChristianTackeGSI (commented):
First round of comments.

Could we please get a test for this tooling somehow?

Maybe we could integrate it in some other test by running the simulation with two seeds and then compare the results? (which should hopefully be quite the same?)

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+            numIdenticalHistograms++;
+            outstream << "**********************" << std::endl;
+            hist1->Clear();
+            hist->Clear();
+            htemp->Clear();
+            continue;
+        } else {
+            outstream << "CHecking Kolmogorov" << endl;
+            leafIsIdentical = CheckHistogramKolmogorov(htemp, hist1, outstream);
+        }
```
dennisklein, line 171:
```suggestion
        
        outstream << "Checking Kolmogorov" << endl;
        auto const kolmo = origHist->KolmogorovTest(newHist);
        outstream << "Result of Kolmogorov test: " << kolmo << endl;
        leafIsIdentical = kolmo > 0.99;
```

* no else block needed -> improve readability
* remove `CheckHistogramKolomogorov()` helper, it is only called once and does not add anything -> netto -12 lines of code

dennisklein (changes requested):

fuhlig1:
The test will be the next step. First I want to get the macro and the script in which are the prerequisites for the test.
I will look for the comments probably only next week.

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+            if (origHist->GetBinContent(x) != newHist->GetBinContent(x)) {
+                return kFALSE;
+            }
+        }
+        outstream << "Histograms are identical." << std::endl;
+        outstream << "**********************" << std::endl;
+        return kTRUE;
+    }
+    return kFALSE;
+}
```
dennisklein, line 237:
```suggestion
bool operator==(TH1 const& lhs, TH1X const& rhs)
{
    for (int x = 0; x < lhs.GetNbinsX() + 1; ++x) {
        if (lhs.GetBinContent(x) != rhs.GetBinContent(x)) {
            return false;
        }
    }
    return true;
}
```

* Remove `CheckHistogramIdentical()` and check my corresponding comment at the call site

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+            numIdenticalEntries++;
+            outstream << "**********************" << std::endl;
+            hist1->Clear();
+            hist->Clear();
+            htemp->Clear();
+            continue;
+        } else {
+            outstream << "CHecking for identical histograms" << endl;
+            leafIsIdentical = CheckHistogramIdentical(htemp, hist1, outstream);
+        }
```
dennisklein, line 159:
```suggestion
        
        outstream << "Checking for identical histograms" << endl;
        leafIsIdentical = (*htemp == *hist1);
        if (leafIsIdentical) {
            outstream << "Histograms are identical." << std::endl;
            outstream << "**********************" << std::endl;
        }
```

* no else block needed -> improve readability
* let's overload operator `==` to implement equality check -> idiomatic C++, readability, netto -11 lines of code

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+        // If the content is differnt one gets a distribution which is
+        // detected.
+        outstream << "Comparing " << leafName << " and " << leafName1 << std::endl;
+        TString command = leafName + "-" + leafName1 + ">>hist(20, -10.,10.)";
+        tree1->Draw(command);
+        auto hist = (TH1F*)gPad->GetPrimitive("hist");
+        numTestedLeaves++;
+
+        // Check if the entries in the tree are identical
+        Bool_t leafIsIdentical = CheckEntriesIdentical(hist, outstream);
```
dennisklein, line 144:
```suggestion

        outstream << "Checking for identical entries" << endl;
        bool leafIsIdentical =    (TMath::Abs(hist->GetMean()) > 0.000001 && TMath::Abs(hist->GetRMS()) > 0.000001)
                               || (0 != hist->GetBinContent(0))
                               || (0 != hist->GetBinContent(hist->GetNbinsX() + 1));
        if (leafIsIdentical) {
            outstream << "Entries are identical." << std::endl;
            outstream << "**********************" << std::endl;
        }
```

* Remove `CheckEntriesIdentical()` -> netto -7 lines of code

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+            outstream << "**********************" << std::endl;
+            outstream << "Entries: " << hist->GetEntries() << std::endl;
+            outstream << "Mean: " << hist->GetMean() << std::endl;
+            outstream << "RMS: " << hist->GetRMS() << std::endl;
+            outstream << "Underflow: " << hist->GetBinContent(0) << std::endl;
+            outstream << "Overflow: " << hist->GetBinContent(hist->GetNbinsX() + 1) << std::endl;
+            outstream << "**********************" << std::endl;
+            okay = kFALSE;
+            numFailedLeaves++;
+        }
```
dennisklein, line 191:
```suggestion

        outstream << "Data are different" << std::endl;
        outstream << "**********************" << std::endl;
        outstream << "Entries: " << hist->GetEntries() << std::endl;
        outstream << "Mean: " << hist->GetMean() << std::endl;
        outstream << "RMS: " << hist->GetRMS() << std::endl;
        outstream << "Underflow: " << hist->GetBinContent(0) << std::endl;
        outstream << "Overflow: " << hist->GetBinContent(hist->GetNbinsX() + 1) << std::endl;
        outstream << "**********************" << std::endl;
        okay = kFALSE;
        numFailedLeaves++;
```

* no else block needed -> readability

examples/scripts/TreeCompareAuto.C
dennisklein: I would also not re-use the `leafIsIdentical` variable, but have three separate ones:
* `bool const entriesIdentical`
* `bool const histogramsIdentical`
* `bool const kolmogorovIdentical`
or similar. Less mutable variables improve readability for me.

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+
+    outstream << "Result of Kolmogorov test: " << kolmo << endl;
+    if (kolmo > 0.99) {
+        return kTRUE;
+    }
+
+    return kFALSE;
+}
+
+std::vector<std::pair<TString, TString>> GetLeafNames(TTree* cbmsim)
```
dennisklein, line 251: `std::pair<TString, TString>` imho this needs better docs (e.g. by aliasing `std::pair<TString, TString>` or simply by a comment). To me, it is not obvious why the names of tree leaf nodes are described with a pair of strings. Two names per leaf?

dennisklein: and then again, why raw pointer and not just a reference - I guess even a const-ref will be fine here?

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+                continue;
+            }
+        }
+    }
+
+    std::vector<std::pair<TString, TString>> leaves;
+    for (auto const element : ListOfLeaves) {
+        TString nameTree2 = "tree2." + element;
+        leaves.emplace_back(make_pair(element, nameTree2));
+    }
```
dennisklein, line 307: This logic does not fit the abstraction level of this function. See also my [comment at the function signature](https://github.com/FairRootGroup/FairRoot/pull/1419/files#r1247860985). How about moving this out of the function just after the call?

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+    // Add the output tree from the file to compare as friend to the tree
+    // of the original file. This allows to access a data member of the
+    // original file by e.g. StsHit.fX and the data element of the second tree
+    // by tree2.StsHit.fX
+    tree1->AddFriend(tree2, "tree2");
+
+    // Define pairs of data members to compare. One from each tree.
+    // This allows to compare them also if names or the structure of
+    // the classses change.
+    std::vector<std::pair<TString, TString>> leaves = GetLeafNames(tree1);
```
dennisklein, line 73:
```suggestion
    auto const leafNames = GetLeafNames(tree1);

    // Define pairs of data members to compare. One from each tree.
    // This allows to compare them also if names or the structure of
    // the classes change.    
    std::vector<std::pair<TString, TString>> leaves;
    for (auto const leafName : leafNames) {
        TString nameTree2 = "tree2." + leafName;
        leaves.emplace_back(make_pair(leafName, nameTree2));
    }
```

* See [my comment down below](https://github.com/FairRootGroup/FairRoot/pull/1419/files#r1247871827)

examples/scripts/checkfiles.sh
```diff
@@ -0,0 +1,73 @@
...
+
+#MQ/serialization/data_io/testinput1.root
+#MQ/serialization/data_io/outputEx1Test.root
+#MQ/serialization/data_io/testinput2.root
+#MQ/serialization/data_io/outputEx2Test.root
+#MQ/pixelDetector/macros/pixel_TGeant4.mc.root
+#MQ/pixelDetector/macros/pixel_TGeant4.digi.root
+#MQ/pixelDetector/macros/pixel_TGeant4.digiToBin.root
+#MQ/pixelDetector/macros/MQ.pixel_TGeant4.hits.root
+#MQ/pixelSimSplit/run/MQ.simulation_TGeant4.data.root
```
dennisklein, line 58: We should generate this list with CMake, having it hardcoded like this will be a nightmare to maintain.

examples/scripts/TreeCompareAuto.C
dennisklein: Please make a const variable for `"tree2"` and use it here and also [in line 68](https://github.com/FairRootGroup/FairRoot/pull/1419/files/4d8743c0ab0d89991abf8d45f633a2fc51e2e020#diff-67a152a49632b90f589b89a3b5439390f14ea2be3fad4d409e6b8dc364ddb731R68).

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+    if (fileName1.IsNull() || fileName2.IsNull()) {
+        cout << "Filenames are not defined" << endl;
+        exit(42);
+    }
+
+    // Get the output tree from the original file
+    TFile* file1 = TFile::Open(fileName1, "READ");
+    if (nullptr == file1)
+        return 42;
+    TTree* tree1 = (TTree*)file1->Get("cbmsim");
```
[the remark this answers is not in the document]
fuhlig1: Done.

examples/scripts/TreeCompareAuto.C
fuhlig1: Fixed. The second name can even be generated on the fly.

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,309 @@
...
+    // by tree2.StsHit.fX
+    tree1->AddFriend(tree2, "tree2");
+
+    // Define pairs of data members to compare. One from each tree.
+    // This allows to compare them also if names or the structure of
+    // the classses change.
+    std::vector<std::pair<TString, TString>> leaves = GetLeafNames(tree1);
+
+    std::stringstream outstream;
+    Bool_t okay{kTRUE};
```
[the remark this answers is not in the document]
fuhlig1: fixed

examples/scripts/TreeCompareAuto.C
fuhlig1: fixed

examples/scripts/TreeCompareAuto.C
fuhlig1: fixed

examples/scripts/TreeCompareAuto.C
fuhlig1: Removed the else blocks everywhere

examples/scripts/TreeCompareAuto.C
fuhlig1: I do not remove the small functions since for me they make the code more readable. I only have to look inside if I want to know how the comparison is done.

examples/scripts/TreeCompareAuto.C
fuhlig1: removed the else block. Did not remove the helper function.

examples/scripts/TreeCompareAuto.C
fuhlig1: removed the else block

examples/scripts/TreeCompareAuto.C
fuhlig1: No. Fixed.

examples/scripts/TreeCompareAuto.C
fuhlig1: Fixed. Leftover.

examples/scripts/TreeCompareAuto.C
fuhlig1: Done.

examples/scripts/TreeCompareAuto.C
fuhlig1: Done

examples/scripts/TreeCompareAuto.C
fuhlig1: Done

examples/scripts/checkfiles.sh
fuhlig1: I agree but I am not sure if this will be so easy. For the time being it would be good if we could keep the hard coded list.

fuhlig1:
@ChristianTackeGSI, @dennisklein,

thanks for all the suggestions. I hope I addressed them now all.

examples/scripts/TreeCompareAuto.C
ChristianTackeGSI: I think, some helper functions are okay, if they reduce the size/complexity of the main thing.

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,293 @@
...
+        exit(42);
+    }
+
+    // Get the output tree from the original file
+    std::unique_ptr<TFile> file1{TFile::Open(fileName1, "READ")};
+    if (!file1) {
+        std::cout << "Could not open file " << fileName1 << std::endl;
+        return 42;
+    }
+    std::unique_ptr<TTree> tree1{file1->Get<TTree>("cbmsim")};
```
dennisklein, line 56: Hardcoding `"cbmsim"` is still not so nice, imho. Should we use `FairRootManager::GetTreeName()` here? (This would also need some cleanup in some examples as well, which hardcode this name as well)

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,293 @@
...
+    if (!tree2) {
+        std::cout << "File " << fileName2 << " does not have the cbmsim tree." << std::endl;
+        return 42;
+    }
+
+    // Add the output tree from the file to compare as friend to the tree
+    // of the original file. This allows to access a data member of the
+    // original file by e.g. StsHit.fX and the data element of the second tree
+    // by tree2.StsHit.fX
+    tree1->AddFriend(tree2.get(), "tree2");
```
dennisklein, line 78:
```suggestion
    auto const friendName = "tree2"
    tree1->AddFriend(tree2.get(), friendName);
```

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,293 @@
...
+    int numEmptyLeaves{0};
+    int numLeaves{0};
+    int numFailedLeaves{0};
+    int numIdenticalEntries{0};
+    int numIdenticalHistograms{0};
+    int numKolmogorovHistograms{0};
+
+    for (auto leaf : leaves) {
+        TString leafName = leaf;
+        TString leafName1 = "tree2." + leafName;
```
dennisklein, line 103:
```suggestion
        TString leafName1 = friendName + "." + leafName;
```

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,293 @@
...
+    // Get the leaf names for all leaves which should be compared
+    // from the first input tree
+    auto leaves = GetLeafNames(*tree1.get());
+
+    if (0 == leaves.size()) {
+        std::cout << "Test passed. Tree does not contain nay branches which must be checked" << std::endl;
+        return 0;
+    }
+
+    TCanvas c1;
```
dennisklein, line 89:
```suggestion
```
and see following comments further down

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,293 @@
...
+        float low{0.};
+        float high{0.};
+        int nBins{0};
+        auto htemp = static_cast<TH1F*>(gPad->GetPrimitive("htemp"));
+        if (htemp) {
+            entries = htemp->GetEntries();
+            nBins = htemp->GetNbinsX();
+            low = htemp->GetXaxis()->GetXmin();
+            high = htemp->GetXaxis()->GetXmax();
+        }
```
dennisklein, line 117:
```suggestion
        TH1F htemp;
        long long entries{tree1->GetEntries()};
        float value = 0;
        tree1->SetBranchAddress(leafName, &value);
        for (long long i = 0; i < entries; ++i) {
            tree1->GetEntry(i);
            htemp->Fill(value);
        }
        float low = htemp.GetXaxis()->GetXmin();
        float high = htemp.GetXaxis()->GetXmax();
        int nBins = htemp.GetNbinsX();
```

Why not something simple like this (have not tested the above code, but there should be some code similar to the above)?

What is the benefit to go through all this `Draw()` logic and global referencing via `gPad`?

dennisklein: For the `hist` case a bit further down, one could use a `TFormula`, but since it is just a subtraction, not even needed?

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,293 @@
...
+
+    // Add the output tree from the file to compare as friend to the tree
+    // of the original file. This allows to access a data member of the
+    // original file by e.g. StsHit.fX and the data element of the second tree
+    // by tree2.StsHit.fX
+    tree1->AddFriend(tree2.get(), "tree2");
+
+    // Get the leaf names for all leaves which should be compared
+    // from the first input tree
+    auto leaves = GetLeafNames(*tree1.get());
```
dennisklein, line 82:
```suggestion
    auto leaves = GetLeafNames(*tree1);
```

See [`unique_ptr::operator*`](https://en.cppreference.com/w/cpp/memory/unique_ptr/operator*)

examples/scripts/TreeCompareAuto.C
dennisklein: Thinking about it, having a dependency on `FairRootManager` here is or even just the config file is not a good idea after all. Could we perhaps list the tree names in the file and just pick the first one?

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,293 @@
...
+    TObjArray* bla1 = cbmsim.GetListOfLeaves();
+    TIter myiter1(bla1);
+    TBranch* branch;
+    while ((branch = (TBranch*)myiter1.Next())) {
+        TString mystring = branch->GetName();
+        //      cout << "Branch Name: " << mystring << endl;
+
+        // Generate leaf names for points, digis and hits
+        if (mystring.Contains("Point") || mystring.Contains("Digi") || mystring.Contains("Hit")) {
+            TObjArray* bla2 = mystring.Tokenize(".");
```
dennisklein, line 260: Regarding the chosen variable names:
* `bla1`
* `myiter1`
* `mystring`
* `bla2`

Please see [S.2](https://github.com/FairRootGroup/FairRoot/blob/dev/CONTRIBUTING.md#s2-prefer-variable-names-describing-what-they-contain)

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,293 @@
...
+            if (bla2->GetEntriesFast() == 4) {
+                TString _branch = ((TObjString*)(bla2->At(2)))->GetString();
```
dennisklein, line 262: C-Style casts are not acceptable. See [CPPGC::Pro.type](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#prosafety-type-safety-profile) and [CPPCG::C.146](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-dynamic_cast)

examples/scripts/checkfiles.sh
dennisklein: An idea would be to glob for `${dir1}/**/*.root` and then iterate this list of files, if such a root file also exists in `$dir2`. We may have to add then another sanity check to the `TreeCompareAuto.C` macro that it will skip non-FairRoot root files, if there were any.

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,293 @@
...
+
+std::vector<TString> GetLeafNames(TTree& cbmsim)
+{
+
+    std::vector<TString> ListOfLeaves;
+
+    TObjArray* bla1 = cbmsim.GetListOfLeaves();
+    TIter myiter1(bla1);
+    TBranch* branch;
+    while ((branch = (TBranch*)myiter1.Next())) {
```
dennisklein, line 254: Here you actually have a classic fail with C-style cast. This container does not contain `TBranch` objects at all, but `TLeaf`s ... The only reason this went unnoticed, is that you use an interface of a common base class `TNamed`!

examples/scripts/TreeCompareAuto.C
dennisklein, line 1: At first I wanted to just re-write the TString/Tokenize logic, but then ended up overhauling the whole `GetLeafNames()` function.

```diff
--- a/examples/scripts/TreeCompareAuto.C
+++ b/examples/scripts/TreeCompareAuto.C
@@ -24,19 +24,23 @@

 #include <RtypesCore.h>
 #include <TBranch.h>
+#include <TCollection.h>   // for TRangeDynCast
 #include <TFile.h>
 #include <TH1.h>
+#include <TLeaf.h>
 #include <TMath.h>
-#include <TObjArray.h>
 #include <TString.h>
 #include <TTree.h>
+#include <functional>   // for std::logical_or
 #include <iostream>
+#include <numeric>   // for std::transform_reduce
 #include <sstream>
 #include <string>
+#include <string_view>
 #include <utility>
 #include <vector>

-std::vector<TString> GetLeafNames(TTree&);
+std::vector<std::string> GetLeafNames(TTree&);

 bool operator==(TH1 const&, TH1 const&);

@@ -243,51 +247,72 @@ bool operator==(TH1 const& lhs, TH1 const& rhs)
     return true;
 }

-std::vector<TString> GetLeafNames(TTree& cbmsim)
+std::vector<std::string_view> split(std::string_view sv, char by)
 {
+    std::vector<std::string_view> components;
+    std::string_view::size_type first{0};
+    std::string_view::size_type last{sv.find(by, first)};
+    while (last != std::string_view::npos) {
+        components.emplace_back(sv.substr(first, last - first));
+        first = last + 1;
+        last = sv.find(by, first);
+    }
+    components.emplace_back(sv.substr(first, std::string_view::npos));
+    return components;
+}
+
+// to be replaced by std::string::contains in C++23
+// https://en.cppreference.com/w/cpp/string/basic_string/contains
+bool contains(std::string_view sv, std::string_view what)
+{
+    return sv.find(what) != std::string_view::npos;
+}

-    std::vector<TString> ListOfLeaves;
-
-    TObjArray* bla1 = cbmsim.GetListOfLeaves();
-    TIter myiter1(bla1);
-    TBranch* branch;
-    while ((branch = (TBranch*)myiter1.Next())) {
-        TString mystring = branch->GetName();
-        //      cout << "Branch Name: " << mystring << endl;
-
-        // Generate leaf names for points, digis and hits
-        if (mystring.Contains("Point") || mystring.Contains("Digi") || mystring.Contains("Hit")) {
-            TObjArray* bla2 = mystring.Tokenize(".");
-            if (bla2->GetEntriesFast() == 4) {
-                TString _branch = ((TObjString*)(bla2->At(2)))->GetString();
-                TString _leaf = ((TObjString*)(bla2->At(3)))->GetString();
-                if (_leaf.EqualTo("fLink")) {
-                    TString name = _branch + "." + _leaf + ".fLinks";
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf + ".fLinks.fFile");
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf + ".fLinks.fType");
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf + ".fLinks.fEntry");
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf + ".fLinks.fIndex");
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf + ".fLinks.fWeight");
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf + ".fEntryNr.fFile");
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf + ".fEntryNr.fType");
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf + ".fEntryNr.fEntry");
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf + ".fEntryNr.fIndex");
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf + ".fEntryNr.fWeight");
-                } else {
-                    ListOfLeaves.emplace_back(_branch + "." + _leaf);
-                }
-            }
+std::vector<std::string> GetLeafNames(TTree& simTree)
+{
+    std::vector<std::string> leaves;
+
+    for (auto leaf : TRangeDynCast<TLeaf>(simTree.GetListOfLeaves())) {
+        if (!leaf) {
+            continue;
         }

-        if (mystring.Contains("MCTrack")) {
-            TObjArray* bla2 = mystring.Tokenize(".");
-            if (bla2->GetEntriesFast() == 4) {
-                TString _branch = ((TObjString*)(bla2->At(2)))->GetString();
-                TString _leaf = ((TObjString*)(bla2->At(3)))->GetString();
-                ListOfLeaves.emplace_back(_branch + "." + _leaf);
-            }
+        std::string fullName = leaf->GetName();
+
+        auto const keywords = {"Point", "Digi", "Hit", "MCTrack"};
+        auto const match = std::transform_reduce(
+            std::cbegin(keywords), std::cend(keywords), false, std::logical_or(), [&](auto keyword) {
+                return contains(fullName, keyword);
+            });
+        if (!match) {
+            continue;
+        }
+
+        auto const parts = split(fullName, '.');
+        if (parts.size() != 4) {
+            continue;
         }
+
+        std::string const branchName{parts[2]};
+        std::string const leafName{parts[3]};
+        std::string const name{branchName + "." + leafName};
+
+        if (leafName == "fLink" && !contains(fullName, "MCTrack")) {
+            leaves.emplace_back(name + ".fLinks.fFile");
+            leaves.emplace_back(name + ".fLinks.fType");
+            leaves.emplace_back(name + ".fLinks.fEntry");
+            leaves.emplace_back(name + ".fLinks.fIndex");
+            leaves.emplace_back(name + ".fLinks.fWeight");
+            leaves.emplace_back(name + ".fEntryNr.fFile");
+            leaves.emplace_back(name + ".fEntryNr.fType");
+            leaves.emplace_back(name + ".fEntryNr.fEntry");
+            leaves.emplace_back(name + ".fEntryNr.fIndex");
+            leaves.emplace_back(name + ".fEntryNr.fWeight");
+            continue;
+        }
+
+        leaves.emplace_back(name);
     }

-    return ListOfLeaves;
+    return leaves;
 }
```

or to read it in one flow:

```cpp
std::vector<std::string_view> split(std::string_view sv, char by)
{
    std::vector<std::string_view> components;
    std::string_view::size_type first{0};
    std::string_view::size_type last{sv.find(by, first)};
    while (last != std::string_view::npos) {
        components.emplace_back(sv.substr(first, last - first));
        first = last + 1;
        last = sv.find(by, first);
    }
    components.emplace_back(sv.substr(first, std::string_view::npos));
    return components;
}

// to be replaced by std::string::contains in C++23
// https://en.cppreference.com/w/cpp/string/basic_string/contains
bool contains(std::string_view sv, std::string_view what)
{
    return sv.find(what) != std::string_view::npos;
}

std::vector<std::string> GetLeafNames(TTree& simTree)
{
    std::vector<std::string> leaves;

    for (auto leaf : TRangeDynCast<TLeaf>(simTree.GetListOfLeaves())) {
        if (!leaf) {
            continue;
        }

        std::string fullName = leaf->GetName();

        auto const keywords = {"Point", "Digi", "Hit", "MCTrack"};
        auto const match = std::transform_reduce(
            std::cbegin(keywords), std::cend(keywords), false, std::logical_or(), [&](auto keyword) {
                return contains(fullName, keyword);
            });
        if (!match) {
            continue;
        }

        auto const parts = split(fullName, '.');
        if (parts.size() != 4) {
            continue;
        }

        std::string const branchName{parts[2]};
        std::string const leafName{parts[3]};
        std::string const name{branchName + "." + leafName};

        if (leafName == "fLink" && !contains(fullName, "MCTrack")) {
            leaves.emplace_back(name + ".fLinks.fFile");
            leaves.emplace_back(name + ".fLinks.fType");
            leaves.emplace_back(name + ".fLinks.fEntry");
            leaves.emplace_back(name + ".fLinks.fIndex");
            leaves.emplace_back(name + ".fLinks.fWeight");
            leaves.emplace_back(name + ".fEntryNr.fFile");
            leaves.emplace_back(name + ".fEntryNr.fType");
            leaves.emplace_back(name + ".fEntryNr.fEntry");
            leaves.emplace_back(name + ".fEntryNr.fIndex");
            leaves.emplace_back(name + ".fEntryNr.fWeight");
            continue;
        }

        leaves.emplace_back(name);
    }

    return leaves;
}
```

dennisklein: forget the `std::transform_reduce`, it's easier:

```cpp
auto const keywords = {"Point", "Digi", "Hit", "MCTrack"};
auto const match = std::any_of(std::cbegin(keywords), std::cend(keywords), [&](auto keyword) {
    return contains(fullName, keyword);
});
```

vs

```cpp
auto const match =    contains(fullName, "Point")
                   || contains(fullName, "Digi")
                   || contains(fullName, "Hit")
                   || contains(fullName, "MCTrack");
```

That is the question :P hmmm

dennisklein: the `auto const keywords = {"Point", "Digi", "Hit", "MCTrack"};` could even be moved outside of the loop body

examples/scripts/TreeCompareAuto.C
rbx: @fuhlig1 
`bla1` -> `list`/`listOfLeaves`
`myiter1` -> `iter`/`listIter`/`leavesIter`
`mystring` -> `branchName`
`bla2` -> `branchNameTokens`

examples/scripts/checkfiles.sh
fuhlig1: The problem is that not all ROOT files should be tested. Not all off them contain the tree. That means that we have to filter the list afterwards to get only the list of files we want. In my opinion this isn't better than having a fixed list.

One solution would be to provide some benchmark files you want to compare and loop over these files and compare them with the newly produced ones. The downside of this is that we have to provide the benchmark files somehow.

examples/scripts/TreeCompareAuto.C
fuhlig1: Fixed. I loop over all TKeys in the file and extract the name of the TTree from the object, so no explicit tree name is needed any longer. The only requirement is that the names of the trees are identical in both files, otherwise the test fails-

examples/scripts/TreeCompareAuto.C
fuhlig1: Fixed.

examples/scripts/TreeCompareAuto.C
fuhlig1: Fixed.

examples/scripts/TreeCompareAuto.C
fuhlig1: This is needed to suppress a warning/info from ROOT. I add a comment and renamed the variable.

examples/scripts/TreeCompareAuto.C
fuhlig1: I was sure I tested it and it didn't work, but you are right it works as expected. Fixed.

examples/scripts/TreeCompareAuto.C
fuhlig1: I think this will not work since the value can be anything and must not be a float.

examples/scripts/TreeCompareAuto.C
fuhlig1: Fixed.

examples/scripts/TreeCompareAuto.C
fuhlig1: Fixed. Changed several ocurence but overlooked them.

examples/scripts/checkfiles.sh
dennisklein: Another approach would be:

* Add a CMake list variable `FAIRROOT_SIMTREE_FILES_FOR_TESTING`
* Append file names to it in each of the local `CMakeLists.txt` files from each example/test the file belongs to
* Then `configure_file()` this `checkfiles.sh` script in a way that the contents of the CMake variable is added

Then the "authority" of naming relevant root files stays local to the individual example/test/lib. What do you think?

examples/scripts/TreeCompareAuto.C
fuhlig1: You are right. Already the function name should have told me that.

examples/scripts/TreeCompareAuto.C
dennisklein: See also my [rewrite study for inspiration](https://github.com/FairRootGroup/FairRoot/pull/1419#discussion_r1253744499). Imho, we should avoid all those variable and just use this syntax:

```cpp
for (auto leaf : TRangeDynCast<TLeaf>(simTree.GetListOfLeaves())) {
    if (!leaf) {
        continue;
    }
    
    // ...
}
```

* This `nullptr` check is still ugly and error-prone (people will forget it), but that is a consequence of living in the reference semantics world which is a fundamental choice of FairRoot/ROOT...
* The second ugly aspect that we cannot use `auto const leaf` because `GetName()` has no `const`-correct overload.

But besides those, I think this is much safer and more readable than having 3 extra named boilerplate variables for just facilitating a read loop.

examples/scripts/TreeCompareAuto.C
dennisklein: Then why do you cast it to a `TH1F*` where the `F` stands for `float`? (Also, how are `low` and `high` then supposed to work if it is not a float?)

examples/scripts/TreeCompareAuto.C
fuhlig1: I changed the function. Looks good. I still have to check why I now get one leaf less than with the original version.

examples/scripts/TreeCompareAuto.C
dennisklein: > (Also, how are low and high then supposed to work if it is not a float?)

Btw, [`TAxis::GetXmax()`](https://root.cern.ch/doc/v624/classTAxis.html#a5ae359f6f1fbbb078cd425a383582b11) returns a `double`?!

dennisklein: Also check [`Hist.Precision.1D`](https://github.com/root-project/root/blob/f83a40912dbae9b42e7df1a89774a87ebbc7a8fb/README/ReleaseNotes/v610/index.md?plain=1#L124-L131). If I understand this correctly, `TTree::Draw()` will always produce single precision histograms, unless you change it globally in your ROOT setting?!

dennisklein: In our case, it should rather depend on the data in the root file, should it not?

fuhlig1: You can pass any number to TH1F::Fill() which will find the proper bin to increase. The type of the histogram only defines the maximum value which can be stored per bin. TH1C can only store values less than 128 per bin. TH1F stores the bin content in a float value.

These are the values you have to pass to the constructor which is the same for all types of one dimensional histograms

[in] | name | name of histogram (avoid blanks) | const char*
[in] | title | histogram title. | const char*
[in] | nbins | number of bins | int
[in] | xlow | low edge of first bin | double
[in] | xup | upper edge of last bin (not included in last bin) | double

TAxis::GetMax returns the value value of the upper edge of the last bin which is a double.

dennisklein: Ok, but then the common type to fill the histogram with is just `double`? So, read the tree content into doubles?

dennisklein: Or as a question: Which `H1F` API is used by `TTree::Draw` to supply add its stored values to the histogram?

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,344 @@
...
+        // When executing the draw command "Leaf1 - Leaf2" the subtraction is
+        // executed entry by entry. If the content of the class members are
+        // identical the result is a histogram with a delta function at 0
+        // If the content is differnt one gets a distribution which is
+        // detected.
+        outstream << "Comparing " << leafName << " and " << leafName1 << std::endl;
+
+        TString command = leafName + "-" + leafName1 + ">>hist(20, -10.,10.)";
+        tree1->Draw(command);
+        auto hist = static_cast<TH1F*>(gPad->GetPrimitive("hist"));
```
ChristianTackeGSI, line 187: Can we please call this histogram `hdifferences` or something like that?

examples/scripts/TreeCompareAuto.C
dennisklein: And whatever the code looks like, my points are:
1. Why do we need to involve a `TCanvas` object, a `TPad` object and a function called `Draw` - all clearly graphics related entities - in a script that does not want to produce any graphics? Especially, if we have to create a random `TCanvas` object in the middle of it to silence warnings? Don't you agree that this is a terrible program to write, read and maintain?
2. Can we produce the histogram object not through the indirection of the global object `gPad`?

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,344 @@
...
+        return 42;
+    }
+
+    // Find a tree in the file
+    TList* keylist = file1->GetListOfKeys();
+    TKey* key;
+    TIter keyIterator(keylist);
+    TString treeName;
+    int numTreeInFile{0};
+    while ((key = static_cast<TKey*>(keyIterator()))) {
```
ChristianTackeGSI, line 67: Can this be replaced by a TRangeDynCast?
Something like this maybe?
```c++
for (auto key : TRangeDynCast<TKey const>(keylist)) {
    if (!key) {
        continue;
    }
    …
}
```

ChristianTackeGSI (commented):
I didn't follow all the discussions. So please excuse, if I duplicate anything! (In that case, just write a short "dupe" and press the resolve button!)

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,344 @@
...
+#include <string>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+std::vector<std::string> GetLeafNames(TTree&);
+
+bool operator==(TH1 const&, TH1 const&);
+
+int TreeCompareAuto(TString fileName1 = "", TString fileName2 = "")
```
dennisklein, line 47:
```suggestion
int TreeCompareAuto(TString fileName1, TString fileName2)
```

We anyways expect non-empty arguments.

ChristianTackeGSI:
Hi @fuhlig1 could you consider adding a compile check for the macro like in #1421?

examples/scripts/TreeCompareAuto.C
```diff
@@ -0,0 +1,344 @@
...
+    TKey* key;
+    TIter keyIterator(keylist);
+    TString treeName;
+    int numTreeInFile{0};
+    while ((key = static_cast<TKey*>(keyIterator()))) {
+        if (key->ReadObj()->InheritsFrom("TTree")) {
+           treeName     = key->GetName();
+           std::cout << "Found TTree with name " << treeName << std::endl;
+           numTreeInFile++;
+        }
```
ChristianTackeGSI, line 72: `TKey::ReadObj` returns an owning pointer (see #1461).

Maybe move this in front of the loop?
```cpp
    std::unique_ptr<TTree> tree1;
```

And then:

```suggestion
        tree1.reset(key->ReadObject<TTree>());
        if (tree1) {
           treeName     = key->GetName();
           std::cout << "Found TTree with name " << treeName << std::endl;
           numTreeInFile++;
        }
```

ChristianTackeGSI (commented):
As I recently learned, `ReadObj` returns an owning pointer.

PR #1423, opened 2023-07-07
chore: Use new unique_ptr based SetEventHeader

ChristianTackeGSI:
Also update examples to current standards and compile check
them.

Also update `codemeta.json`.

---

Checklist:

### DIFF
--- a/examples/MQ/pixelDetector/macros/run_dAsciiSource.C
+++ b/examples/MQ/pixelDetector/macros/run_dAsciiSource.C
@@ -71,8 +71,8 @@ void run_dAsciiSource(TString mcEngine = "TGeant3")
     //  PixelDigiReadFromFile* digiRead = new PixelDigiReadFromFile();
     //  run->AddTask(digiRead);
 
-    PixelFindHits* hitFinderTask = new PixelFindHits();
-    run.AddTask(hitFinderTask);
+    auto hitFinderTask = std::make_unique<PixelFindHits>();
+    run.AddTask(std::move(hitFinderTask));
 
     run.Init();
 
--- a/examples/MQ/pixelDetector/macros/run_dAsciiSource.C
+++ b/examples/MQ/pixelDetector/macros/run_dAsciiSource.C
@@ -1,15 +1,36 @@
 /********************************************************************************
- *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
  *                  copied verbatim in the file "LICENSE"                       *
  ********************************************************************************/
+
+#if !defined(__CLING__) || defined(__ROOTCLING__)
+// Example includes
+#include "PixelDigiSource.h"
+#include "PixelEventHeader.h"
+#include "PixelFindHits.h"
+
+// FairRoot includes
+#include "FairParAsciiFileIo.h"
+#include "FairParRootFileIo.h"
+#include "FairRootFileSink.h"
+#include "FairRunAna.h"
+#include "FairRuntimeDb.h"
+#include "FairSystemInfo.h"
+#endif
+
+#include <TStopwatch.h>
+#include <TString.h>
+#include <iostream>
+#include <memory>
+
+using std::cout;
+using std::endl;
+
 void run_dAsciiSource(TString mcEngine = "TGeant3")
 {
-    // Verbosity level (0=quiet, 1=event level, 2=track level, 3=debug)
-    Int_t iVerbose = 0;   // just forget about it, for the moment
-
     // Parameter file
     TString parFile = "pixel_";
     parFile = parFile + mcEngine + ".params.root";
@@ -26,19 +47,17 @@ void run_dAsciiSource(TString mcEngine = "TGeant3")
     // -----   Timer   --------------------------------------------------------
     TStopwatch timer;
 
-    PixelEventHeader* pixelEventHeader = new PixelEventHeader();
-
     // -----   Reconstruction run   -------------------------------------------
-    FairRunAna* fRun = new FairRunAna();
-    fRun->SetEventHeader(pixelEventHeader);
-    fRun->SetSink(new FairRootFileSink(outFile));
+    FairRunAna run{};
+    run.SetEventHeader(std::make_unique<PixelEventHeader>());
+    run.SetSink(std::make_unique<FairRootFileSink>(outFile));
 
     PixelDigiSource* digiSource = new PixelDigiSource("Pixel Digi Source");
     digiSource->SetInputFileName("digis.p0.dat");
 
-    fRun->SetSource(digiSource);
+    run.SetSource(digiSource);
 
-    FairRuntimeDb* rtdb = fRun->GetRuntimeDb();
+    FairRuntimeDb* rtdb = run.GetRuntimeDb();
     FairParRootFileIo* parInput1 = new FairParRootFileIo();
     parInput1->open(parFile.Data());
 
@@ -50,15 +69,15 @@ void run_dAsciiSource(TString mcEngine = "TGeant3")
 
     // -----   TorinoDetector hit  producers   ---------------------------------
     //  PixelDigiReadFromFile* digiRead = new PixelDigiReadFromFile();
-    //  fRun->AddTask(digiRead);
+    //  run->AddTask(digiRead);
 
     PixelFindHits* hitFinderTask = new PixelFindHits();
-    fRun->AddTask(hitFinderTask);
+    run.AddTask(hitFinderTask);
 
-    fRun->Init();
+    run.Init();
 
     timer.Start();
-    fRun->Run();
+    run.Run();
 
     // -----   Finish   -------------------------------------------------------
 
--- a/codemeta.json
+++ b/codemeta.json
@@ -239,6 +239,12 @@
     },
     {
       "@type": "Person",
+      "@id": "https://orcid.org/0000-0002-7006-7986",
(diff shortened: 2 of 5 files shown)


ChristianTackeGSI:
@YanzhaoW does that look good to you?

examples/MQ/pixelDetector/macros/run_dAsciiSource.C
```diff
@@ -26,19 +47,17 @@ void run_dAsciiSource(TString mcEngine = "TGeant3")
...
 
-    PixelEventHeader* pixelEventHeader = new PixelEventHeader();
-
     // -----   Reconstruction run   -------------------------------------------
-    FairRunAna* fRun = new FairRunAna();
-    fRun->SetEventHeader(pixelEventHeader);
-    fRun->SetSink(new FairRootFileSink(outFile));
+    FairRunAna run{};
+    run.SetEventHeader(std::make_unique<PixelEventHeader>());
+    run.SetSink(new FairRootFileSink(outFile));
```
YanzhaoW, line 53: I think we should indicate the ownership transferring in our example code:
```suggestion
auto outFile = std::make_unique<FairRootFileSink>();
run.SetSink(outFile.release());
```

examples/MQ/pixelDetector/macros/run_dAsciiSource.C
```diff
@@ -50,15 +69,15 @@ void run_dAsciiSource(TString mcEngine = "TGeant3")
 
     // -----   TorinoDetector hit  producers   ---------------------------------
     //  PixelDigiReadFromFile* digiRead = new PixelDigiReadFromFile();
-    //  fRun->AddTask(digiRead);
+    //  run->AddTask(digiRead);
 
     PixelFindHits* hitFinderTask = new PixelFindHits();
-    fRun->AddTask(hitFinderTask);
+    run.AddTask(hitFinderTask);
```
YanzhaoW, line 75: Same here
```suggestion
auto task = std::make_unique<PixelFindHits>();
run.AddTask(task.release());
```

YanzhaoW (commented):
Hi, thanks for the PR. There is no problem concerning the event header. It's ok to be merged.

However I would like to urge that we still need to specify the ownership transferring for other objects, like FileSource, FileSink, FairTask. Maybe we could have some PRs to address this issue?

examples/MQ/pixelDetector/macros/run_dAsciiSource.C
ChristianTackeGSI: We already have a `SetSink(unique_ptr)`. So you're right, I should use that!

We don't have that for `SetSource` yet, because SetSource is a `virtual` thing which we need to understand first.

ChristianTackeGSI: I have updated the commit to use unique_ptr for FairSink. But not yet for FairSource.

Let's please postpone the SetSource thing until we understand that mess a bit better.

examples/MQ/pixelDetector/macros/run_dAsciiSource.C
ChristianTackeGSI: I have added a new commit which adds a unique_ptr based `FairRun::AddTask()`. It's only a new API "for the public", not yet fixing any internals (because that looks like a bigger thing).

And it only uses the new API in the examples that I touch in this PR anyway…

examples/MQ/pixelDetector/macros/run_dAsciiSource.C
YanzhaoW: But fSource in the FairRun is already a unqiue_ptr:

https://github.com/FairRootGroup/FairRoot/blob/fa7108b45684c729c2e88b12d7fbf4bbb5a4dc65/base/steer/FairRun.h#L243-L245

Why can we not have a setter taking a unique_ptr if the FairSource is an abstract class?

examples/MQ/pixelDetector/macros/run_dAsciiSource.C
dennisklein: Can you elaborate on this? I cannot follow tbh.
* Are you proposing `AddTask(void*)`?
* How is `AddTask(PixelFindHits())` keeping the temporary object alive?

YanzhaoW: The signature of AddTask could be:
```c++
void AddTask(Task task);
```
and the Task class could be defined as:
```c++
class Task{

public:
   template<typename T>
   Task(T&& obj) {}

  class Concept{ ... };

   template <typename T>
   class Model : Concept{ ...};

private:
   std::unique_ptr<Model> obj;

};
```
Yeah, it's complicated for developers to implement it. But it's super easy for users to use it. For more information you can check this [website](https://www.modernescpp.com/index.php/type-erasure) or this [video](https://www.youtube.com/watch?v=qn6OqefuH08)

dennisklein:
```
run.AddTask(std::make_unique<MyTask>());
```

If this is common one-liner and you dont like the longish "std::make_unique", I guess, we could additionally offer an API like this:

```cpp
run.AddTask<MyTask>(); // call default ctor of MyTask
run.AddTask<MyTask>(myparam1, ...); // call some other ctor of MyTask
```

Implementation would look something like this:

```cpp
template <typename Task, typename ... Args>
FairRun::AddTask(Args&&... args)
{
    AddTask(new Task(std::forward<Args>(args)...));
    // or
    AddTask(std::make_unique<Task>(std::forward<Args>(args)...));
}
```

YanzhaoW: Yes, this is the classic way of storing different types. But there is an issue with this: virtual functions. Using type erasure technique, we can totally eliminate the virtual functions while achieving the same functionality and extensibility but much more simplicity.

dennisklein: I don't see the removal of FairRoot's type hierarchies any time soon, as there is tons of user code depending on those. Not sure, if there is a practical way to transition bit by bit into a non-virtual world - as you stated yourself, it would increase the cognitive load for an experiment framework developer significantly and perhaps also for an experiment user if there would be a mix of both worlds present (without clear distinction) during transition.

What would be very interesting, is to have a prototype, how a FairRoot API (at least for the top-level types that model the "Run", "Task", "Configuration", and related concepts) could look like if it was done without relying on polymorphism via inheritance (if this is the desire ;) ). But perhaps we should do this in a separate experimental namespace - there we would also be more free to fix many more other aspects as well?

... and besides, it is also a serious discussion to have whether your statement "virtual function are a big(ger) issue" is even true ;)

dennisklein: The top two issues in FairRoot/ROOT - from my point of view - are
* unclear/messy object ownership, and
* tons of tight coupling and global state (which renders testability very hard, unit testing is non-existent).

Virtual functions are really not a big issue compared to the above, but I am happy to learn and open to arguments and evidence to the contrary :)

ChristianTackeGSI: I have tried to understand some of this (watched first half or of the video–quite good one!).

For example I don't know, what external polymorphism really buys us? And we're currently somewhat forced by ROOT to use classic (runtime) polymorphism anyway.

As Dennis already pointed out: Our current problem is the unclear ownership. Which is a much bigger problem then using sophisticated tools ("Yeah, it's complicated for developers to implement it.") that some of our developers are not accustomed at using. And for fixing the ownership stuff, it currently feels better to be quite explicit about it.

And to be honest: *I* like the verbose `make_unique`. It's more transparent, more explicit. (So I actually *do* understand, why you don't like "my" "facade like" helpers. And this time I don't like the hiding of things.)

YanzhaoW: Sorry, I didn't mean we need adopt this idiom as soon as possible, but rather just want to give an information that type erasure could be used in this situation.

@ChristianTackeGSI 
Actually the external polymorphism is just for making unrelated objects behave like objects with polymorphism. And type erasure idiom wraps a class around it and add more operations to make the object of that class like a normal value (copiable and movable). The biggest advantage of the type erasure (or some calls it value semantics) could exactly solve our first problem: unclear ownership. Users only need to think about the ownership when a pointer (or unique_ptr) is involved. But if they just use values, like a "type erasure" value, then the whole concept of ownership is non-existent and they don't even need to think about that. Of course, whenever it's easier for users, it's always harder for developers.

Yeah, like I said, I don't mean we should start working on this now without solving other major issues first. ;D

ChristianTackeGSI: Cool!

So let's first work on getting the ownership checked around and fix this stuff. Then later on we can take a look at this again and improve on the then existing structure?
(And maybe a unique_ptr<FairSource> is type erased enough? :wink:)

So that said:

Can we resolve this thread? Anything else I should fix for this PR?

YanzhaoW: @ChristianTackeGSI  Yeah, I think so.

examples/MQ/pixelDetector/macros/run_dAsciiSource.C
YanzhaoW: Oh, sorry, I understood incorrectly. I didn't see SetSource is a virtual function. 

This is bad. Setters and getters should not be a virtual function.

ChristianTackeGSI: The original SetSink is virtual too, but luckily noone (in FairRoot) overrides it. We probably should remove the virtual. But as we want to deprecate it anyway…

karabowi: I think the `FairRun::SetSource` has been made `virtual` only to implement `FairRunAnaProof::SetSource`.
I vote for removing `virtual`.

karabowi (approved):

PR #1427, opened 2023-07-11
Remove `virtual` from `FairRun::SetSource()`

karabowi:
It was made `virtual` only to override it in `FairRunAnaProof`. Removed this function from `FairRunAnaProof` and moved the contents of the function to `Init` on the main thread.

It addresses a comment from PR #1423 .

---

Checklist:

### DIFF
--- a/fairroot/base/steer/FairRunAnaProof.cxx
+++ b/fairroot/base/steer/FairRunAnaProof.cxx
@@ -85,6 +85,11 @@ void FairRunAnaProof::Init()
         // input chain. Do a check if the added files are of the same type
         // as the the input file. Same type means check if they contain the
         // same branch.
+        // FairRunAnaProof should accept only FairFileSource
+        fProofFileSource = dynamic_cast<FairFileSource*>(GetSource());
+        if (!fProofFileSource) {
+            LOG(error) << "FairRunAnaProof. Only FairFileSource is accepted as source";
+        }
     }
     // Load Geometry from user file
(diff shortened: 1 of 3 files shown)


fairroot/base/steer/FairRunAnaProof.cxx
```diff
@@ -85,6 +85,12 @@ void FairRunAnaProof::Init()
         // input chain. Do a check if the added files are of the same type
         // as the the input file. Same type means check if they contain the
         // same branch.
+        // FairRunAnaProof should accept only FairFileSource
+        FairSource* tempSource = GetSource();
+        if (strncmp(tempSource->GetName(), "FairFileSource", 14) != 0) {
+            LOG(error) << "FairRunAnaProof. Only FairFileSource is accepted as source";
+        }
+        fProofFileSource = static_cast<FairFileSource*>(tempSource);
```
ChristianTackeGSI, line 93: I know you just moved the code, that's fine!

Just an idea:

```suggestion
        fProofFileSource = dynamic_cast<FairFileSource*>(GetSource());
        if (!fProofFileSource) {
            LOG(error) << "FairRunAnaProof. Only FairFileSource is accepted as source";
        }
```

This is a small change, because it would allow also classes derived from `FairFileSource`. I hope that might be okay?

karabowi: Agree.

PR #1429, opened 2023-07-16
fix(examples): Memory Leak of FairRunSim

ChristianTackeGSI:
Use value semantics to not leak the FairRunSim in some examples.

And use unique_ptr in some others.

---

Checklist:

### DIFF
--- a/examples/MQ/pixelDetector/src/devices/FairMQSimDevice.h
+++ b/examples/MQ/pixelDetector/src/devices/FairMQSimDevice.h
@@ -17,6 +17,7 @@
 #define FAIRMQSIMDEVICE_H_
 
 #include "FairMQRunDevice.h"
+#include "FairRunSim.h"
 
 #include <Rtypes.h>
 #include <TString.h>
@@ -24,7 +25,6 @@
 #include <memory>
 #include <string>
 
-class FairRunSim;
 class FairField;
 class FairParIo;
 class FairPrimaryGenerator;
@@ -59,6 +59,7 @@ class FairMQSimDevice : public FairMQRunDevice
 
   protected:
     void InitTask() override;
+    void ResetTask() override;
     void PreRun() override;
     void PostRun() override {}
     bool ConditionalRun() override;
@@ -69,7 +70,7 @@ class FairMQSimDevice : public FairMQRunDevice
 
     bool fRunInitialized;   // false, set to true after initialization in the run stage (!)
 
-    FairRunSim* fRunSim;
+    std::unique_ptr<FairRunSim> fRunSim;
     // ------ FairRunSim settings ------
     int64_t fNofEvents;
     std::string fTransportName;
--- a/examples/MQ/pixelDetector/src/devices/FairMQSimDevice.cxx
+++ b/examples/MQ/pixelDetector/src/devices/FairMQSimDevice.cxx
@@ -17,7 +17,6 @@
 #include "FairModule.h"
 #include "FairParSet.h"
 #include "FairPrimaryGenerator.h"
-#include "FairRunSim.h"
 #include "FairRuntimeDb.h"
 
 #include <TCollection.h>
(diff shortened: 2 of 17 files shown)


examples/MQ/pixelDetector/src/devices/FairMQSimDevice.h
```diff
@@ -69,7 +69,7 @@ class FairMQSimDevice : public FairMQRunDevice
 
     bool fRunInitialized;   // false, set to true after initialization in the run stage (!)
 
-    FairRunSim* fRunSim;
```
fuhlig1, line 72: Why not initialise the member already here?

std::unique_ptr<FairRunSim> fRunSim{new FairRunSim()};

@dennisklein,

or must it be created in the InitTask method?

ChristianTackeGSI: We could move even to a `FairRunSim fRunSim;` and use value semantics. I considered that.

But…

I am planning to create a `FairExampleRunSim` that derives from `FairRunSim` to show that deriving is fully okay and that it might be useful to implement standard beheviour for an experiment. One of my ideas for this: The constructor takes an `mcEngine` parameter to set the transport engine. the transport engine isn't known in the ctor of `FairMQSimDevice`. So constructing the `FairExampleRunSim` wouldn't work.

(p.s. I actually have that all already mostly prepared. It's just waiting in a local branch for me to take another look at it.)

dennisklein: > or must it be created in the InitTask method?

I wouldn't say "must", but yes, one should consider the lifetimes. In the [FairMQ state machine](https://github.com/FairRootGroup/FairMQ/blob/master/docs/Device.md#13-state-machine), the application should be configured and initialized in `INITIALIZING TASK` and cleaned in `RESETTING TASK`. As `FairRunSim` seems to be a default constructed placeholder object which gets its configuration supplied after construction anyways, I guess, it does not really matter *when* you create it. The only reason to not construct it with the Device would be, if it had side effects that would better only happen during `INITIALIZING TASK`.

dennisklein: > I am planning to create a `FairExampleRunSim` that derives from `FairRunSim` to show that deriving is fully okay and that it might be useful to implement standard beheviour for an experiment. One of my ideas for this: The constructor takes an `mcEngine` parameter to set the transport engine. the transport engine isn't known in the ctor of `FairMQSimDevice`. So constructing the `FairExampleRunSim` wouldn't work.

:+1:

ChristianTackeGSI: > \[…\] In the [FairMQ state machine](https://github.com/FairRootGroup/FairMQ/blob/master/docs/Device.md#13-state-machine), the application should best be configured and initialized in `INITIALIZING TASK` and cleaned in `RESETTING TASK`. \[…\]

Following this, it sounds like we should `fRunSim.reset()` in `ResetTask` to properly clean up and get back to the state before InitTask.

Technically it wouldn't be needed, as `unique_ptr` takes care of most details for us. But it might be a "good citizen" thing. For example if there are two devices in the same process this would be needed (as there can only be one FairRun at any given time).

Should I add this?

dennisklein: > Following this, it sounds like we should fRunSim.reset() in ResetTask to properly clean up and get back to the state before InitTask.

Sure, if it makes sense for the `FairRunSim`.

Alternatively, you could say a reset cycle is not supported and add some kind of an error, if devices go through a InitTask twice. (At the moment, it is not possible for a user to change the FMQ state machine.)

ChristianTackeGSI: IMHO FairRunSim should support all of this. So let's have a proper ResetTask that cleans up/etc.

I changed the commit accordingly.

karabowi (approved):

Issue #1418, opened 2023-06-26
CTest errors fail to be caught by the CI workflow


YanzhaoW:
**Describe the bug**
Multiple crashes in ctests are not caught by the CI workflow.

### Unused file source in run_reco_timebased.C
https://github.com/FairRootGroup/FairRoot/blob/fa7108b45684c729c2e88b12d7fbf4bbb5a4dc65/examples/advanced/Tutorial3/macro/run_reco_timebased.C#L42-L48
Here even though the file source `fFileSource` is initialised, it isn't used by `FairRun`. Obviously this should have crashed the program since FairRunAna must have a source to analyze (it does crash the the PR #1254).

### Double deleting of file source in FairRootManager and FairRun
In `FairRootManager`, the file source is deleted here:
https://github.com/FairRootGroup/FairRoot/blob/fa7108b45684c729c2e88b12d7fbf4bbb5a4dc65/base/steer/FairRootManager.cxx#L136-L141

In `FairRun`, the file source is deleted automatically using `std::unique_ptr`:
https://github.com/FairRootGroup/FairRoot/blob/fa7108b45684c729c2e88b12d7fbf4bbb5a4dc65/base/steer/FairRun.h#L242-L246

However, deleting from `FairRootManager` is sometimes prevented by:
https://github.com/FairRootGroup/FairRoot/blob/fa7108b45684c729c2e88b12d7fbf4bbb5a4dc65/base/steer/FairRun.cxx#L86-L91

But everything falls apart if `FairRootManager` is deleted before `FairRun`, as has happened in Ctests such as `run_digi_timebased.C`:
https://github.com/FairRootGroup/FairRoot/blob/fa7108b45684c729c2e88b12d7fbf4bbb5a4dc65/examples/advanced/Tutorial3/macro/run_digi_timebased.C#L40-L45
Here `FairRunAna` is initialised with the "evil" `new` operator and not deleted at the end of the program. As the result, the destructor of `FairRootManager` is executed followed by the destructor of `FairRun`. Therefore, we have the double deleting of one file source pointer, which is an UB (some time the program crashes, sometimes not).

**Additional context**
I have never understood what is the use of this "idiom". It's also used at many places in R3BRoot. 
```cpp
if(a_ptr != nullptr)
    delete a_ptr;
```
[...]

PR #1430, opened 2023-07-19
fix: Handle Sink/Source ownership more properly

ChristianTackeGSI:
Waiting for:
- [x] #1442

FairRootManager and FairRun both used to `delete` fSink and fSource. In specific situations (FairRootManager getting destructed before FairRun) this leads to a double delete!

Deprecate the old FairRootManager::Set{Source,Sink} APIs that take something like an owning pointer.  Introduce new Set{Sink,Source}NonOwning internal APIs.

See: https://github.com/FairRootGroup/FairRoot/issues/1418 (see "Double deleting of file source in FairRootManager and FairRun")

---

Checklist:

## Summary by CodeRabbit

- **New Features**
  - Deprecated `SetSource` and `SetSink` methods now use modern smart pointers for better memory management.
  - Introduced deprecation warnings for `SetOutputFile` and `SetOutputFileName` methods.

- **Improvements**
  - Enhanced memory management by replacing raw pointers with `std::unique_ptr`.
  - Updated to the latest copyright year.

- **Code Maintenance**
  - Removed outdated and commented-out code.
  - Improved code clarity by adjusting `#include` directives and adding necessary using statements.

### DIFF
--- a/fairroot/base/steer/FairRootManager.cxx
+++ b/fairroot/base/steer/FairRootManager.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -84,10 +84,8 @@ FairRootManager::FairRootManager()
     , fBrPerMap()
     , fFillLastData(kFALSE)
     , fEntryNr(0)
-    , fSource(0)
     , fSignalChainList()
     , fEventHeader(new FairEventHeader())
-    , fSink(nullptr)
     , fUseFairLinks(kFALSE)
     , fFinishRun(kFALSE)
 {
@@ -105,10 +103,6 @@ FairRootManager::~FairRootManager()
 
     delete fEventHeader;
     delete fSourceChain;
-    if (fSink)
-        delete fSink;
-    if (fSource)
-        delete fSource;
 
     // Global cleanup
 
@@ -388,6 +382,7 @@ void FairRootManager::WriteFolder()
         fSink->WriteObject(fTimeBasedBranchNameList, "TimeBasedBranchList", TObject::kSingleKey);
     }
 }
+
 Bool_t FairRootManager::SpecifyRunId()
 {
     if (!fSource) {
@@ -892,7 +887,7 @@ std::string FairRootManager::GetNameFromFile(const char* namekind)
 {
     std::string default_name = "";
 
-    if (strcmp(namekind, "treename=") && strcmp(namekind, "foldername=")) {
+    if ((strcmp(namekind, "treename=") != 0) && (strcmp(namekind, "foldername=") != 0)) {
         LOG(fatal) << "FairRootManager, requested " << namekind << ", while only treename= and foldername= available.";
         return default_name;
     }
--- a/fairroot/base/steer/FairRootManager.h
+++ b/fairroot/base/steer/FairRootManager.h
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -8,7 +8,7 @@
 #ifndef FAIR_ROOT_MANAGER_H
 #define FAIR_ROOT_MANAGER_H
 
-#include "FairLogger.h"
+#include "FairMemory.h"
 #include "FairSink.h"
 #include "FairSource.h"
 
@@ -17,7 +17,8 @@
 #include <TObject.h>     // for TObject
 #include <TRefArray.h>   // for TRefArray
 #include <TString.h>     // for TString, operator<
-#include <map>           // for map, multimap, etc
+#include <fairlogger/Logger.h>
+#include <map>   // for map, multimap, etc
 #include <memory>
 #include <string>
 #include <type_traits>   // is_pointer, remove_pointer, is_const, remove...
@@ -49,6 +50,8 @@ class TTree;
 
 class FairRootManager : public TObject
 {
+    friend class FairRun;
+
   public:
     /**dtor*/
     ~FairRootManager() override;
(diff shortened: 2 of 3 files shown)


ChristianTackeGSI:
@YanzhaoW: This should fix the second part of #1418. Could you take a look?

fairroot/base/steer/FairRootManager.cxx
```diff
@@ -120,6 +114,30 @@ FairRootManager::~FairRootManager()
...
+    fSink = sink;
+    fSinkOwned.reset(sink);
+}
+
+void FairRootManager::SetSinkNonOwning(FairSink* sink)
+{
+    fSink = sink;
+    fSinkOwned.reset();
+}
+
```
YanzhaoW, line 140: If the input parameter is a raw pointer, according to the common convention, it already means a non-owning. So I would say the `NonOwning` in the function name is unnecessary. 

 Second, since fSink is owned by the FairRootManager, its setter should be owning:
```cpp
std::unique_ptr<FairSink> fSink;
void FairRootManager::SetSink(std::unique_ptr<FairSink> sink)
{
     fSink = std::move(sink);
}
[[deprecated]]void FairRootManager::SetSink(FairSink* sink)
{
     fSink.reset(sink);
}
```

I also don't quite understand why we have two member variables for the sink in this case.

YanzhaoW (commented):
Hi
Thanks for the PR. 

We also have `FairRun::SetSink(std::unique_ptr<FairSink> sink)`. Just from the class signatures, it's very confused which class owns the sink file. Maybe we should also deprecate the `FairRun::SetSink` and tell users to set the sink directly from the `FairRootManager`. Or we could just let `FairRun` own the sink file?

fairroot/base/steer/FairRootManager.cxx
ChristianTackeGSI: First things first:

The idea is to let FairRun own the sink and source. FairRun::Set{Sink,Source} should be the APIs for users. FairRootManager only needs a reference to the sink and source. So `FairRootManager::Set{Sink,Source}*` *should* be internal and only take non-owning pointers.

And yes, I should add that to the commit body and PR message above. Will do so next week after #1432 gets merged.

That said, we have a problem:
The current FairRootManager::Set{Sink,Source}` member functions are public and not marked as internal. So people are allowed to use them (and I think, they're used). Those APIs take a half-owning pointer currently (which you correctly diagnosed in your issue!). This is quite broken, really, and should be deprecated.
So:
* We need to keep the old (broken) API for a bit of time for the current users.
* Introduce a new non-owning, internal API for FairRun.

We need a name for the new API. I made up this Set{Sink,Source}NonOwning() name. I am fully open for better names. Maybe something with "internal" in it? I don't know.

YanzhaoW: Sorry for the late reply. Maybe we should just deprecate the setters of FairRootManager, without adding any new setters for the sink and source?

> So FairRootManager::Set{Sink,Source}* should be internal and only take non-owning pointers.

There shouldn't be a setter even for the internal usage. Sink and source should be set directly from users, not from anything else. If there is a internal way to set the sink or source, users basically would have no idea about it and this could cause a lot of problems in the future.

ChristianTackeGSI: FairRootManager is a mess. We're trying to reduce the messiness somewhat. For this PR I am trying to fix the double-delete and not introduce too much messiness while still keeping (and deprecating) public APIs.

For this discussion, let's please assume that FairRootManager is itself an internal helper class (or at least it should be) that needs to have a reference to a sink and/or source to be able to fulfil its job. To set this reference … well … I think, we should use a setter?

As FairRootManager isn't yet an internal only helper class, we need to make that setter "internal". Either by adding docs about it being internal, or by making it `private` and using `friend` to allow FairRun to call it. I somewhat prefer comments.

YanzhaoW: > let's please assume that FairRootManager is itself an internal helper class (or at least it should be) that needs to have a reference to a sink and/or source to be able to fulfil its job.

There could be a lot of problems if we have a source/sink member variables (either by pointer or reference) and the "internal setters" in FairRootManager. The biggest one is about the lifetime of source/sink. Since they are not owned by FairRootManager, they could be deleted by FairRun and then accessed by the FairRootManager (, which causes a seg fault). A good design should disallow this to happen.

So I would suggest to completely remove source/sink related (public & private) member variables and functions. If FairRootManager wants to use it, simply get it from FairRun, which owns the source/sink resource:
```cpp
void FairRootManager::MyFunction()
{
     auto* source = FairRun::Instance()->GetSource();
     UseThisSource(source);
}
```
With this, the lifetime of source/sink resources can always be guaranteed.

ChristianTackeGSI: `FairRun::Instance()` references global state. We want to reduce global state in the long run, really.

That said, let's look at lifetime.

First: Thanks for poking on this one! I fixed an issue now.

So the main risk is that FairRootManager (FRM) has a dangling reference, because FairRun deleted the sink/source. All the member functions that get a new sink/source properly reset the FRM thing before `reset()`in the unique_ptr (and thus deleting the old sink/source. So those should be safe. I missed the destructor! So now the destructor also resets the FRM sink/source, before the member variable destructors of `FairRun::{fSink,fSource}` delete the sink/source. So this should be safe as well now.

YanzhaoW: >All the member functions that get a new sink/source properly reset the FRM thing before reset()in the unique_ptr (and thus deleting the old sink/source. So those should be safe. I missed the destructor! So now the destructor also resets the FRM sink/source, before the member variable destructors of FairRun::{fSink,fSource} delete the sink/source. So this should be safe as well now.

Hi, 

I would say using this trick just makes sure the algorithm can run successfully and it doesn't guarantee the safety of the code. For example, what if FairRootManager needs to write data to the fSink and FairRun is deleted before this action takes place? We could have the program running successfully but it fails to produce expected result. And a successfully running program is even harder to debug than a program that crashes. For example, using `auto* run = new FairRun()` (forgetting to delete it in the end) and `auto run = std::make_unique<FairRun>()` can get totally different results.

> We want to reduce global state in the long run, really.

Generally, I am not fan of neither `reset` or singleton. We don't have control of when the reset functions are called. And bad things could happen when they are called either too early or too late or not being called at all. In case of singleton, we can't have control over when constructors or destructors are called.

Like I have mentioned above, there is a better way to manage this: **chain of responsibility**. If sink/source files are owned by FairRun, everything related to them should be handled via their owner, FairRun. If FairRootManager needs to write the data to sink file, all it needs to do is to own FairRun and pass the FairRun pointer to the callee. In this case, FairRootManager doesn't deal with sink and source file directly. And therefore, it doesn't need to care about the lifetime of sink or source file anymore. It only needs to care about whether the owning FairRun is still alive or not.

ChristianTackeGSI: TBH, I think, most of your points are quite academical now.

> For example, what if FairRootManager needs to write data to the fSink and FairRun is deleted before this action takes place?

To let this happen, one has to create a quite complex setup:

* FairRootManager is a `thread_local`. So each thread has one.
* main thread (prepare):
  * Setup FairRootManager& `FRM`
  * Setup FairRun* `run`.
  * `run->SetSink(std::move(somesink));`
  * calls `FRM.SetSinkNonOwning`.
  * We're ready.
* main thread: Fire up a new thread "deleter", hand over `run`
* "deleter" thread needs time to start up / sleeps a bit.
* main thread: start to call `FRM.SomeSinkWriter`. It loads `fSink` into a register, and sleeps
* "deleter" thread now finally `delete run;`. sink goes away. (and it calls `FRM.SetSinkNonOwning(nullptr)` (on the "right" FRM, because FairRun has a reference to the FRM in the other thread), but as the reference to the sink is already in a register, that's not important.
* main thread awakes and uses the sink in the register to act on the destructed sink.
* Booom.

Yes, this is possible.

I *think* that FairRoot has way more problematic thread issues than this one. In other words: If a user uses threads manually (like in this case), then they should be very careful on their own.

Did I miss anything?

ChristianTackeGSI: I think, most of this discussion is now mostly out dated with the current maybe_owning_ptr implementation. Or it's highly academic concerning the current real world use cases.

Yes, the implementation is problematic. But it's less problematic then the current state. And it `[[deprecated]]`s some current problematic cases anyway.

fairroot/base/steer/FairRootManager.cxx
```diff
@@ -892,7 +887,7 @@ std::string FairRootManager::GetNameFromFile(const char* namekind)
 {
     std::string default_name = "";
 
-    if (strcmp(namekind, "treename=") && strcmp(namekind, "foldername=")) {
+    if ((strcmp(namekind, "treename=") != 0) && (strcmp(namekind, "foldername=") != 0)) {
```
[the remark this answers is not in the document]
ChristianTackeGSI: Out of context for this PR.

dennisklein (approved):

PR #1416, opened 2023-06-23
fix: Do not leak TFile in Fair{File,Mixed}Source

ChristianTackeGSI:
- [x] #1442

And refactor into FairFileSourceBase.

---

Checklist:

* [X] Followed the [Contributing Guidelines](https://github.com/FairRootGroup/FairRoot/blob/dev/CONTRIBUTING.md)

## Summary by CodeRabbit

- **New Features**
  - Improved memory management for file handling using smart pointers, enhancing application stability and reducing memory leaks.
  - Added error handling for file operations, providing more robust and reliable file management.

- **Bug Fixes**
  - Addressed issues related to file pointer initialization and management, ensuring consistent and error-free file handling.

- **Refactor**
  - Updated internal file management to use `std::unique_ptr` for better ownership semantics and memory safety.
  - Removed obsolete methods and variables related to file handling, streamlining the codebase.

PR #1442, opened 2023-08-16
feature: Introduce `maybe_owning_ptr`

ChristianTackeGSI:
See the doxygen comments for some details.

Will be needed in #1416, #1430.

---

Checklist:

### DIFF
--- a/fairroot/tools/FairMemory.h
+++ b/fairroot/tools/FairMemory.h
@@ -1,0 +1,67 @@
+/********************************************************************************
+ *    Copyright (C) 2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
+ *                                                                              *
+ *              This software is distributed under the terms of the             *
+ *              GNU Lesser General Public Licence (LGPL) version 3,             *
+ *                  copied verbatim in the file "LICENSE"                       *
+ ********************************************************************************/
+
+#ifndef FAIRROOT_MEMORY_H_
+#define FAIRROOT_MEMORY_H_
+
+#include <memory>
+#include <optional>
+
+/**
+ * \brief Internal implementation details
+ *
+ * Any declarations in here are strictly for internal usage.
+ * They may never enter public APIs.
+ *
+ * \note Anything in here can change at any time without notice!
+ */
+namespace fairroot::detail
+{
+
+template<typename T, typename Deleter = std::default_delete<T>>
+struct optional_deleter
+{
+    std::optional<Deleter> deleter{std::in_place};
+
+    optional_deleter() = default;
+    optional_deleter(const Deleter& x)
+        : deleter(std::in_place, x)
+    {}
+    optional_deleter(std::nullopt_t)
+        : deleter()
+    {}
+
+    void operator()(T* ptr) const
+    {
+        if (deleter) {
+            (*deleter)(ptr);
+        }
+    }
+};
+
+/**
+ * \brief unique_ptr that alternatively can hold a reference
+ *
+ * In rare cases we need a unique_ptr that can alternatively
+ * contain a reference (and thus not delete its content
+ * on destruction).
+ *
+ * \note Think twice before using this!
+ *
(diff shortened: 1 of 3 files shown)


dennisklein:
Can you post a pseudo code example on how you want to use this?

ChristianTackeGSI:
Real code example of usage options: https://godbolt.org/z/5EG5qz55j

Or did you mean on how this will end up in FairRoot?

See both #1416 and #1430: Both have naive `T* fFoo; unique_ptr<T> fFooOwned;` way of handling the problamatic parts. I want to replace that with a `maybe_owning_ptr<T> fFoo;`.

fairroot/tools/FairMemory.h
```diff
@@ -0,0 +1,67 @@
...
+        : deleter()
+    {}
+
+    void operator()(T* ptr) const
+    {
+        if (deleter) {
+            (*deleter)(ptr);
+        }
+    }
+};
```
dennisklein, line 45:
```suggestion
template <typename T, typename Deleter>
struct optional_deleter
{
    std::optional<Deleter> deleter{std::in_place};

    optional_deleter() = default;
    optional_deleter(const Deleter& x)
        : deleter(std::in_place, x)
    {}
    optional_deleter(std::nullopt_t)
        : deleter()
    {}

    void operator()(T* ptr) const
    {
        if (deleter) {
            (*deleter)(ptr);
        }
    }
};
```

* Or is there a good reason to hard-code `std::default_delete` here I am missing+?
* Perhaps there is even some way to inspect the `T` from `Deleter`? (to just have `template <typename Deleter> struct optional_deleter`)

fairroot/tools/FairMemory.h
```diff
@@ -0,0 +1,67 @@
...
+ * on destruction).
+ *
+ * \note Think twice before using this!
+ *
+ * It's mostly meant for cases, where FairRoot has a public API
+ * that enforces this behaviour. This should be deprecated and
+ * removed sooner than later.
+ */
+template<typename T>
+using maybe_owning_ptr = std::unique_ptr<T, optional_deleter<T>>;
```
dennisklein, line 61:
```suggestion
template <typename T, typename Deleter = std::default_delete<T>>
using maybe_owning_ptr = std::unique_ptr<T, optional_deleter<T, Deleter>>;
```

see my other comment

dennisklein:
> Real code example of usage options: https://godbolt.org/z/5EG5qz55j
> 
> Or did you mean on how this will end up in FairRoot?

No, thx, the above was what I was looking for :)

fairroot/tools/FairMemory.h
ChristianTackeGSI: > * Or is there a good reason to hard-code std::default_delete here which I am missing?

Only KISS. We hopefully wont ever need optional other deleters in FairRoot.

As I (intentionally) named this thing more like in the `std::` namespace (and not using the FairRoot conventions), I wrote it with the hidden idea that others could copy and re-use this code, if they needed that functionality. So being more generic is good!

> * Perhaps there is even some way to inspect the T from Deleter? 

```diff
-    void operator()(T* ptr) const
+    void operator()(auto* ptr) const
```

this would remove the last trace of `T` from `optional_deleter`. But this it seems, that this is only possible with C++20. So we can't have this yet.

So for I changed it to
```c++
template<typename T, typename Deleter = std::default_delete<T>>
```

so that it can be used without the deleter if wanted.

ChristianTackeGSI:
> Can you post a pseudo code example on how you want to use this?

Should we have examples in the doxygen docs?

dennisklein:
> Should we have examples in the doxygen docs?

Would be a bonus exercise, not required from my point of view. If you want to add more docs, I'd prefer some unit test coverage - as "living" usage documentation.

ChristianTackeGSI:
Yes, let's create a catch2 test for this, after your catch2 PR gets merged. :-)

PR #1446, opened 2023-08-22
feat: Start with FairRunSim::ls()

ChristianTackeGSI:
Allow to list the object hierarchy under a FairRunSim.

Scroll to the end of this [CDash log](https://cdash.gsi.de/testDetails.php?test=15939953&build=406391) to see a typical output.

---

Checklist:

### DIFF
--- a/fairroot/base/steer/FairRunSim.cxx
+++ b/fairroot/base/steer/FairRunSim.cxx
@@ -34,6 +34,7 @@
 #include <TList.h>         // for TList
 #include <TObjString.h>    // for TObjString
 #include <TObject.h>       // for TObject
+#include <TROOT.h>         //
 #include <TRandom.h>       // for gRandom
 #include <TSystem.h>       // for TSystem, gSystem
 #include <iostream>        // for cout, endl, ostream
@@ -72,6 +73,9 @@ FairRunSim::FairRunSim(Bool_t isMaster)
     fginstance = this;
     fRunId = 0;
     fAna = kFALSE;
+    fIons->SetName("fIons");
+    fParticles->SetName("fParticles");
+    ListOfModules->SetName("ListOfModules");
 }
 #pragma GCC diagnostic pop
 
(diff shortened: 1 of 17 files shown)


fairroot/base/steer/FairRunSim.cxx
```diff
@@ -381,4 +382,32 @@ FairMCEventHeader* FairRunSim::GetMCEventHeader()
...
+        fGen->ls(option);
+    }
+#pragma GCC diagnostic pop
+    if (fField) {
+        fField->ls(option);
+    }
+    // if (fSimulationConfig) {
+    //     fSimulationConfig->ls(option);
+    // }
+    ListOfModules->ls();
```
dennisklein, line 401: I guess for the `TObjArray`s we should print something else?

```
OBJ: TObjArray	TObjArray	An array of objects : 0
```

is not that meaningful.

ChristianTackeGSI: Well, the `ls` member function of TObjArray is implemented by ROOT. So either we write our own free standing handler or improve the one in ROOT.

That said, we can start by giving our TObjArrays some name/title. So that at least the name/title is shown instead of `TObjArray … An array of objects`.

dennisklein: How can you give it a name?

ChristianTackeGSI: [`TCollection::SetName`](https://root.cern/doc/v628/classTCollection.html#a6d6c3b2d795b309c0d8d2a7eebbb30d4)

I had hoped for `::SetTitle`. But that does not exist for TObjArray. So we can't set the title.

ChristianTackeGSI: Added a few `SetName()`.

See end of: https://cdash.gsi.de/testDetails.php?test=15932527&build=406173

PR #1447, opened 2023-08-30
Fix exported target name

fuhlig1:
With the latest version of yaml-cpp (0.8.0) the name of the exported target has changed from yaml-cpp to yaml-cpp::yaml-cpp which breaks the compilation.
To work around the problem an alias yaml-cpp::yaml-cpp is created in case the exported target name is yaml-cpp.

---

Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
  #                  copied verbatim in the file "LICENSE"                       #
  ################################################################################
 cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
-cmake_policy(VERSION 3.15...3.23)
+cmake_policy(VERSION 3.15...3.27)
 
 set(PROJECT_MIN_CXX_STANDARD 17)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,8 +5,8 @@
  #              GNU Lesser General Public Licence (LGPL) version 3,             #
  #                  copied verbatim in the file "LICENSE"                       #
  ################################################################################
-cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
-cmake_policy(VERSION 3.15...3.27)
+cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
+cmake_policy(VERSION 3.18...3.27)
 
 set(PROJECT_MIN_CXX_STANDARD 17)
 
@@ -260,12 +260,11 @@ endif()
 # The target name includes the namespace yaml-cpp begining
 # from version 0.8.0. So the new exported target is now
 # yaml-cpp::yaml-cpp.
-if(yaml-cpp_FOUND)
-  if(TARGET yaml-cpp::yaml-cpp)
-    set(YAMLCPPLIB yaml-cpp::yaml-cpp)
-  else()
-    set(YAMLCPPLIB yaml-cpp)
-  endif()
+# To work also with previous versions of yaml-cpp an alias
+# is created and the new target name is used in the project
+if(yaml-cpp_FOUND AND TARGET yaml-cpp AND NOT TARGET yaml-cpp::yaml-cpp)
+  message(VERBOSE "Creating target alias yaml-cpp::yaml-cpp")
+  add_library(yaml-cpp::yaml-cpp ALIAS yaml-cpp)
 endif()
 
 if(BUILD_TESTING)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,6 +256,17 @@ if(    yaml-cpp_FOUND
     INTERFACE_INCLUDE_DIRECTORIES "${YAML_CPP_INCLUDE_DIR}"
   )
 endif()
+# Workaround to fix a problem with the imported target name
+# The target name includes the namespace yaml-cpp begining
+# from version 0.8.0. So the new exported target is now
+# yaml-cpp::yaml-cpp.
+if(yaml-cpp_FOUND)
+  if(TARGET yaml-cpp::yaml-cpp)
+    set(YAMLCPPLIB yaml-cpp::yaml-cpp)
+  else()
+    set(YAMLCPPLIB yaml-cpp)
+  endif()
+endif()
 
 if(BUILD_TESTING)
   Include(FetchContent)
--- a/fairroot/mcconfigurator/CMakeLists.txt
+++ b/fairroot/mcconfigurator/CMakeLists.txt
@@ -25,7 +25,7 @@ target_include_directories(${target} PUBLIC
 target_link_libraries(${target}
   PUBLIC
   FairRoot::Base # FairGenericVMCConfig
-  yaml-cpp
+  ${YAMLCPPLIB}
 
   PRIVATE
   FairRoot::Tools
--- a/fairroot/mcconfigurator/CMakeLists.txt
+++ b/fairroot/mcconfigurator/CMakeLists.txt
@@ -25,7 +25,7 @@ target_include_directories(${target} PUBLIC
 target_link_libraries(${target}
   PUBLIC
   FairRoot::Base # FairGenericVMCConfig
-  ${YAMLCPPLIB}
+  yaml-cpp::yaml-cpp
 
   PRIVATE
   FairRoot::Tools


dennisklein (approved):

CMakeLists.txt
```diff
@@ -256,6 +256,16 @@ if(    yaml-cpp_FOUND
     INTERFACE_INCLUDE_DIRECTORIES "${YAML_CPP_INCLUDE_DIR}"
   )
 endif()
+# Workaround to fix a problem with the imported target name
+# The target name includes the namespace yaml-cpp begining
+# from version 0.8.0. So the new exported target is now
+# yaml-cpp::yaml-cpp.
+# To work also with previous versions of yaml-cpp an alias
+# is created and the new target name is used in the project
+if(yaml-cpp_FOUND AND TARGET yaml-cpp)
```
ChristianTackeGSI, line 265: Would this be too paranoid?

```suggestion
if(yaml-cpp_FOUND AND TARGET yaml-cpp AND NOT TARGET yaml-cpp::yaml-cpp)
```

CMakeLists.txt
```diff
@@ -256,6 +256,16 @@ if(    yaml-cpp_FOUND
...
+  message(STATUS "Creating target alias yaml-cpp::yaml-cpp")
```
ChristianTackeGSI, line 266:
```suggestion
  message(VERBOSE "Creating target alias yaml-cpp::yaml-cpp")
```

I don't think this message is of interest to most normal users in most cases?

fuhlig1: The message can be removed completely. I add it for debugging.

CMakeLists.txt
fuhlig1: I don't know if this is too paranoid but it shouldn't hurt. I will try if I see problems or if it simply works.

CMakeLists.txt
fuhlig1: Put your suggestion.

CMakeLists.txt
```diff
@@ -256,6 +256,16 @@ if(    yaml-cpp_FOUND
...
 endif()
+# Workaround to fix a problem with the imported target name
+# The target name includes the namespace yaml-cpp begining
+# from version 0.8.0. So the new exported target is now
+# yaml-cpp::yaml-cpp.
+# To work also with previous versions of yaml-cpp an alias
+# is created and the new target name is used in the project
+if(yaml-cpp_FOUND AND TARGET yaml-cpp AND NOT TARGET yaml-cpp::yaml-cpp)
+  message(VERBOSE "Creating target alias yaml-cpp::yaml-cpp")
+  add_library(yaml-cpp::yaml-cpp ALIAS yaml-cpp)
```
ChristianTackeGSI, line 267: Okay, something strange is happening:

```console
CMake Error at CMakeLists.txt:267 (add_library):
  add_library cannot create ALIAS target "yaml-cpp::yaml-cpp" because target
  "yaml-cpp" is imported but not globally visible.
```
<https://cdash.gsi.de/buildSummary.php?buildid=405765#!#Configure>

I don't understand that now.

ChristianTackeGSI: Or we do, what we did with `VMCLIB`?

```cmake
if(yaml-cpp_FOUND)
  if(TARGET yaml-cpp::yaml-cpp)
    set(YAMLCPPLIB yaml-cpp::yaml-cpp)
  else()
    set(YAMLCPPLIB yaml-cpp)
  endif()
endif()
```

dennisklein: > Okay, something strange is happening:
> 
> ```
> CMake Error at CMakeLists.txt:267 (add_library):
>   add_library cannot create ALIAS target "yaml-cpp::yaml-cpp" because target
>   "yaml-cpp" is imported but not globally visible.
> ```
> 
> https://cdash.gsi.de/buildSummary.php?buildid=405765#!#Configure
> 
> I don't understand that now.

I believe that is an issue with CMake < 3.18, see https://cmake.org/cmake/help/latest/release/3.18.html#id8 (first bullet point under "Commands")

fuhlig1: In that case I see the two possibilities.

1. Either require CMake > 3.18 in our CMakeLists.txt
2. Use the option proposed by Christian which should work with any CMake version.

For the compilation of FairSoft a CMake Version >= 3.18 is needed since nov22. So we could require the same version also for FairRoot. I have to check if CMake 3.18 is enough to compile FairSoft jun23. I did my tests up to now only with the latest version.

What is your opinion which option to use?

dennisklein: > What is your opinion which option to use?

+1 for option 1 from me

dennisklein: Some of the CI builds are failing now due to too low CMake version, I will look into updating those. That is overdue anyways.

ChristianTackeGSI: Ubuntu 20.04LTS comes with 3.16.3.

Debian-10 has 3.18.4-2 in the backports. So that should be okay.

ChristianTackeGSI: BTW: Are the transitive dependencies on mcconfigurator "unrolled"? That is: Does mcconfigurator use the aliased-to value and not the alias?

Because otherwise ExperimentRoot would need the same mechanism (and then I would vote for option 2).

dennisklein: > I think, we will backport this to 18.x at some point, right? Do we want to raise the requirements there as well? If not, I would vote for option (2).

Well, a backport can choose a different solution (in other words, it may be more work than just a simple cherry-pick), if necessary? Otherwise, you can never move forward in your dev branch?

dennisklein: > Because otherwise ExperimentRoot would need the same mechanism (and then I would vote for option 2).

Hm, what is the problem with having to adapt to new FairRoot releases? Isn't that the natural way of things?

dennisklein: Ok, I think I see it now:
* That a project has to deal with the changed target name of yaml-cpp, is not caused by FairRoot. So, if downstream chooses to use yaml-cpp 0.8+, it needs to adapt, nothing to do in this case.
* However, if downstream chooses to use yaml-cpp < 0.8, it will have to change something as well. For this case, we have a choice:
   1. Make sure to not leak our local target alias to downstream (basically option (2))
   2. Export the target alias, if we create it

fuhlig1: > BTW: Are the transitive dependencies on mcconfigurator "unrolled"? That is: Does mcconfigurator use the aliased-to value and not the alias?
> 
> Because otherwise ExperimentRoot would need the same mechanism (and then I would vote for option 2).

This is the produced CMake code which shows that the original target name is used.

```
# Create imported target FairRoot::MCConfigurator
add_library(FairRoot::MCConfigurator SHARED IMPORTED)

set_target_properties(FairRoot::MCConfigurator PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_17"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "FairRoot::Base;yaml-cpp;ROOT::RIO"
)

```

dennisklein: > This means that we don't leak our internal target.

Unexpected, but even better :)

ChristianTackeGSI: > Well, a backport can choose a different solution (in other words, it may be more work than just a simple cherry-pick), if necessary? Otherwise, you can never move forward in your dev branch?

I would then propose to
* Use the "old way" in dev (wait!)
* backport by cherry-pick
* Update dev to the new way, including updating the handling of VMC to the "new way".

dennisklein: > I would then propose to
> 
> * Use the "old way" in dev (wait!)
> * backport by cherry-pick
> * Update dev to the new way, including updating the handling of VMC to the "new way".

Hmm, ok, but then the only reason to keep the "old way" in the 18 releases is to avoid the bump of required CMake?

dennisklein: So, it will likely go like this:
1. ExperimentRoot will likely adopt yaml-cpp 0.8 first (because it is there now and FairRoot 19 not), they look at latest FairRoot 18 point release and copy the "old way"
2. Then FairRoot 19 is released, downstream adopts it, and keeps using "old way" because it keeps working
I guess this should be fine?

ChristianTackeGSI: > Hmm, ok, but then the only reason to keep the "old way" in the 18 releases is to avoid the bump of required CMake?

At least that would be my idea. So that FairRoot 18.x would still work on old systems with old CMake.


> 1. ExperimentRoot will likely […] copy the "old way"

Well, they only need to copy things if they need yaml-cpp in their own code. If they just want to use mcconfigurator, then it should have the "right" dependencies already and people only need to have a `find_package(yaml-cpp)` in their CMakeLists.txt, right?

ChristianTackeGSI (approved):
I changed the milestone to 18.8, so that we remember to backport it, once it is merged.

CMakeLists.txt
dennisklein: And now add another commit that modernizes it to an alias target for yaml and vmc? @ChristianTackeGSI that was your plan, right?

fuhlig1: I thought first this one should be merged and ported and only after that step I would do a new PR.

dennisklein: But why wait?

fuhlig1: Okay. Now also the new version is available.

dennisklein: Now both commits have the same title line, how about rewording the commit msg a bit, e.g.

```
build: Support yaml-cpp 0.8+

Beginning from version 0.8.0 the exported target has been renamed to `yaml-cpp::yaml-cpp`.
To support also older versions choose the correct target name stored in a CMake variable.
```

```
build: Modernize `YAMLCPPLIB` variable with alias target

Since CMake 3.18 we can use the more expressive and more strictly typed target alias
instead of using a simple variable that just contains an arbitrary string. This commit
modernizes the variable by creating a local target alias if the new target name is not
defined. The alias is local because CMake resolves it during target export. This means
users of an exported FairRoot target will not see our alias name and will **not** need
any name switching logic on their end for the purpose of using any FairRoot library
depending on yaml-cpp.
```

I dropped "workaround" as I think it is a bit misleading here. We are just adapting to newly released upstream changes. "Workaround" suggests (to me at least) the new target name is a bug and in the future it will be reverted and the workaround can go away again?!

fuhlig1: Done.

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
  #                  copied verbatim in the file "LICENSE"                       #
  ################################################################################
 cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
-cmake_policy(VERSION 3.15...3.23)
+cmake_policy(VERSION 3.15...3.27)
```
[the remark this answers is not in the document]
fuhlig1: Done.

dennisklein (approved):

dennisklein:
centos7 and ubuntu20.04 ci failures expected, they need to be updated, I am on it. Merging. Thx all!

PR #1455, opened 2023-09-15
chore: Drop TMemberInspector declaration

ChristianTackeGSI:
No code in FairRoot needs a TMemberInspector directly. If anything from ROOT needs it, then that include should declare it.

---

Checklist:


fuhlig1:
@ChristianTackeGSI,

if I remember the issue correctly the forward came in when we used the tool include-what-you-use to get our includes complete. Probably we did not took care at that point and did the change automatically which brings in the unneeded forward declaration.

fuhlig1 (approved):

PR #1457, opened 2023-09-25
chore: Drop Fair{ParSet,Container} forward decls

ChristianTackeGSI:
and add some override, while in those files.

---

Checklist:

### DIFF
--- a/cmake/modules/CheckCompiler.cmake
+++ b/cmake/modules/CheckCompiler.cmake
@@ -123,7 +123,10 @@ endif (CMAKE_SYSTEM_NAME MATCHES Linux)
 
 
 if (CMAKE_SYSTEM_NAME MATCHES Darwin)
-   EXEC_PROGRAM("sw_vers -productVersion | cut -d . -f 1-2" OUTPUT_VARIABLE MAC_OS_VERSION)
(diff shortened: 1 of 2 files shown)


fuhlig1:
@ChristianTackeGSI,

why do you delete the forward declarations? The classes are used in the code, so in my opinion they belong there even if they come by chance via the include statement.

ChristianTackeGSI:
Most places use those classes in overriden methods. So that original declaration of that method must have the right declarations already. So those declarations do not come "by chance".

And in the examples, we should not use forward declarations of main FairRoot classes. We should use the correct `#include`s in those cases: Users should learn to use correct includes and not forward declarations for FairRoot classes. For example we could switch classes to `using`s with the exact same API. But usings and classes are not the same.

YanzhaoW:
a subtle comment: don't default a destructor if it can be removed.

There is always a preference for "Rule of 0", which means no definition of any special member functions (destructor, move constructor, copy constructor, move assignment, copy assignment). If any of them need to be defined, they should be defined all together. (Rule of 5)

ChristianTackeGSI:
Moved the `override` and `=default` into its own commit.

PR #1458, opened 2023-10-06
chore: Drop useless `return;`

ChristianTackeGSI:
`return;` at the end of a function is quite pointless.

Also fix some C-Style Casts

---

Checklist:

### DIFF
--- a/examples/MQ/pixelAlternative/src/PixelAltFindHits.cxx
+++ b/examples/MQ/pixelAlternative/src/PixelAltFindHits.cxx
@@ -188,7 +188,10 @@ void PixelAltFindHits::GetParList(TList* tempList)
 void PixelAltFindHits::InitMQ(TList* tempList)
 {
     LOG(info) << "********************************************** PixelAltFindHits::InitMQ()";
-    fDigiPar = (PixelDigiPar*)tempList->FindObject("PixelDigiParameters");
+    fDigiPar = dynamic_cast<PixelDigiPar*>(tempList->FindObject("PixelDigiParameters"));
+    if (!fDigiPar) {
+        throw std::runtime_error("no PixelDigiParameters");
+    }
 
     fFeCols = fDigiPar->GetFECols();
     fFeRows = fDigiPar->GetFERows();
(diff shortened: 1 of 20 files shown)


examples/MQ/pixelAlternative/src/PixelAltFindHits.cxx
```diff
@@ -183,14 +183,12 @@ void PixelAltFindHits::GetParList(TList* tempList)
...
     fDigiPar = new PixelDigiPar("PixelDigiParameters");
     tempList->Add(fDigiPar);
-
-    return;
 }
 
 void PixelAltFindHits::InitMQ(TList* tempList)
 {
     LOG(info) << "********************************************** PixelAltFindHits::InitMQ()";
-    fDigiPar = (PixelDigiPar*)tempList->FindObject("PixelDigiParameters");
```
fuhlig1, line 193: I think in the Init there should be a dynamic_cast with a check afterwards. We get an object by name which actually can be anything.
 
This statement holds for all usage of "getting by name".

ChristianTackeGSI: Thanks for noting this! You're right!

I added checks for all the ones I have touched.

examples/MQ/pixelAlternative/src/PixelAltFindHits.cxx
```diff
@@ -215,10 +214,12 @@ void PixelAltFindHits::ExecMQ(TList* inputList, TList* outputList)
     //  << "," << outputList->GetName() << "), Event " << fTNofEvents; LOG(info) <<
     //  "********************************************** PixelAltFindHits::ExecMQ(), Event " << fTNofEvents; LOG(info) <<
     //  "h" << FairLogger::flush;
-    fDigis = (TClonesArray*)inputList->FindObject("PixelDigis");
+    fDigis = dynamic_cast<TClonesArray*>(inputList->FindObject("PixelDigis"));
```
fuhlig1, line 217: For the digis I would propose to check also if the contained data class is the expected one. Later on we do a static_cast to this type.

fDigis = dynamic_cast<TClonesArray*>(inputList->FindObject("PixelDigis"));
  if (fDigis) {
    // Check if the stored data class has the right type  
    TString objectName = fDigis->GetClass()->GetName();
    if (0 != objectName.CompareTo("PixelDigi")) {
      throw std::runtime_error("no PixelDigis");
    }
else {
   throw std::runtime_error("no PixelDigi container");
}

ChristianTackeGSI: I don't like the string based things. I tried with `InheritsFrom` (which [TCollection::AssertClass](https://root.cern/doc/v628/classTCollection.html#a54d17025bcc640ac5a4f0ffa64cc819a) uses internally).

PR #1461, opened 2023-11-13
fix: Use TKey::ReadObject<T> instead of ReadObj

ChristianTackeGSI:
`TKey::ReadObj` returns an owning pointer.
Using dynamic_cast and ignoring the "not castable" case can leak memory.

Also one `std::unique_ptr`

See: https://root-forum.cern.ch/t/possible-leak-with-dynamic-cast-and-tkey-readobj/56799
See: https://github.com/root-project/root/pull/13931

---

Checklist:


fuhlig1 (approved):
The changes looks reasonable for me.

Issue #1462, opened 2023-11-16
FileSink is not closed if FairRunOnline::Run() is skipped and never closed for FairRunAna


YanzhaoW:
**Describe the bug**
FileSink is not closed when FairRunOnline::Run() function is skipped resulting in **programming crushing** when ROOT starts up clean up processes. (Never closed for FairRunAna)

**To Reproduce**
Use any examples with FileSink and comment out the FairRun::Run() calling.

**Causes**

For FairRunOnline:

https://github.com/FairRootGroup/FairRoot/blob/fa7108b45684c729c2e88b12d7fbf4bbb5a4dc65/online/steer/FairRunOnline.cxx#L338-L354

For FairRunAna:

https://github.com/FairRootGroup/FairRoot/blob/fa7108b45684c729c2e88b12d7fbf4bbb5a4dc65/base/steer/FairRunAna.cxx#L599-L611

This function is never called.

Similar for FairRunSim.

**Logs / Screenshots**

<details>
  <summary>
      Here is the stacktrace if anyone is interested
  </summary>

```text
#0  0x00007f7527d3f2d7 in __GI___waitpid (pid=28130, stat_loc=stat_loc
entry=0x7ffc17b31a08, options=options
entry=0) at ../sysdeps/unix/sysv/linux/waitpid.c:30
#1  0x00007f7527cbd70f in do_system (line=<optimized out>) at ../sysdeps/posix/system.c:149
#2  0x00007f75296f6913 in TUnixSystem::Exec (shellcmd=<optimized out>, this=0x55ae1e3537c0) at /home/uhlig/containers/home/debian10/software/fairsoft/nov22p1/build/Source/root/core/unix/src/TUnixSystem.cxx:2104
#3  TUnixSystem::StackTrace (this=0x55ae1e3537c0) at /home/uhlig/containers/home/debian10/software/fairsoft/nov22p1/build/Source/root/core/unix/src/TUnixSystem.cxx:2395
#4  0x00007f75296f8c55 in TUnixSystem::DispatchSignals (this=0x55ae1e3537c0, sig=kSigSegmentationViolation) at /home/uhlig/containers/home/debian10/software/fairsoft/nov22p1/build/Source/root/core/unix/src/TUnixSystem.cxx:3615
#5  <signal handler called>
#6  0x00007f7527cfc0af in _int_malloc (av=av
entry=0x7f7527e33c40 <main_arena>, bytes=bytes
entry=64) at malloc.c:3678
[...]

PR #1464, opened 2023-11-17
Add FinishTask to FairSource

YanzhaoW:
Add an inheritance interface to FairSource such that certain actions can be taken in the end of the run.

---

Checklist:


YanzhaoW:
@ChristianTackeGSI  Something wrong in CI?

I got an error saying cmake version is too old in CI machine.
 > CMake Error at CMakeLists.txt:8 (cmake_minimum_required):
  CMake 3.18 or higher is required.  You are running version 3.17.5

ChristianTackeGSI:
> @ChristianTackeGSI Something wrong in CI?
> 
> I got an error saying cmake version is too old in CI machine.
> 
> > CMake Error at CMakeLists.txt:8 (cmake_minimum_required):
> > CMake 3.18 or higher is required.  You are running version 3.17.5

We have some issues with some of our CI container images. @dennisklein wanted to update them. Seems this hasn't happened yet.

Please just ignore them. If some succeed (especially the ubuntu-rolling one), then everything is fine for now.

ChristianTackeGSI (commented):
I am not against this. (And I like virtual functions being private!). Having hooks for things is great.

That said, I have two issues with this:
1. The "*at the end of the run*" argument. If you want to do anything at the end of the run, what about deriving from FairTask a dedicated MyAtEndofRun class that overrides `FairTask::Finish()`?

   All the places that call `GetSource()->Finish();` are preceded by `fTask->FinishTask();` (which calls `FairTask::Finish()`). So it should be the right place for your "end of the run" idea?

   And if you want to do something "run specific", it should not be hooked to the Source. A FairSource::Finish should do something specific to the source.

3. As you intend to derive from FairSource anyways, you could alternatively override `FairSource::Close()` (and call the base class' `Close` at the end of your `Close()`). So if you intend to do something about the source just before it's getting closed, that could be your place.

YanzhaoW:
Thanks very much for your comment.

> The "at the end of the run" argument. If you want to do anything at the end of the run, what about deriving from FairTask a dedicated MyAtEndofRun class that overrides FairTask::Finish()

In my case, it is the readers in the FairSource that are doing the heavy-lifting list mode data analysis instead of FairTask. Actually no FairTask is needed when reading the list mode data. So in the end, all computation results are obtained in those readers along with histograms during the analysis. 

> As you intend to derive from FairSource anyways, you could alternatively override FairSource::Close() (and call the base class' Close at the end of your Close()). So if you intend to do something about the source just before it's getting closed, that could be your place.

So there is a problem that when those histograms should be written to file and whether the target files are available at that moment. My first attempt is of course writing histogram during `FairSource::Close()` to `FairRootFileSink`, which requires sink file to be ready for writing for that moment. But sometimes I got segmentation fault saying sink file is not available. But in the code of FairRunOnline, FairSource::Close() is actually be called before the FairSink::Close(). So my current guess is somehow FairSource::Close() is called twice, one before FairSink::Close and another after. The one after causes the segmentation fault.

And in my opinion, this is also a design issue because there is no logic reason why FairSource::Close() must be called before FairSink::Close() and we shouldn't impose this non-trivial restriction neither. Close() function should just be used to shutdown the status of each resources and the shutdown of one resource shouldn't be dependent on other resources. If we still need to use one resource, it should be done before any shutdown starts, guaranteeing the safety of the action.

Please tell me if I need more elaboration.

fairroot/base/steer/FairRunAna.cxx
```diff
@@ -551,6 +551,7 @@ void FairRunAna::TerminateRun()
 {
     fRootManager->StoreAllWriteoutBufferData();
     fTask->FinishTask();
+    GetSource()->Finish();
```
ChristianTackeGSI, line 554: And this one is the one where I wonder, whether Close is called at all.

This is in `FairRunAna::TerminateRun`. And the only place I could find that calls into `TerminateRun` is in FairAnaSelector, which is only used in FairRunAnaProof. So it looks like this is never called. I again wonder, where `GetSource()->Close()` gets really called?

ChristianTackeGSI (commented):
Just one thing ahead: I am relatively new to FairRoot and still learning. So if I miss something, just let me know!

I think, we should have a deeper look into how `Close` is called for the Source.

We have a suspected double Close on the one side. And even a missed close on the other side.

> So my current guess is somehow FairSource::Close() is called twice, one before FairSink::Close and another after (We had lots of double deleting. So double closing could be very possible). The one after causes the segmentation fault.

Could you have a look, if this really happens? Maybe add a LOG(DEBUG) to your Close? (would be really interesting to learn how this happens!)

I found one suspicious Close():
https://github.com/FairRootGroup/FairRoot/blob/71042508e6d738d3a56456599bd235e8b189535b/fairroot/online/source/FairLmdSource.cxx#L169

ChristianTackeGSI:
Hi @YanzhaoW,

are you using FairLmdSource? If so, could you take a look at #1465?

YanzhaoW:
Sure, I will check it out. 

I don't know what's the best way to fix this currently. Closed function should always be used in a dtor with the correct ownership. Maybe a total re-structure?

ChristianTackeGSI:
Many thanks for taking a look! :-)

We can't call `Close` directly from the destructor. Calling a virtual member function from a constructor doesn't do what you would expect. So each destructor should care for the member variables / resources of its class. That's why I invented this private `CloseLmd`, so that this code can be reused from the destructor and `Close`.

That said, yes, all of this needs another look. That's why I started with #1465.

YanzhaoW:
> Calling a virtual member function from a constructor doesn't do what you would expect.

You mean "destructor" instead? I don't quite understand here. If so, what kind of behaviour is unexpected?

fairroot/base/steer/FairRunAna.cxx
YanzhaoW: The TerminateRun is never called (see this issue #1462)

GetSource()->Close() is only called in FairRunOnline::Finish(), which is called in FairRun::Run(). It won't be called neither if Run() is skipped.

ChristianTackeGSI: Let's take a deeper look at #1462 first, please!

PR #1465, opened 2023-11-20
chore: Improve closing on FairLmdSource

ChristianTackeGSI:
`Close` is a virtual method. It's getting called internally in a scenario where only FairLmdSource specific resources should be closed.

Also the desctructor did not close.

This was triggered by a comment here: https://github.com/FairRootGroup/FairRoot/pull/1464#pullrequestreview-1738378541

---

Checklist:

### DIFF
--- a/fairroot/online/source/FairLmdSource.cxx
+++ b/fairroot/online/source/FairLmdSource.cxx
@@ -54,6 +54,7 @@ FairLmdSource::FairLmdSource(const FairLmdSource& source)
 
 FairLmdSource::~FairLmdSource()
 {
+    CloseLmd();
     fFileNames->Delete();
     delete fFileNames;
 }
@@ -121,6 +122,8 @@ Bool_t FairLmdSource::Init()
 
 Bool_t FairLmdSource::OpenNextFile(TString fileName)
 {
+    CloseLmd();
+
     Int_t inputMode = GETEVT__FILE;
     fxInputChannel = new s_evt_channel;
     void* headptr = &fxInfoHeader;
@@ -133,6 +136,8 @@ Bool_t FairLmdSource::OpenNextFile(TString fileName)
 
     if (status) {
         LOG(error) << "File " << fileName << " opening failed.";
+        delete fxInputChannel;
+        fxInputChannel = nullptr;
         return kFALSE;
     }
 
@@ -166,7 +171,7 @@ Int_t FairLmdSource::ReadEvent(UInt_t)
         }
 
         if (GETEVT__NOMORE == status) {
-            Close();
+            CloseLmd();
         }
 
         TString name = (static_cast<TObjString*>(fFileNames->At(fCurrentFile)))->GetString();
@@ -236,7 +241,17 @@ Int_t FairLmdSource::ReadEvent(UInt_t)
 
 void FairLmdSource::Close()
 {
+    CloseLmd();
+}
+
+void FairLmdSource::CloseLmd()
+{
+    if (!fxInputChannel) {
+        return;
+    }
     f_evt_get_close(fxInputChannel);
+    delete fxInputChannel;
+    fxInputChannel = nullptr;
     Unpack(reinterpret_cast<Int_t*>(fxBuffer), sizeof(s_bufhe), -4, -4, -4, -4, -4);
     fCurrentEvent = 0;
 }
--- a/fairroot/online/source/FairLmdSource.h
+++ b/fairroot/online/source/FairLmdSource.h
@@ -65,6 +65,9 @@ class FairLmdSource : public FairMbsSource
 
     FairLmdSource& operator=(const FairLmdSource&);
 
+  private:
+    void CloseLmd();
+
     ClassDefOverride(FairLmdSource, 0);
 };
 


YanzhaoW:
Hi, thanks for the PR.

There are few problems:
1. The overridden Closed function from Fairlmd should be called from its base class (dtor of bass class) instead of itself.
2. The overridden Closed function should be a private member function. I don't understand why you have private Closelmd and a public Close instead of just making Close private.

ChristianTackeGSI:
> The overridden Closed function from Fairlmd should be called from its base class (dtor of bass class) instead of itself.

That does not work.

Inside the base class destructor the object is of type base-class. So any calls to any virtual functions in the dtor go to the virtual methods of that base class (or its parents).

That's basically, why calling virtual functions in the dtor / ctor (there it's the same story) is a bad idea, because it does not do what you would think it should do.

There's even a clang-tidy warning for this: [clang-analyzer-optin.cplusplus.VirtualCall](https://clang.llvm.org/extra/clang-tidy/checks/clang-analyzer/optin.cplusplus.VirtualCall.html). And yes, we have [enabled this one in our CI](https://alfa-ci.gsi.de/job/FairRootGroup/job/FairRoot/view/change-requests/job/PR-1465/4/clang-tidy/category.-2042718063/). And yes, we should fix them…


> The overridden Closed function should be a private member function. I don't understand why you have private Closelmd and a public Close instead of just making Close private.

`Close` is a public API of FairSource. So if we want to make it private, we should declare that a breaking change.

And IMHO it is okay for a FairSource to have Open and Close functions. The Close one can be virtual (yes, it would be nicer, if the virtual functions would be private/protected and the public interface would be another member functions, but we don't have that yet), so that FairRun can close the source when it's done, but doesn't have to destruct it yet.

So the idea is:
* Each class can have a private (non-virtual!) helper function to cleanup resources / close the resources.
* Call that helper from the dtor.
* Override `Close`, call your helper
  And call the parent class' `Close` (if the parent class has one).
  (So yes, we probably should give the base class an empty "default" `Close`.)

YanzhaoW:
> Inside the base class destructor the object is of type base-class. So any calls to any virtual functions in the dtor go to the virtual methods of that base class (or its parents).

You are right! I always thought calling a virtual function in dtor is same as in any other member functions. Glad I learned this new thing.

> And IMHO it is okay for a FairSource to have Open and Close functions. The Close one can be virtual (yes, it would be nicer, if the virtual functions would be private/protected and the public interface would be another member functions, but we don't have that yet), so that FairRun can close the source when it's done, but doesn't have to destruct it yet.

Even though we can't call the Close in FairSource dtor, we could still call the Fair::Close in the dtor of its owner, as is also a better place to call it.

So in my opinion, what's the correct way is to have a public non-virtual close member function in the base class (in our case, `FairSource`) and also another virtual private member function, for example, `Close_imp()`, which is called in `Close`. Then in the derived class, we only override this private member function `Close_imp()` (without overloading public method `Close()`). If we decides `FairSource` is owned by `FairRun` (I'm still not sure whether is FairRun or FairRooManager), in the dtor of `FairRun`, we call the public method `FairSource::Close`, which actually executes the `Close_imp()` from the true type of the "source" object.

fairroot/online/source/FairLmdSource.cxx
```diff
@@ -54,6 +54,7 @@ FairLmdSource::FairLmdSource(const FairLmdSource& source)
 
 FairLmdSource::~FairLmdSource()
 {
+    CloseLmd();
```
YanzhaoW, line 57: This step should be executed by its owner.

fairroot/online/source/FairLmdSource.cxx
```diff
@@ -236,7 +241,17 @@ Int_t FairLmdSource::ReadEvent(UInt_t)
 
 void FairLmdSource::Close()
 {
+    CloseLmd();
+}
```
YanzhaoW, line 245: This should be implemented in the base class `FairSource`:
```cpp
void FairSource::Close() {Close_imp();}
```

fairroot/online/source/FairLmdSource.h
```diff
@@ -65,6 +65,9 @@ class FairLmdSource : public FairMbsSource
 
     FairLmdSource& operator=(const FairLmdSource&);
 
+  private:
+    void CloseLmd();
```
YanzhaoW, line 69: This should be an overridden private virtual function.

YanzhaoW (commented):
Following my comment above.

fairroot/online/source/FairLmdSource.cxx
ChristianTackeGSI: Well, no.

I don't think, the "owner" should be resposible for calling Close. It could be considered "being a good citizen", but it should not be strictly needed.

```cpp
FairLmdSource mysource;
mysource.Open(…);
if (true)
  throw std::runtime_exception("Something else went wrong");
mysource.Close();
```

This should not leak resources! (See "RAII printciple").

So yes, This is some overhead for the developer of a FairSource/Sink. But that's how C++ is like.

fairroot/online/source/FairLmdSource.cxx
ChristianTackeGSI: Can we postpone this for another PR?

fairroot/online/source/FairLmdSource.h
ChristianTackeGSI: Let's please postpone this for another PR as well.

kresan (approved):
Much cleaner closing now. Looks good.

ChristianTackeGSI:
Should we backport this to 18.8? (or even 18.6)?

PR #1466, opened 2023-11-20
Fix the error that ROOT header files can't be found

YanzhaoW:
ROOT header files can't be found because of an typo in FairMacros.cmake.

Fixed by changing ROOT_INCLUDE_DIRS to ROOT_INCLUDE_DIR

Edit:
It seems that this was changed deliberately by PR #1119.

---

Checklist:


YanzhaoW:
Hi, @ChristianTackeGSI 

I'm trying to use the dev version of FairRoot in R3BRoot. We are still using the MODULE mode to find the ROOT, which runs through FindRoot.cmake. In this case, the ${ROOT_INCLUDE_DIR} is defined properly but ${ROOT_INCLUDE_DIRS} is empty.

What is the suggested way to include ROOT package?

ChristianTackeGSI:
FairRoot dev has a major CMake rewrite. See the [Chabgelog](https://github.com/FairRootGroup/FairRoot/blob/dev/CHANGELOG.md#190-unreleased---2023-xx-xx).

Handling this in R3BRoot in a way that is compatible with FairRoot 18.x and 19.x at the same time is not trivial!
I once did this for PandaRoot, but I don't have a recommended way of doing it nicely / right now. Not to mention 19.x is still a moving target (also I would hope we'll release sooner than later).

The idea is that FairRoot 18.8 has most C++ changes and 19.0 has only CMake changes. I don't know, how close we are to this idea still. But once it was like this.

YanzhaoW:
Great to hear this change.

If using CONFIG mode to find ROOT is preferred, should we just use cmake targets without global variable `${SYSTEM_INCLUDE_LIBRARY}` or `${BASE_INCLUDE_LIBRARY}` that include root include dir? [This](https://cliutils.gitlab.io/modern-cmake/chapters/packages/ROOT.html) is exactly what ROOT suggests. 

In this case, we should keep it as ${ROOT_INCLUDE_DIR} for the MODULE mode and do not use global ${ROOT_INCLUDE_DIRS} if users choose to use CONFIG mode.

ChristianTackeGSI:
Right, the idea is to use exported targets from ROOT with their include paths. So one should only need to properly declare dependencies on ROOT::Core/etc and be done without needing to take care about include paths.

YanzhaoW:
Hi

Sorry for the late reply. 

> Right, the idea is to use exported targets from ROOT with their include paths. So one should only need to properly declare dependencies on ROOT::Core/etc and be done without needing to take care about include paths.

It's great the hear that again. But I always have a problem with ROOTCINT when generating dictionary. The problem is that it seems ROOTCINT can't use the included path from the CMake targets. Every time I try to compile a library, which cmake targets has dependencies, rootcint seems to always fail to find the header files:
```cmake
set(DEPENDENCIES R3BBase)
GENERATE_LIBRARY()
``` 
even though the header file does exist inside an include folder of `R3BBase` cmake target. Normally this has to go round with this:
```cmake
include_directories(folder)
```
which sets the global include_directories for all targets (including those who even don't need this folder). This works but would fail if  `R3BBase` needs another new folder. The fix requires all libraries which are depending on `R3BBase` to have this new folder included in `include_directories` as well. All of these efforts is just to make rootcint generate dictionaries.  The generation of the library actually doesn't need this because cmake targets can automatically propagate the new properties to all depended targets.

Is there any way to make rootcint work with cmake targets? Or even is there actually a plan to implement it in FairMacros?

YanzhaoW:
So I tried the solution mentioned above. It works (it has actually been documented in [ROOT manual](https://cliutils.gitlab.io/modern-cmake/chapters/packages/ROOT.html))

Please find the example in [this PR](https://github.com/R3BRootGroup/R3BRoot/pull/911) proposed in R3BRoot. The boilerplate part is:
```cmake
add_library(R3BSource SHARED ${SRCS})
target_include_directories(R3BSource PUBLIC ${DIRS} ${CMAKE_CURRENT_SOURCE_DIR})
target_link_libraries(R3BSource PUBLIC R3BData MbsAPI Ucesb)
root_generate_dictionary(R3BSource_dict ${HEADERS} MODULE R3BSource LINKDEF SourceLinkDef.h)
```
With this ROOT can automatically find those needed include dirs with using `include()`.

YanzhaoW:
This PR has been covered with the release of FairRoot 19.

ChristianTackeGSI:
Thanks for taking care and cleaning up! :-)

PR #1468, opened 2023-12-02
Remove calling Close method in FairRootManager/FairRun

YanzhaoW:
- ~~Deprecation of Close method in FairSource/Sink.~~
- Move the Close executions in the derived classes to their dtor (if needed).
- Remove callings of the Close methods in FairRootManager and FairRun.

---

Checklist:

### DIFF
--- a/examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
+++ b/examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
@@ -106,8 +106,7 @@ Int_t PixelDigiBinSource::ReadEvent(UInt_t i)
     for (Int_t idata = 0; idata < head[3]; idata++) {
         LOG(debug) << "    --/" << idata << "/-->    " << dataCont[idata * dataSize + 0] << " / "
                    << dataCont[idata * dataSize + 1] << " / " << dataCont[idata * dataSize + 2] << " / "
-                   << dataCont[idata * dataSize + 3] << " / "
-                   << " 0.";
+                   << dataCont[idata * dataSize + 3] << " / " << " 0.";
         new (fDigis[fNDigis]) PixelDigi(-1,
                                         dataCont[idata * dataSize + 0],
                                         dataCont[idata * dataSize + 1],
@@ -138,11 +137,12 @@ Bool_t PixelDigiBinSource::ActivateObject(TObject** obj, const char* BrName)
     return kTRUE;
 }
 
-void PixelDigiBinSource::Close() { fInputFile.close(); }
-
 void PixelDigiBinSource::Reset() {}
 
-Int_t PixelDigiBinSource::CheckMaxEventNo(Int_t /*EvtEnd*/) { return -1; }
+Int_t PixelDigiBinSource::CheckMaxEventNo(Int_t /*EvtEnd*/)
+{
+    return -1;
+}
 
 void PixelDigiBinSource::FillEventHeader(FairEventHeader* feh)
 {
--- a/examples/MQ/pixelDetector/src/PixelDigiBinSource.h
+++ b/examples/MQ/pixelDetector/src/PixelDigiBinSource.h
@@ -33,7 +33,6 @@ class PixelDigiBinSource : public FairSource
     Bool_t Init() override;
 
     Int_t ReadEvent(UInt_t i = 0) override;
-    void Close() override;
     void Reset() override;
     Bool_t SpecifyRunId() override
     {
@@ -77,6 +76,9 @@ class PixelDigiBinSource : public FairSource
 
     PixelDigiBinSource(const PixelDigiBinSource&);
     PixelDigiBinSource& operator=(const PixelDigiBinSource&);
+
+    // Private member functions:
+    void Close() override { fInputFile.close(); }
 
     ClassDefOverride(PixelDigiBinSource, 1);
 };
--- a/fairroot/base/sink/FairRootFileSink.cxx
+++ b/fairroot/base/sink/FairRootFileSink.cxx
@@ -202,7 +202,7 @@ void FairRootFileSink::TruncateBranchNames(TBranch* b, TString ffn)
 
 void FairRootFileSink::Close()
 {
-    if (fRootFile) {
+    if (fRootFile != nullptr) {
         fRootFile->Close();
     }
 }
@@ -251,7 +251,8 @@ void FairRootFileSink::WriteFolder()
     }
 }
 
-namespace impl {
+namespace impl
+{
 // a helper function to demangle a type_name
 inline std::string demangle(const char* name)
 {
--- a/fairroot/base/sink/FairSink.h
+++ b/fairroot/base/sink/FairSink.h
@@ -40,7 +40,7 @@ class FairSink
     virtual ~FairSink();
 
     virtual Bool_t InitSink() = 0;
-    virtual void Close() = 0;
+    virtual void Close() {}
     virtual void Reset() = 0;
 
     virtual Sink_Type GetSinkType() = 0;
--- a/fairroot/base/source/FairFileSource.h
+++ b/fairroot/base/source/FairFileSource.h
@@ -40,11 +40,10 @@ class FairFileSource : public FairFileSourceBase
     FairFileSource(const TString* RootFileName, const char* Title = "InputRootFile", UInt_t identifier = 0);
     FairFileSource(const TString RootFileName, const char* Title = "InputRootFile", UInt_t identifier = 0);
     // FairFileSource(const FairFileSource& file);
-    ~FairFileSource() override;
+    ~FairFileSource() override = default;
 
     Bool_t Init() override;
     Int_t ReadEvent(UInt_t i = 0) override;
-    void Close() override;
 
     /**Check the maximum event number we can run to*/
     Int_t CheckMaxEventNo(Int_t EvtEnd = 0) override;
@@ -65,9 +64,9 @@ class FairFileSource : public FairFileSourceBase
     void CreateNewFriendChain(TString inputFile, TString inputLevel);
     TTree* GetInTree() { return fInChain->GetTree(); }
     TChain* GetInChain() { return fInChain; }
-    void CloseInFile()
+    [[deprecated("Use Close() function")]] void CloseInFile()
     {
-        if (fRootFile) {
+        if (fRootFile != nullptr) {
             fRootFile->Close();
         }
     }
--- a/fairroot/base/source/FairFileSourceBase.h
+++ b/fairroot/base/source/FairFileSourceBase.h
@@ -54,6 +54,14 @@ class FairFileSourceBase : public FairSource
     /** list of folders from all input (and friends) files*/
     TObjArray fListFolder{16};   //!
 
+    // private virtual methods:
+    void Close() override
+    {
+        if (fRootFile != nullptr) {
+            fRootFile->Close();
+        }
+    }
+
     ClassDefOverride(FairFileSourceBase, 0);
 };
 
--- a/fairroot/base/source/FairMixedSource.h
+++ b/fairroot/base/source/FairMixedSource.h
@@ -43,11 +43,10 @@ class FairMixedSource : public FairFileSourceBase
                     const char* Title = "InputRootFile",
                     UInt_t identifier = 0);
     //  FairMixedSource(const FairMixedSource& file);
-    ~FairMixedSource() override;
+    ~FairMixedSource() override = default;
 
     Bool_t Init() override;
     Int_t ReadEvent(UInt_t i = 0) override;
-    void Close() override;
 
     /**Check the maximum event number we can run to*/
     Int_t CheckMaxEventNo(Int_t EvtEnd = 0) override;
--- a/fairroot/base/steer/FairRootManager.h
+++ b/fairroot/base/steer/FairRootManager.h
@@ -18,12 +18,12 @@
 #include <TRefArray.h>   // for TRefArray
 #include <TString.h>     // for TString, operator<
 #include <fairlogger/Logger.h>
+#include <list>
 #include <map>   // for map, multimap, etc
 #include <memory>
 #include <string>
 #include <type_traits>   // is_pointer, remove_pointer, is_const, remove...
 #include <typeinfo>
-#include <list>
 
 class BinaryFunctor;
 class FairEventHeader;
@@ -69,7 +69,7 @@ class FairRootManager : public TObject
 
     void CloseSink()
     {
-        if (fSink) {
+        if (fSink != nullptr) {
             fSink->Close();
         }
     }
--- a/fairroot/base/steer/FairRunAna.cxx
+++ b/fairroot/base/steer/FairRunAna.cxx
@@ -558,7 +558,6 @@ void FairRunAna::TerminateRun()
     //   cout << ">>>------------------------------------------------<<<" << endl;
     fRootManager->LastFill();
     fRootManager->Write();
-    fRootManager->CloseSink();
 }
 //_____________________________________________________________________________
 
--- a/fairroot/base/steer/FairRunSim.cxx
+++ b/fairroot/base/steer/FairRunSim.cxx
@@ -323,9 +323,6 @@ void FairRunSim::Run(Int_t NEvents, Int_t)
 {
     fApp->RunMC(NEvents);
     fWasMT = fApp->GetIsMT();
-    if (fSink) {
-        fSink->Close();
-    }
 }
 
 void FairRunSim::SetField(FairField* field)
--- a/fairroot/online/steer/FairRunOnline.cxx
+++ b/fairroot/online/steer/FairRunOnline.cxx
@@ -309,9 +309,6 @@ void FairRunOnline::Finish()
     fTask->FinishTask();
     fRootManager->LastFill();
     fRootManager->Write();
-    GetSource()->Close();
-
-    fRootManager->CloseSink();
 }
 
 void FairRunOnline::ActivateHttpServer(Int_t refreshRate, Int_t httpServer)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
     * `virtual bool operator<(const FairTimeStamp* rValue) const` changed to `bool operator<(const FairTimeStamp& rValue) const`
   * FairModule::svList is gone. This was never intended as a public
     API.
+  * Remove calling `Close()`
+    * `FairSink::Close()` and `FairSource::Close()` are no longer called in FairRoot.
+    * Derived classes from `FairSink` and `FairSource` should close the resources automatically in their destructors.
 
 ### Deprecations
 
--- a/examples/MQ/pixelDetector/src/FairOnlineSink.h
+++ b/examples/MQ/pixelDetector/src/FairOnlineSink.h
@@ -33,7 +33,6 @@ class FairOnlineSink : public FairSink
     FairOnlineSink& operator=(const FairOnlineSink&) = delete;
 
     Bool_t InitSink() override { return kTRUE; }
-    void Close() override {}
     void Reset() override {}
 
     Sink_Type GetSinkType() override { return kONLINESINK; }
--- a/examples/MQ/pixelDetector/src/PixelDigiSource.cxx
+++ b/examples/MQ/pixelDetector/src/PixelDigiSource.cxx
@@ -136,11 +136,12 @@ Bool_t PixelDigiSource::ActivateObject(TObject** obj, const char* BrName)
     return kTRUE;
 }
 
-void PixelDigiSource::Close() { fInputFile.close(); }
-
 void PixelDigiSource::Reset() {}
 
-Int_t PixelDigiSource::CheckMaxEventNo(Int_t /*EvtEnd*/) { return -1; }
+Int_t PixelDigiSource::CheckMaxEventNo(Int_t /*EvtEnd*/)
+{
+    return -1;
+}
 
 void PixelDigiSource::FillEventHeader(FairEventHeader* feh)
 {
--- a/examples/MQ/pixelDetector/src/PixelDigiSource.h
+++ b/examples/MQ/pixelDetector/src/PixelDigiSource.h
@@ -33,7 +33,6 @@ class PixelDigiSource : public FairSource
     Bool_t Init() override;
 
     Int_t ReadEvent(UInt_t i = 0) override;
-    void Close() override;
     void Reset() override;
     Bool_t SpecifyRunId() override
     {
@@ -77,6 +76,8 @@ class PixelDigiSource : public FairSource
 
     PixelDigiSource(const PixelDigiSource&);
     PixelDigiSource& operator=(const PixelDigiSource&);
+
+    void Close() override { fInputFile.close(); }
 
     ClassDefOverride(PixelDigiSource, 1);
 };
--- a/fairroot/base/sim/FairMCApplication.cxx
+++ b/fairroot/base/sim/FairMCApplication.cxx
@@ -478,10 +478,6 @@ void FairMCApplication::FinishRunOnWorker()
     LOG(debug) << "FairMCApplication::FinishRunOnWorker: ";
 
     FinishRun();
-
-    // Close the sink on workers so that the execution of main FairRunSim->Run()
-    // finishes with the output files properly closed.
-    fRootManager->CloseSink();
 }
 
 //_____________________________________________________________________________
--- a/fairroot/base/sink/FairRootFileSink.h
+++ b/fairroot/base/sink/FairRootFileSink.h
@@ -38,7 +38,6 @@ class FairRootFileSink : public FairSink
     ~FairRootFileSink() override = default;
 
     Bool_t InitSink() override;
-    void Close() override;
     void Reset() override;
 
     Sink_Type GetSinkType() override { return kFILESINK; }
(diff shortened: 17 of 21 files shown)


YanzhaoW:
a sidenote:
I have rewrote the `FairFileSource` class in R3BRoot in [this PR](https://github.com/R3BRootGroup/R3BRoot/pull/871) (actually it's R3BFileSource, which is a copy-paste from FairFileSource). Maybe we could migrate the new class here?

fairroot/base/sink/FairRootFileSink.cxx
```diff
@@ -68,6 +68,13 @@ FairRootFileSink::FairRootFileSink(const TString RootFileName, const char* Title
     LOG(debug) << "FairRootFileSink created------------";
 }
 
+FairRootFileSink::~FairRootFileSink()
+{
+    if (fRootFile) {
+        fRootFile->Close();
+    }
+}
+
```
ChristianTackeGSI, line 77: I think, this is not needed, as `fRootFile` is a `std::unique_ptr<TFile>` and so it is properly `delete`d (and the dtor is being called).

fairroot/base/sink/FairSink.h
```diff
@@ -40,7 +40,7 @@ class FairSink
     virtual ~FairSink();
 
     virtual Bool_t InitSink() = 0;
-    virtual void Close() = 0;
+    [[deprecated("Please close your resources in derived class dtor")]] virtual void Close();
```
ChristianTackeGSI, line 43: 1. Well, sadly enough the deprecation message is only shown, when calling the function. So we should have something for people calling it.
2. If you don't say `= 0;`, you promise to implement it somewhere. Let's just implement it inline!

```suggestion
    [[deprecated("Close will go away, the dtor closes resources")]] virtual void Close() {}
```

fairroot/base/steer/FairRootManager.h
```diff
@@ -63,12 +63,6 @@ class FairRootManager : public TObject
     0 : Branch does not exist   */
     Int_t CheckBranch(const char* BrName);
 
-    void CloseSink()
-    {
-        if (fSink) {
-            fSink->Close();
-        }
-    }
```
ChristianTackeGSI, line 71: Sadly enough this is "public" API. So we need to deprecate it as well before removing it…

fairroot/base/steer/FairRunAna.cxx
```diff
@@ -558,7 +558,6 @@ void FairRunAna::TerminateRun()
     //   cout << ">>>------------------------------------------------<<<" << endl;
     fRootManager->LastFill();
     fRootManager->Write();
-    fRootManager->CloseSink();
```
ChristianTackeGSI, line 561: Sadly enough this is also part of our public API (at least in my view). So we should keep calling the `Close` member functions in the deprecation phase. So that people can actually continue using their stuff that depends on their Sink's/Source's Close member function being called.

Maybe we should replace this with something like this at least?
```cpp
    iff(fSink) {
        fSink->Close();
    }
```

examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
```diff
@@ -138,11 +143,12 @@ Bool_t PixelDigiBinSource::ActivateObject(TObject** obj, const char* BrName)
     return kTRUE;
 }
 
-void PixelDigiBinSource::Close() { fInputFile.close(); }
-
```
ChristianTackeGSI, line 142: Sadly enough this is IMHO also part of the public API: People can call Close and expect the resource to be closed. Of course they now get a new warning, that they should not call Close.

One could argue, that we just "leak" the resource until the dtor is called (hopefully) anyway…

I guess, I would appreciate it, if we keep this for a bit until we start removing things?

ChristianTackeGSI (changes requested):
First initial set of comments. I haven't looked deep enough yet.

fairroot/base/sink/FairRootFileSink.cxx
YanzhaoW: But TFile needs both closed and deleted, right? I don't see anywhere saying file will be closed automatically when deleted.

fairroot/base/sink/FairSink.h
YanzhaoW: Fixed

fairroot/base/steer/FairRootManager.h
YanzhaoW: Sure.

fairroot/base/steer/FairRunAna.cxx
YanzhaoW: But if we do this, we will always get deprecation warning no matter what. I think deprecation means asking people to fix something to get rid of the warning. But in this case, even they fix it and put the resource closing in the dtor, they still get this deprecation warning because the calling is guaranteed in the library code.

ChristianTackeGSI: The deprecation warning only happens in the FairRoot "library" code while compiling it. So normal users wont be hit by it.

And we should just stop that warning with a set of
```cpp
#pragma GCC diagnostic push
    iff(fSink) {
        fSink->Close();
    }
#pragma GCC diagnostic pop
```

Once we finally remove the Close member function, we will get a nice compile error on this, and we can remove all of this.

examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
```diff
@@ -38,6 +38,11 @@ PixelDigiBinSource::PixelDigiBinSource(TString inputFileName)
     LOG(debug) << "PixelDigiBinSource created------------";
 }
 
+PixelDigiBinSource::~PixelDigiBinSource()
+{
+    fInputFile.close();
+}
+
```
ChristianTackeGSI, line 45: `fInputFile` is a std::ifstream. So its dtor should handle that for us. :-)

ChristianTackeGSI (commented):
One other note.

examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
ChristianTackeGSI: Okay, please excuse! My fault!

In `examples/` we can of course perform everything we want (we even should), because examples are not public API. They show, how one should use things.

So really, here everything you did, was good!

My fault!

fairroot/base/sink/FairRootFileSink.cxx
```diff
@@ -200,13 +207,6 @@ void FairRootFileSink::TruncateBranchNames(TBranch* b, TString ffn)
     delete BrIter;
 }
 
-void FairRootFileSink::Close()
-{
-    if (fRootFile) {
-        fRootFile->Close();
-    }
-}
-
```
ChristianTackeGSI, line 209: But I think, this is public API still?

fairroot/base/steer/FairRootManager.h
```diff
@@ -63,12 +63,7 @@ class FairRootManager : public TObject
...
+    [[deprecated("Close will go away, the dtor closes resources")]] void CloseSink() {}
```
ChristianTackeGSI, line 66: Please keep the complete old version as well here.

fairroot/base/steer/FairRunSim.cxx
```diff
@@ -315,9 +315,6 @@ void FairRunSim::Run(Int_t NEvents, Int_t)
 {
     fApp->RunMC(NEvents);
     fWasMT = fApp->GetIsMT();
-    if (fSink) {
-        fSink->Close();
-    }
```
ChristianTackeGSI, line 320: This one should be kept IMHO as well.

fairroot/online/steer/FairRunOnline.cxx
```diff
@@ -309,9 +309,6 @@ void FairRunOnline::Finish()
     fTask->FinishTask();
     fRootManager->LastFill();
     fRootManager->Write();
-    GetSource()->Close();
-
-    fRootManager->CloseSink();
```
ChristianTackeGSI, line 314: And this one as well.

ChristianTackeGSI (commented):
More things to keep, IMHO.

Surround them with
```cpp
#pragma GCC diagnostic push
    …
#pragma GCC diagnostic pop
```

as needed.

fairroot/base/sink/FairRootFileSink.cxx
YanzhaoW: Ok, there is a Close method in the dtor of TFile.

examples/MQ/pixelDetector/src/PixelDigiBinSource.cxx
YanzhaoW: ok

fairroot/base/sink/FairRootFileSink.cxx
YanzhaoW: Yes, without the declaration here, `FairRootFileSink` still have Close public API, which falls back to the implementation of its base class.

karabowi:
I have tested FairRoot's private branch without FairSampler::Close() and FairSink::Close()
with PandaRoot. I did not observe any crash.

ChristianTackeGSI:
### Deprecations in General

For me a deprecation means, that
1. All old code stays as it is (as long as it can be considered public API). So we should not change too much code. So that users' code continues to work as without this PR!
2. Any public API that should not be used in the future by users, should be marked as deprecated (either via `[[deprecated]]`, or via comments, if the former is not possible).

### In FairRoot: Public API

Let's look at (1):

Public API is basically everything in `fairroot/`. So we should not change any code or behaviour there, basically. Of course we can do some simple stuff: Instead of calling `FairRootManager::CloseSink` (which calls `fSink->Close();`) we can call `fSink->Close();` directly. This does not change any actual behaviour.

When it comes to `examples/` we should go the full route! We should show to users what we expect their code to look like in the future.

### FairRoot: The confusing Close situation

In a cool world, the *old* situation would look like this:

```cpp
class FairSourceSink {
public:
    void CloseForUsers() { CloseImpl(); }
private:
    // Please override in your implementation
    virtual void CloseImpl() = 0;
private:
    // So that the library can call private member functions
    friend FairRun;
    friend FairRootManager;
    // Will be called internally by the library as a "call back like thing"
    void CloseForInternalUse() { CloseImpl(); }
};
```

Let's go step by step.

1. We want to deprecate `CloseForUsers`, so that users do not call it any more:

   ```cpp
   public:
       [[deprecated]] void CloseForUsers() { CloseImpl(); }
   ```

   Note: `CloseForInternalUse` is not deprecated. We want to continue calling it from the library, so that the CloseImpl is still called.

   Corollary: We would remove any calls to `CloseForUsers` from `examples/`!

2. We want to deprecate overriding `CloseImpl`:

   ```cpp
   private:
       // \deprecated Please stop overriding it!
       virtual void CloseImpl() {}
   ```

   Note 1: There is no point in marking it `override`. We want to deprecate overriding it. No need to deprecate calling it. It can only be called by us. All of our business.

   Note 2: We still continue calling `CloseForInternalUse` (to keep bevahiour exactly as it is)

   Note 3: We would still keep the overriding in `fairroot/`, because we did not yet remove `CloseForUsers()`. So code that depends on `CloseForUsers()` calling something should be happy.

   Corollary: We would remove any overrding of `CloseImpl` from `examples/`.

3. Remove `CloseForUsers`, keep everything else the same

4. Remove `CloseImpl`, and remove all of `CloseForInternalUse`.

That's my current idea on how this should look like.

The problem is now, that all of those three (`CloseImpl`, `CloseForInternalUse`, and `CloseForUsers`) are currently in one function `Close` to make things complex.

To maybe simplify things, we could temporarily invent:
```cpp
private
   void CloseForInternalUse() {
       // All of these pragmas, because we would call
       // CloseImpl (which is not deprecated internally),
       // but we have to call Close, which is deprecated
       // from (1).
#pragma GCC diagnostic push
       Close();
#pragma GCC diagnostic pop
   }
```

That's the whole problem here from my point of view.

YanzhaoW:
Thanks for the replies.

Sorry I'm very busy working on other things and don't have time to deal with this PR these days. But probably on this Friday I will read the replies in details and continue to fix the things here if needed.

YanzhaoW:
Hi,

I thought it over again. Yes, the procedures described above is probably the right way to go. But there is still something weird that I couldn't pinpoint what exactly it is in the beginning, especially when I see:

```cpp
private
   void CloseForInternalUse() {
       // All of these pragmas, because we would call
       // CloseImpl (which is not deprecated internally),
       // but we have to call Close, which is deprecated
       // from (1).
#pragma GCC diagnostic push
       Close();
#pragma GCC diagnostic pop
   }
```  

For me, we are kind of contradicting ourself when we enable a warning from a new deprecation but at the same time we also disable it. I guess the people who designed the "deprecation" in C++ also didn't expect the people to disable it at the same time using a compiler-specific macro (There are some FairRoot users using apple-clang).

Our main dispute is whether we should remove the calling of the `FairSink::Close()` or keep it in the public API of `FairRootManager`:

```cpp
void CloseSink()
{
    if (fSink) {
        fSink->Close();
    }
}
```

Here are my arguments why it should just be removed based on the following principles:

1. Deprecation should only be applied to public API.
2. Public API shouldn't be called internally.
3. Virtual functions should be private and shouldn't be called in the derived classes.

The point 2 and 3 are supported by C++ guidelines from Herb Sutter about [why virtual functions should be private](http://www.gotw.ca/publications/mill18.htm). In this articles, it says

> Prefer to use Template Method to make the interface stable and nonvirtual, while delegating customizable work to nonpublic virtual functions that are responsible for implementing the customizable behavior. After all, virtual functions are designed to let derived classes customize behavior; it's better to not let publicly derived classes also customize the inherited interface, which is supposed to be consistent.

It also implies here that the derived classes which overrides the virtual functions are just customization. For example, if we are designing an abstract class for a person wearing a T-shirt and pants. It's the job of the derived classes (User's classes) to decide what kind of T-shirt and pants are (e.g colors and sizes). But the logic connection between those objects are determined by the abstract class (the library implementer). Users shouldn't be concerned about in which order they are put up or whether they are put up or not. Suppose if we want to let the person wear T-shirt first instead of pants first, we could just change that order without doing any deprecation of the old order. Deprecation is used to notify the users. If it's irrelevant to users, it should not be deprecated but rather simply removed.

So now let's get back to `CLoseSink()`. Here the closing of the fSink is an internal logic for `FairRootManager`. Users of `FairRootManager` only interact with the public API `CloseSink` instead of `fSink` directly. For them, what's inside this public API shouldn't be their concerns and shouldn't break their code if changed. So if we need to discard the Close() method of `FairSink`, I think it's totally fine to be removed in `FairRootManager::CloseSink()` now.

YanzhaoW:
Hi, @ChristianTackeGSI 

I'm ready to restart working on this PR. But before I change any code, I would like to discuss the strategy and see whether we could reach some agreements:

1. No need to deprecate the Close() method of FairSink/FairSource
     We could follow the way how STL and ROOT deal with file sources: By default, closing a resource should be done automatically with dtor. But we could still provide a close api for some special needs from user side.  
2. Do not call any close in the internal implementation of `FairRootManager`. 
     Calling time of the Close function should only be determined by the derived class implementors and their users. It should be allowed that derived classes can manage the order of closings themselves.

ChristianTackeGSI:
For me (1) and (2) sound reasonable by now!

So I think, the first step is to make sure that all sinks/sources actually clean up their resources (close) in their dtors, unless they haven't done so?

Next step is to remove all Close calls in FairRootManager / FairRun. And document this as a feature or even a breaking change?

Does that sound right?

YanzhaoW:
Yes. Good, then I will proceed this PR very shortly.

> So I think, the first step is to make sure that all sinks/sources actually clean up their resources (close) in their dtors, unless they haven't done so?

I think so. I may double check it.

> Next step is to remove all Close calls in FairRootManager / FairRun. And document this as a feature or even a breaking change?

Yes, I will also put this change in the CHANGELOG

ChristianTackeGSI:
Awesome!

If this is easier for you, consider handling the first step in this PR and when it's done, create a new PR for the second step.

fairroot/base/sink/FairRootFileSink.cxx
```diff
@@ -202,9 +202,7 @@ void FairRootFileSink::TruncateBranchNames(TBranch* b, TString ffn)
 
 void FairRootFileSink::Close()
 {
-    if (fRootFile) {
-        fRootFile->Close();
-    }
+    fRootFile->Close();
```
ChristianTackeGSI, line 205: Please keep the old `if (fRootFile)`. Or can we guarantee that this pointer is always non-nullptr?

fairroot/base/source/FairFileSource.h
```diff
@@ -65,12 +63,7 @@ class FairFileSource : public FairFileSourceBase
     void CreateNewFriendChain(TString inputFile, TString inputLevel);
     TTree* GetInTree() { return fInChain->GetTree(); }
     TChain* GetInChain() { return fInChain; }
-    void CloseInFile()
-    {
-        if (fRootFile) {
-            fRootFile->Close();
-        }
-    }
+    [[deprecated("Use Close() function")]] void CloseInFile() {}
```
ChristianTackeGSI, line 66: As you only deprecate it, it should still have the old functionality. So please keep the full implementation.

fairroot/base/source/FairFileSourceBase.h
```diff
@@ -50,10 +50,15 @@ class FairFileSourceBase : public FairSource
 
   private:
     /**ROOT file*/
+    // TODO : Ownership should be clear. Use std::unique_ptr?
     fairroot::detail::maybe_owning_ptr<TFile> fRootFile;   //!
     /** list of folders from all input (and friends) files*/
     TObjArray fListFolder{16};   //!
 
+    // private virtual methods:
+    void Close() override { fRootFile->Close(); }
```
ChristianTackeGSI, line 59: Please add an `if (fRootFile)`. We can't gaurantee that it is non-nullptr!

fairroot/base/steer/FairRootManager.h
```diff
@@ -67,12 +67,8 @@ class FairRootManager : public TObject
...
     Int_t CheckBranch(const char* BrName);
 
-    void CloseSink()
-    {
-        if (fSink) {
-            fSink->Close();
-        }
-    }
+    // TODO: Redundant API? fSink is exposed and users can directly call FairSink::Close
+    void CloseSink() { fSink->Close(); }
```
ChristianTackeGSI, line 71: Please keep the `if (fSink)`. We can't guarantee that it is non-nullptr.

(I think, we even have some valid use cases for this one.)

fairroot/base/source/FairFileSourceBase.h
```diff
@@ -50,10 +50,15 @@ class FairFileSourceBase : public FairSource
 
   private:
     /**ROOT file*/
+    // TODO : Ownership should be clear. Use std::unique_ptr?
     fairroot::detail::maybe_owning_ptr<TFile> fRootFile;   //!
```
ChristianTackeGSI, line 54: The `maybe_owning_ptr` IS a unique_ptr.

It has a special mode where it degrades to a non-owning ptr.

So I don't see what your comment wants to say?

And we know that `maybe_owning_ptr` is nothing good in the long run (I think, it's ev en documented in that case).

ChristianTackeGSI (commented):
Thanks for the update!

I added a few comments.

examples/MQ/pixelDetector/src/PixelDigiBinSource.h
```diff
@@ -28,12 +28,10 @@ class PixelDigiBinSource : public FairSource
 {
   public:
     PixelDigiBinSource(TString inputFileName = "test.dat");
-    ~PixelDigiBinSource() override = default;
```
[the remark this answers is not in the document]
YanzhaoW: Should we follow "the rule of zero"? Here is a discussion about the rule of zero for derived classes: https://stackoverflow.com/questions/21637250/does-the-rule-of-zero-also-apply-for-classes-with-virtual-methods

fairroot/base/sink/FairRootFileSink.cxx
YanzhaoW: I also thought about this. The fRootFile is owned by the sink object anyway and it can't be nullptr when this method is called. Of course, someone could deliberately delete the pointer obtained from the getter. But in this crazy case, we should just let program crash instead of letting the program keep running and making the debugging difficult.

fairroot/base/source/FairFileSource.h
YanzhaoW: ok, I will change this.

fairroot/base/source/FairFileSourceBase.h
YanzhaoW: I see. When does it degrade to no-owning pointer and what controls the degradation? For me  compared to syd::unique_ptr, it's very unclear whether it's owned or not.

And I also don't quite why we have unique_ptr in other file source/sink classes but not in this class.

examples/MQ/pixelDetector/src/PixelDigiBinSource.h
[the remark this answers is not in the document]
ChristianTackeGSI: The second comment has an interesting part:

> `= default`ing special members does not violate the Ro0 in any way.

That's part one of my (our?) take on this.

Part two goes like this:
- The base class of a class with virtual methods *should* have a virtual dtor.
- How can we make sure in a derived class, that the base class has a virtual dtor? By using `override`!
- As we don't want to implement any special logic (Ro0?), let's `= default` it.

fairroot/base/sink/FairRootFileSink.cxx
ChristianTackeGSI: I would still prefer the "save" version, please.

fairroot/base/source/FairFileSourceBase.h
ChristianTackeGSI: > When does it degrade to no-owning pointer

Search for `non_owning` to find the cases, where it is set with a non-owning ptr. An owning-ptr can't ever be changed to a non-owning one directly. (You must `.release()` it first and then set it again as a `non_owning` one. Very, very explicitly.)

> and what controls the degradation? 

See the test suite for all the use cases: https://github.com/FairRootGroup/FairRoot/blob/dev/fairroot/tools/tests/maybe_owning.cxx

And also see the brief documentation: https://github.com/FairRootGroup/FairRoot/blob/1b5ae3e55bdbe71080c429bec96a53ef3f82a2c6/fairroot/tools/FairMemory.h#L48
It has all the relevant notes.

If you think, `maybe_owning_ptr` needs even more documentation, please write an issue or ping me on mattermost.

examples/MQ/pixelDetector/src/PixelDigiBinSource.h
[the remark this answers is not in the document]
YanzhaoW: I think except the constructor, we make special member functions default by not defining it. Otherwise, you should also default 5 member functions. Anyway, this is a stylish issue and I will change it according to your standard.

fairroot/base/sink/FairRootFileSink.cxx
YanzhaoW: Well, there is no concept of "safety" if the program is already used in an extremely unsafe way. The best way to deal with this case, IMO, is let the program crash as soon as possible such that the bug can be more easily found.

I also said something wrong in the last comment. If a user delete the pointer, `if(fRootFile)` won't help here because `fRootFile` is a non-nullptr pointer pointing to an invalid memory. And I don't think there is any way that a user can set `fRootFile` to be a nullptr. As long as the source class lives, `fRootFile` can never be nullptr.

fairroot/base/source/FairFileSourceBase.h
YanzhaoW: Sorry, even though it's very very explicit, I still don't know whether `fRootFile` is owned by `FairFileSourceBase`. This is important because if root file is not owned, we shouldn't close it in the `FairFileSourceBase`. Then we have to define `maybe_close` to address this "maybe" situation.

fairroot/base/source/FairMixedSource.h
```diff
@@ -43,11 +43,9 @@ class FairMixedSource : public FairFileSourceBase
                     const char* Title = "InputRootFile",
                     UInt_t identifier = 0);
     //  FairMixedSource(const FairMixedSource& file);
-    ~FairMixedSource() override;
```
ChristianTackeGSI, line 46: Please keep all the other (`=defaulted`) ctors as well. :-)

ChristianTackeGSI (commented):
This looks quite good now!

Please go over all changed files and update the copyright header to include 2024 (some probably already have).

And check the Conversations and resolve the remaining ones, please.

Oh, and consider rebasing to latest dev, please.

fairroot/base/source/FairMixedSource.h
ChristianTackeGSI: Can you please re-introduce all those removed dtors again? Should even reduce the diff-size. :wink:

YanzhaoW: I think I did with 
```c++
~FairMixedSource() override = default;
```
Isn't this what you suggested?

ChristianTackeGSI: I also meant the other ones. Like

```cpp
~PixelDigiSource() override = default;
~FairRootFileSink() override = default;
FairFileSource::~FairFileSource() = default;
```

YanzhaoW: Ok, I have changed according to your request.

ChristianTackeGSI (approved):
This is now all fine for me!

I would prefer @karabowi or @fuhlig1 to have a final look as well.

PR #1477, opened 2023-12-20
Use CTest instead of Dart module

fuhlig1:
The Dart module is deprecated sine CMake version 3.27. The CTest module provides the same functionality.

https://cmake.org/cmake/help/latest/module/Dart.html

---

Checklist:


dennisklein:
obsolete

PR #1478, opened 2023-12-21
Fix a runtime error with newer XCode versions

fuhlig1:
When using boost/program_options in a header file the generated dictinonary can't be loaded at runtime and the macro crashes. Guard the usage of boost/program_options in the header file such that code is only used during compilation but not when using cling.
The problem was already reported to the ROOT team and will not be fixed since the underlying problem is obviously in the libc++ shipped by Apple. (see https://github.com/root-project/root/issues/12762)

---

Checklist:

### DIFF
--- a/examples/simulation/Tutorial1/src/FairSimConfig.cxx
+++ b/examples/simulation/Tutorial1/src/FairSimConfig.cxx
@@ -1,13 +1,16 @@
-/********************************************************************************
- * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
- *                                                                              *
- *              This software is distributed under the terms of the             *
- *              GNU Lesser General Public Licence (LGPL) version 3,             *
- *                  copied verbatim in the file "LICENSE"                       *
- ********************************************************************************/
+/*********************************************************************************
+ *  Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ *                                                                               *
+ *              This software is distributed under the terms of the              *
+ *              GNU Lesser General Public Licence (LGPL) version 3,              *
+ *                  copied verbatim in the file "LICENSE"                        *
+ *********************************************************************************/
 #include "FairSimConfig.h"
 
 #include "FairLogger.h"
+
+#include <boost/program_options.hpp>
+namespace po = boost::program_options;
 
 #include <iostream>
 #include <vector>
@@ -15,18 +18,11 @@
 using std::string;
 using std::vector;
 
-FairSimConfig::FairSimConfig()
-    : fDescription("Options")
-    , fHelp(false)
-    , fnEvents(1)
-    , fEngine("TGeant3")
-    , fMultiThreaded(false)
-    , fOutputFile("sim.root")
-    , fParameterFile("par.root")
-    , fRandomSeed(0)
+static auto makeDescription()
 {
+    po::options_description description{"Options"};
     // clang-format off
-    fDescription.add_options()
+    description.add_options()
         ("help",                                        "Print this message")
         ("nevents",        po::value<int>(),            "Number of events to simulate")
         ("engine",         po::value<vector<string>>(), "Monte Carlo engine")
@@ -35,46 +31,47 @@ FairSimConfig::FairSimConfig()
         ("parameter-file", po::value<vector<string>>(), "Parameter file")
         ("random-seed",    po::value<int>(),            "Seed for the random number generator");
     // clang-format on
+    return description;
 }
-
-FairSimConfig::~FairSimConfig() {}
 
 int FairSimConfig::ParseCommandLine(int argc, char* argv[])
 {
+    po::variables_map map;
+    auto description = makeDescription();
     try {
-        po::store(po::parse_command_line(argc, argv, fDescription), fMap);
+        po::store(po::parse_command_line(argc, argv, description), map);
 
-        po::notify(fMap);
+        po::notify(map);
     } catch (po::error& e) {
         LOG(error) << e.what();
         fHelp = true;
         return 0;
     }
 
-    if (fMap.count("help")) {
+    if (map.count("help")) {
         fHelp = true;
         return 0;
     }
-    if (fMap.count("nevents")) {
-        fnEvents = fMap["nevents"].as<int>();
+    if (map.count("nevents")) {
+        fnEvents = map["nevents"].as<int>();
     }
-    if (fMap.count("engine")) {
-        fEngine = fMap["engine"].as<vector<string>>().at(0);
+    if (map.count("engine")) {
+        fEngine = map["engine"].as<vector<string>>().at(0);
         if (!GetEngine().EqualTo("TGeant3") && !GetEngine().EqualTo("TGeant4")) {
             LOG(error) << "Option engine can be either TGeant3 or TGeant4";
             return 1;
         }
     }
-    if (fMap.count("output-file")) {
-        fOutputFile = fMap["output-file"].as<vector<string>>().at(0);
+    if (map.count("output-file")) {
+        fOutputFile = map["output-file"].as<vector<string>>().at(0);
     }
-    if (fMap.count("parameter-file")) {
-        fParameterFile = fMap["parameter-file"].as<vector<string>>().at(0);
+    if (map.count("parameter-file")) {
+        fParameterFile = map["parameter-file"].as<vector<string>>().at(0);
     }
-    if (fMap.count("random-seed")) {
-        fRandomSeed = fMap["random-seed"].as<int>();
+    if (map.count("random-seed")) {
+        fRandomSeed = map["random-seed"].as<int>();
     }
-    if (fMap.count("multi-threaded")) {
+    if (map.count("multi-threaded")) {
         LOG(info) << "YUPYUPYUP";
         fMultiThreaded = true;
     }
@@ -83,5 +80,6 @@ int FairSimConfig::ParseCommandLine(int argc, char* argv[])
 
 void FairSimConfig::PrintHelpMessage()
 {
-    std::cout << fDescription << std::endl;
+    auto description = makeDescription();
+    std::cout << description << std::endl;
 }
--- a/examples/simulation/Tutorial1/src/FairSimConfig.h
+++ b/examples/simulation/Tutorial1/src/FairSimConfig.h
@@ -1,26 +1,22 @@
-/********************************************************************************
- *    Copyright (C) 2014 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH    *
- *                                                                              *
- *              This software is distributed under the terms of the             *
- *              GNU Lesser General Public Licence (LGPL) version 3,             *
- *                  copied verbatim in the file "LICENSE"                       *
- ********************************************************************************/
+/*********************************************************************************
+ *  Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ *                                                                               *
+ *              This software is distributed under the terms of the              *
+ *              GNU Lesser General Public Licence (LGPL) version 3,              *
+ *                  copied verbatim in the file "LICENSE"                        *
+ *********************************************************************************/
 #ifndef FAIRSIMCONFIG
 #define FAIRSIMCONFIG
-
-#include "boost/program_options.hpp"
 
 #include <Rtypes.h>
 #include <TString.h>
 #include <string>
 
-namespace po = boost::program_options;
-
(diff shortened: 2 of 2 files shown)


fuhlig1:
The problem showed up with the latest XCode version 15. According to the bug report reported to the ROOT team the problem already appeared with XCode 13.3. 

The fix should be ported to v18.6_patches and v18.8_patches.

dennisklein (commented):
Shouldn't it only be switched off with apple libc++?

fuhlig1:
It doesn't hurt to be switched off for any system and compiler. The data members are not needed by the interpreter so they can be hidden.

dennisklein:
Well, someone else needs to approve this. I find it aweful (the ifdef cling), why do we add such files to the LinkDef list at all? And why `FairSimConfig` (suggests to be an official FairRoot framework class) in examples/tutorial1?

ChristianTackeGSI (changes requested):
This might seem to work, but it's doing something very strange to the ABI.

As an absolute minimum, please change the `ClassDef(FairSimConfig, 1);` to have a 0 at the end: No marshalling code should ever be create for this class.

Other steps I would like to ask about:
* Could we remove the `ClassDef` completely? When using the header in the interpreter, it will still create on "on the fly" thing. But at least it doesn't feel like we want to say "This is a perfect ROOT class"

fuhlig1:
> Well, someone else needs to approve this. I find it aweful (the ifdef cling), why do we add such files to the LinkDef list at all? And why `FairSimConfig` (suggests to be an official FairRoot framework class) in examples/tutorial1?

I also don't like the fix. We need ROOT to generate the glue code such that the class can be used from the macro. When this class was added and why it was put to the examples I can't tell.

examples/simulation/Tutorial1/src/FairSimConfig.h
```diff
@@ -35,8 +37,10 @@ class FairSimConfig
     int GetRandomSeed() const { return fRandomSeed; }
 
   private:
+#if !defined(__CLING__) && !defined(__ROOTCLING__)
     po::options_description fDescription;   //!
     po::variables_map fMap;                 //!
+#endif
 
     bool fHelp;
```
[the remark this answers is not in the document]
fuhlig1: Right.

ChristianTackeGSI:
So, having read the root issue, it seems that this is a plain bug in apple's libraries.

> We generally don't support broken things
> (<https://github.com/root-project/root/issues/12762#issuecomment-1606919636>)

I think, we should follow that!

If you still want to fix this for macos, I see only one way forward: Hide the whole boost program options stuff from the interpreter without using preprocessor guards.

1. Either try to go for some forward declaration of `class FairSimConfig;` in the relevant macros / etc.

   This might not be trivial, because things like
   ```cpp
   FairSimConfig config;
   DoSomething(config);
   ```
   don't work with a declaration.

   One would probably need to go for something like
   ```cpp
   void DoSomething(FairSimConfig const* config = nullptr)
   {
       FairSimConfig default_config;
       if (!config) {
           config = &default_config;
       }
   }
   ```
   which isn't pleasant either.

2. Do a pimpl style thing
   ```cpp
   #include <memory>
   class FairExampleProgramOptions;
   class FairSimConfig
   {
     private:
       std::unique_ptr<FairExampleProgramOptions> po;
   };
   ```

   … and move all of `FairExampleProgramOptions` into the `.cxx`.


I really don't know how to go forward here.

ChristianTackeGSI:
And yes, we have a huge amount of names in the example directory that look like they're official FairRoot classes/etc. My personal favourite: `FairStack`. I think, I have even seen one project using `#include <FairStack.h>`.

We should really start renaming (and/or moving them into `fairroot::examples::`?) them!

dennisklein:
> I think, we should follow that!
> 
> If you still want to fix this for macos, I see only one way forward: Hide the whole boost program options stuff from the interpreter without using preprocessor guards.

I like this approach as well. Another idea how to hide it: To me, it looks like there is no reason currently to have boost in the header at all, the member variables `FairSimConfig::fDescription` and `FairSimConfig::fMap` can be moved to local variables inside `FairSimConfig::ParseCommandLine()`. Then boost will disappear from the header and the initialization of the description object can also happen locally in `ParseCommandLine()`.

ChristianTackeGSI:
Moving this into local variables sounds even better!

Just `FairSimConfig::PrintHelpMessage` uses `fDescription` as well.

```cpp
struct FairExampleTut1OptionDescription
{
    po::options_description fDescription;
    FairExampleTut1OptionDescription() { /* setup fDescription */ }
}

int FairSimConfig::ParseCommandLine(int argc, char* argv[])
{
    FairExampleTut1OptionDescription description;
    po::variables_map map;
    // … description.fDescription), map …
}

void FairSimConfig::PrintHelpMessage()
{
    FairExampleTut1OptionDescription description
    std::cout << description.fDescription << std::endl;
}
```

dennisklein:
One could cache the help message in a string. Just a bit wasteful, if it is never printed though.

ChristianTackeGSI:
> One could cache the help message in a string. Just a bit wasteful, if it is never printed though.

Classic cpu time vs. space optimization.
* Either cache the string and do not double initialize fDescription
* Or initialize fDescription two times (first for ParseCommandLine, second for PrintHelpMessage

I don't really care, a few kB vs. a few µs.

fuhlig1:
I used the second option and initialise the description two times.

examples/simulation/Tutorial1/src/FairSimConfig.cxx
```diff
@@ -35,53 +45,58 @@ FairSimConfig::FairSimConfig()
...
+      ("nevents",        po::value<int>(),            "Number of events to simulate")
+      ("engine",         po::value<vector<string>>(), "Monte Carlo engine")
+      ("multi-threaded",                              "Geant4 multi threaded")
+      ("output-file",    po::value<vector<string>>(), "Output file")
+      ("parameter-file", po::value<vector<string>>(), "Parameter file")
+      ("random-seed",    po::value<int>(),            "Seed for the random number generator");
+  // clang-format on
+  std::cout << description << std::endl; }
+
+ClassImp(FairSimConfig);
```
ChristianTackeGSI, line 102: Where is this ClassImp coming from?!

examples/simulation/Tutorial1/src/FairSimConfig.cxx
```diff
@@ -1,32 +1,42 @@
...
+ *                                                                               *
+ *              This software is distributed under the terms of the              *
+ *              GNU Lesser General Public Licence (LGPL) version 3,              *
+ *                  copied verbatim in the file "LICENSE"                        *
+ *********************************************************************************/
 #include "FairSimConfig.h"
 
 #include "FairLogger.h"
 
+#include "boost/program_options.hpp"
```
ChristianTackeGSI, line 12:
```suggestion
#include <boost/program_options.hpp>
```

examples/simulation/Tutorial1/src/FairSimConfig.cxx
```diff
@@ -35,53 +45,58 @@ FairSimConfig::FairSimConfig()
...
+  // clang-format off
+  description.add_options()
+      ("help",                                        "Print this message")
+      ("nevents",        po::value<int>(),            "Number of events to simulate")
+      ("engine",         po::value<vector<string>>(), "Monte Carlo engine")
+      ("multi-threaded",                              "Geant4 multi threaded")
+      ("output-file",    po::value<vector<string>>(), "Output file")
+      ("parameter-file", po::value<vector<string>>(), "Parameter file")
+      ("random-seed",    po::value<int>(),            "Seed for the random number generator");
+  // clang-format on
```
ChristianTackeGSI, line 99: DRY (Do-not-Repeat-Yourself)!

Keeping those two in sync seems simple. But someone will forget about it at some point.

That's why I suggested the small struct to put into the .cxx.

dennisklein, line 99: But let's not duplicate this code please.

```suggestion
  auto description = makeDescription();
```

```cpp
auto makeDescription() {
  po::options_description description;
  // clang-format off
  description.add_options()
      ("help",                                        "Print this message")
      ("nevents",        po::value<int>(),            "Number of events to simulate")
      ("engine",         po::value<vector<string>>(), "Monte Carlo engine")
      ("multi-threaded",                              "Geant4 multi threaded")
      ("output-file",    po::value<vector<string>>(), "Output file")
      ("parameter-file", po::value<vector<string>>(), "Parameter file")
      ("random-seed",    po::value<int>(),            "Seed for the random number generator");
  // clang-format on
  return description;
}
```

dennisklein (changes requested):

examples/simulation/Tutorial1/src/FairSimConfig.cxx
fuhlig1: Stupid copy and paste. I did the changes in v18.8_patches and copied them over. Fixed.

examples/simulation/Tutorial1/src/FairSimConfig.cxx
fuhlig1: Fixed.

examples/simulation/Tutorial1/src/FairSimConfig.cxx
fuhlig1: Fixed.

fuhlig1: Thanks. That was the solution I was looking for yesterday but somehow I couldn't find it. Fixed.

examples/simulation/Tutorial1/src/FairSimConfig.cxx
```diff
@@ -1,32 +1,27 @@
...
-    , fHelp(false)
-    , fnEvents(1)
-    , fEngine("TGeant3")
-    , fMultiThreaded(false)
-    , fOutputFile("sim.root")
-    , fParameterFile("par.root")
-    , fRandomSeed(0)
-{
+auto makeDescription() {
+    po::options_description description;
```
ChristianTackeGSI, line 22: 1. Please make it static, so that it does not pollute the global namespace
2. The original ctor of fDescription was called with `"Options"`. Do we need this?

```suggestion
static auto makeDescription() {
    po::options_description description{"Options"};
```

ChristianTackeGSI (changes requested):
* Please take a look at the clang tidy warnings. Especially https://github.com/FairRootGroup/FairRoot/pull/1478/files#diff-f10cf89d4a4c599b74c3c056e1a81302bcca2dc3559cafda62e8737aa69388c7L19
* Please take a look at the formatter

examples/simulation/Tutorial1/src/FairSimConfig.cxx
dennisklein: FYI: The `"Options"` argument gives the help output a title line.

fuhlig1: Fixed

fuhlig1:
> Those two are still open:
> 
> > * Please take a look at the clang tidy warnings. Especially https://github.com/FairRootGroup/FairRoot/pull/1478/files#diff-f10cf89d4a4c599b74c3c056e1a81302bcca2dc3559cafda62e8737aa69388c7L19
> > * Please take a look at the formatter

Where do I find the results of the formatter.  Since the CMake version is not matching no format check is done.

ChristianTackeGSI:
> Where do I find the results of the formatter. Since the CMake version is not matching no format check is done.

1. Check the "Some checks were not successful" section at the end of just this PR
2. https://alfa-ci.gsi.de/blue/organizations/jenkins/FairRootGroup%2FFairRoot/detail/PR-1478/5/pipeline/24

fuhlig1:
> > Where do I find the results of the formatter. Since the CMake version is not matching no format check is done.
> 
> 1. Check the "Some checks were not successful" section at the end of just this PR
> 2. https://alfa-ci.gsi.de/blue/organizations/jenkins/FairRootGroup%2FFairRoot/detail/PR-1478/5/pipeline/24

That I did and I clicked on the _Details_ link at the end of the check/format line expecting that this will bring me to the results of this test. Obviously this only brings me to the overview page where I have to chose the correct test.

ChristianTackeGSI (approved):
All my concerns have been addressed. (I like moving to default member variable initializers!).

Someone else might want to take a look?

dennisklein (approved):

ChristianTackeGSI:
There has been plenty of time for new comments. So I merged it now.

ChristianTackeGSI:
Backported to 18.8: #1483
Backported to 18.6: 7a0391683b2f187db9848aa9ef3fee875a47f6aa (without PR, hope that was okay)

PR #1479, opened 2023-12-22
chore: Mark some functions static, Remove some useless casts

ChristianTackeGSI:
- chore: Mark some free standing functions static

  If free standing functions are only ever used in one compile unit, they should be marked `static` to not pollute the namespace during linking.

- chore: Remove useless casts

  Casting to the exact same type is quite useless.

---

Checklist:

### DIFF
--- a/fairroot/base/sim/FairVolumeList.cxx
+++ b/fairroot/base/sim/FairVolumeList.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -58,7 +58,7 @@ FairVolume* FairVolumeList::findObject(TString name)
         obj = static_cast<FairVolume*>(fData->At(i));
         if (obj) {
             if (obj->GetName() == name) {
-                return static_cast<FairVolume*>(obj);
+                return obj;
             }
         }
     }
@@ -68,7 +68,7 @@ FairVolume* FairVolumeList::findObject(TString name)
 
 void FairVolumeList::addVolume(FairVolume* elem)
 {
-    FairVolume* v = static_cast<FairVolume*>(findObject(elem->GetName()));
+    FairVolume* v = findObject(elem->GetName());
 
     if (v) {
         LOG(error) << "FairVolumeList element: " << elem->GetName() << " VolId : " << elem->getVolumeId()
--- a/fairroot/base/source/FairFileSource.cxx
+++ b/fairroot/base/source/FairFileSource.cxx
@@ -388,7 +388,7 @@ void FairFileSource::AddFriendsToChain()
         }
 
         TChain* chain = fFriendTypeList[inputLevel];
-        chain->AddFile(fileName, 1234567890, FairRootManager::GetTreeName());
+        chain->AddFile(fileName, TTree::kMaxEntries, FairRootManager::GetTreeName());
     }
 
     // Check if all friend chains have the same runids and the same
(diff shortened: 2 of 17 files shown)


fairroot/base/sim/FairVolumeList.cxx
```diff
@@ -58,7 +58,7 @@ FairVolume* FairVolumeList::findObject(TString name)
         obj = static_cast<FairVolume*>(fData->At(i));
         if (obj) {
             if (obj->GetName() == name) {
-                return static_cast<FairVolume*>(obj);
+                return obj;
             }
         }
     }
```
dennisklein, line 64: How about?

```cpp
FairVolume* FairVolumeList::findObject(TString name)
{
    return fData->FindObject(name.Data());
}
```

see https://root.cern/doc/v630/classTObjArray.html#a9bae82c7ff2a18715a90602e2aedca61

ChristianTackeGSI: Yeah, I was wondering, why this was not used like this. I was rather thinking of replacing TObjArray with a `std::vector<std::unique_ptr<FairVolume>>`. And then this linear search would be needed again.

Or maybe we should create something like this?
```cpp
namespace fairroot::detail
{
template<typename T>
T* FindObject(std::vector<std::unique_ptr<T>>& v, std::string_view v)
{
    auto name_matches = [&](std::unique_ptr<TObject>& p) { return p->GetName() == name; };
    auto it = std::find_if(std::begin(v), std::end(v), name_matches);
    if (it == std::end(v))
        return nullptr;
    return it->get();
}
}
```

dennisklein: Hm, why even hardcode `vector<unique_ptr<T>>`?

dennisklein: And `FindObject` would be a terrible name. More like `FindByName`?

dennisklein: But thinking about what `FairVolumeList` wants to be, isn't it not just a set?

dennisklein: Looking at where `FairVolumeList::findObject()` is called, I find [`FairModule::AddSensitiveVolume()`](https://github.com/FairRootGroup/FairRoot/blame/00249b0fc465bace608c4a3c5a9291e015c86895/base/sim/FairModule.cxx#L295-L311) which searches the "list" twice (linearly) since [12 years](https://github.com/FairRootGroup/FairRoot/commit/386912ebcc7877927980cbbf7ef46a32dddc3e4b).

We should definitely rename this class to reflect it is a "Set" ;)

dennisklein: https://github.com/FairRootGroup/FairRoot/issues/1495

ChristianTackeGSI: > Hm, why even hardcode `vector<unique_ptr<T>>`?

Because of the `it->get()` at the end.

ChristianTackeGSI: > #1495

Now that we have a dedicated bug to look at all of this: What about we merge this PR just "as is"?

fairroot/base/source/FairFileSource.cxx
```diff
@@ -387,7 +387,7 @@ void FairFileSource::AddFriendsToChain()
             friendType++;
         }
 
-        TChain* chain = static_cast<TChain*>(fFriendTypeList[inputLevel]);
+        TChain* chain = fFriendTypeList[inputLevel];
```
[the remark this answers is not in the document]
ChristianTackeGSI: Hey, I think, I have fixed this in f9f06e386aa1accb0cb46d35aa758f81e4dcdf25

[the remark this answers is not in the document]
ChristianTackeGSI: Check your script: The correct path is `fairroot/base/source/FairFileSource.cxx` and not `base/source/FairFileSource.cxx`.

dennisklein (approved):

PR #1486, opened 2024-02-09
Update Mac CI nodes

rbx:
Will likely still fail atm, but let's see what the problems are.


rbx:
@ChristianTackeGSI thx! I was about to do that :D

ChristianTackeGSI:
One of the build errors was related to CMake 3.28 and [CMP0153](https://cmake.org/cmake/help/latest/policy/CMP0153.html).

So trying to replace `exec_program` by `execute_process`.

I added a commit that tries to fix this. Please verify carefully!

ChristianTackeGSI:
Okay, this gets "macos-14-arm64-apple-clang-15-fairsoft-24.01" a bit further and now has problems in the testsuite. (And I don't know, whether I understand those.

Should I put my "build: Replace exec_program by execute_process" commit in its own PR, so that we can merge it already?!

rbx:
FairSoft brew package is currently not installed on the test machines. But we should still be able to see the effect of the `execute_process` change.

ChristianTackeGSI:
Yes, it's visible on [CDash](https://cdash.gsi.de/viewConfigure.php?buildid=439331). First Configure now succeeds. Second search for `-- Found a Mac OS X System 14.3`.

That's why I ask, whether we should already merge this (using a distinct PR).

rbx:
Sure, let's merge it already and I'll rebase this in a bit.

ChristianTackeGSI:
done in #1487

rbx:
All three MacOS builds produce a bunch of warnings of this type:
```
In file included from /opt/homebrew/opt/fairsoft/include/root/Rtypes.h:23:
In file included from /opt/homebrew/opt/fairsoft/include/root/RtypesCore.h:23:
In file included from /opt/homebrew/opt/fairsoft/include/root/ROOT/RConfig.hxx:23:
[CTest: warning matched] /opt/homebrew/opt/fairsoft/include/root/RConfigure.h:30:4: warning: "The C++ standard in this build does not match ROOT configuration (L); this might cause unexpected issues" [-W#warnings]
[CTest: warning matched] #  warning "The C++ standard in this build does not match ROOT configuration (L); this might cause unexpected issues"
```
First idea was that this is due to lack of a default CXX standard setting in the legacy [cmake configuration of fairsoft](https://github.com/FairRootGroup/FairSoft/blob/master/cmake/legacy.cmake).

But even if I pass a standard (e.g. 17) explicitly out of the homebrew recipe, the warning remained.
Upon further inspection, `/opt/homebrew/opt/fairsoft/include/root/RConfigure.h` contains:
```
#define ROOT__cplusplus L
```
so at some point [this](https://github.com/root-project/root/blob/master/config/RConfigure.in#L25) turns into `L`...

rbx:
This must be failing: https://github.com/root-project/root/blob/v6-30-02/cmake/modules/RootConfiguration.cmake#L659-L664 which ends up with just an "L". 

This happens only in the brew compilation, not if I compile fairsoft manually.

ChristianTackeGSI:
So `CMAKE_MATCH_1` is just empty. So the regex does not match. Is the homebrew compile wrapper (they use one, right?) doing funny things like redirecting stderr?

```console
alfaci@citmac001 /tmp % clang++  -c cpp.cxx                                    
cpp.cxx:5:9: warning: __cplusplus=199711L [-W#pragma-messages]
#pragma message("__cplusplus=" STRINGIFY(__cplusplus))
        ^
1 warning generated.
alfaci@citmac001 /tmp % clang++ -c cpp.cxx 2>/dev/null
```

All that said: Is that the real problem that we're facing or just annoying? (It might be a real problem, if other `try_compile` have the same problem?)

rbx:
> All that said: Is that the real problem that we're facing or just annoying?

So far I didn't see a direct connection to other failures from this.

In any case, I think we can merge this now so that mac run other CI jobs. And I'll follow up on other issues in some test PR or from the other PRs.

ChristianTackeGSI:
> In any case, I think we can merge this now so that mac run other CI jobs. And I'll follow up on other issues in some test PR or from the other PRs.

Generally I am okay with this. This goes well with "best effort support for macos".

I have a small issue: This adds another three failing checks to the list. So we then have 7 out of 21 that just always fail (and checking the details is annoying, to be honest).
I could suggest to only perform the build step on macos and skip the test step (because this fails). But I don't know, if that's helpful either way?

rbx:
> I could suggest to only perform the build step on macos and skip the test step (because this fails).

I would rather look into all the issue and fix them. We can also leave this open until I find time to look into all of them.

rbx:
Most of the issues on newer Mac fail in this way:
```
<<< cling interactive line includer >>>: fatal error: module file '/tmp/fairsoft-20240212-8337-c3i4n4/build/Build/root/lib/Vc.pcm' not found: module file not found
<<< cling interactive line includer >>>: note: imported by module 'MathCore' in '/usr/local/Cellar/fairsoft/jan24/lib/MathCore.pcm'
input_line_4:2:2: error: unknown type name 'include'
#include "TError.h"
 ^
input_line_4:2:1: error: expected unqualified-id
#include "TError.h"
```
So for some reason it looks for a file in the build directly and then the following I don't understand if it is related to that or not.

rbx:
While on the older MacOS they are mostly due to not found Geant4:
```
/Users/alfaci/alfaci/workspace/FairRootGroup_FairRoot_PR-1486/examples/common/gconfig/g4Config.C:44:5: error: unknown type name 'TGeant4'
    TGeant4* geant4 = new TGeant4("TGeant4", "The Geant4 Monte Carlo", runConfiguration);
```

ChristianTackeGSI:
> until I find time to look into all of them

Yeah, that's probably the point. If we find the time soonish, then it would be good to merge it now. So that we have the incentive to fix it for real. If we realize, that it will take much longer. It will just increase our "fails always level" to 7 out of 21.

Really your dicision. I am okay with either way.

ChristianTackeGSI (approved):

PR #1496, opened 2024-03-04
Fix #1495

### LINKED ISSUE #1495: `FairModule::AddSensitiveVolume` searches volume set twice
https://github.com/FairRootGroup/FairRoot/commit/386912ebcc7877927980cbbf7ef46a32dddc3e4b introduced a double linear search of the volume on `vList` in [`FairModule::AddSensitiveVolume`](https://github.com/FairRootGroup/FairRoot/blame/00249b0fc465bace608c4a3c5a9291e015c86895/base/sim/FairModule.cxx#L295-L311):

1. search happens [here](https://github.com/FairRootGroup/FairRoot/commit/386912ebcc7877927980cbbf7ef46a32dddc3e4b#diff-81dfd931f93b7bdaaa11c1651c35692da18a0e7ff3513022304b7f12bdb6d06eR196)
2. search happens [here](https://github.com/FairRootGroup/FairRoot/commit/386912ebcc7877927980cbbf7ef46a32dddc3e4b#diff-81dfd931f93b7bdaaa11c1651c35692da18a0e7ff3513022304b7f12bdb6d06eR199), see [`FairVolumeList::addVolume`](https://github.com/FairRootGroup/FairRoot/commit/386912ebcc7877927980cbbf7ef46a32dddc3e4b#diff-81dfd931f93b7bdaaa11c1651c35692da18a0e7ff3513022304b7f12bdb6d06eR199)

Possible fix involves changing the error behaviour of `FairVolumeList::addVolume` to return the failure to add to the caller, e.g. by returning a pointer to the added volume (which is `nullptr` in case of not adding it).

dennisklein:
~It is not finished, especially there is an open point still where I added the "TODO" comment. But I wanted to start the discusssion if this is the right approach.~

## Summary by CodeRabbit

- **New Features**
  - Introduced functionality for managing a list of FairVolumes, including adding, retrieving, and handling volume entries.

- **Bug Fixes**
  - Improved memory management and logging for FairVolume instances.

- **Refactor**
  - Updated methods to use modern C++ practices for better performance and safety.
  
- **Tests**
  - Added new tests for FairVolumeList functionality to ensure reliability and correctness.


dennisklein:
@coderabbitai read fairroot/base/sim/FairVolumeList[.h|.cxx] and generate additional unit tests for the file tests/base/sim/test_FairVolumeList.cxx.

dennisklein:
@coderabbitai read the git commits in this PR and suggest improvements on the commit msg text for each one. Make sure to state the reasons for your change suggestions.

dennisklein:
@coderabbitai use origin/dev as base branch

dennisklein:
@coderabbitai now do the same also for each commit msg body if present.

dennisklein:
@coderabbitai read the file CHANGELOG.md and generate additional changelog entries for the important user-facing changes in this PR.

dennisklein:
@coderabbitai give it another shot, try to keep each point slightly shorter and focus on important user-facing API changes only

fairroot/base/sim/FairVolumeList.h
```diff
@@ -25,24 +27,39 @@
...
-    ClassDefOverride(FairVolumeList, 1);
+    FairVolume* findObject(const TString& name) { return static_cast<FairVolume*>(fData.FindObject(name.Data())); }
+
+    FairVolume* addVolume(std::unique_ptr<FairVolume> vol);
+    [[deprecated]] FairVolume* addVolume(FairVolume* elem) { return addVolume(std::unique_ptr<FairVolume>(elem)); }
+
+    Int_t getEntries() { return fData.GetEntries(); }
+    FairVolume* At(Int_t pos) { return (dynamic_cast<FairVolume*>(fData.At(pos))); }
+
+    ClassDefOverride(FairVolumeList, 2);
```
ChristianTackeGSI, line 62: I think, all current uses of `FairVolumeList` do not expect it to be streamed. So maybe let's just remove the streaming?

```suggestion
    ClassDefOverride(FairVolumeList, 0);
```

dennisklein: fine with me, will adopt your suggestion in the next edit round

dennisklein:
@coderabbitai Go through the git commits in this PR (origin/dev is base branch) and evaluate the commit granularity. Make suggestions which commits to split or merge based on whether they contain related or unrelated changes. The goal is a rather fine grained history in compliance with the conventional commits spec v1.0.0 (www.conventionalcommits.org).

dennisklein:
@coderabbitai read the file CONTRIBUTING.md (dev branch) and all the weblinks provided in there. Evaluate whether the coding guideline rules provided there have been violated by this PR (base branch origin/dev). If you find violations, give a short summary and generate a markdown link referencing the violated guideline as specified in the CONTRIBUTING.md file.

dennisklein:
> Without direct access to the PR's code changes, it's not possible to definitively identify violations of the coding guidelines.

@coderabbitai use this PR's code changes (base branch origin/dev) in your evaluation to identify violations of the coding guidelines.

fairroot/base/sim/FairVolumeList.h
```diff
@@ -30,7 +30,7 @@ class FairVolumeList : public TObject
     TObjArray fData;
 
   public:
-    FairVolumeList();
+    FairVolumeList() { fData.SetOwner(kTRUE); }
```
ChristianTackeGSI: Consider squashing this part into "feat(Base): Avoid extra allocation in `FairVolumeList`"?

ChristianTackeGSI (commented):
Please consider some of this:

- Turn `vList` into a value member: currently at thread exit the FairVolumeList is not destructed
- Turn `vList` into a std container (maybe a `std::map<std::string, std::unique_ptr<FairVolume>>` to avoid complex compare/hash functions; or maybe `std::set`) and deprecate FairVolumeList

In the long run, we might consider to move those thread locals to member variables of FairRunSim (which handles the "simulation session"). See for example #1497.

dennisklein:
> Turn vList into a std container (maybe a `std::map<std::string, std::unique_ptr<FairVolume>>` to avoid complex compare/hash functions; or maybe `std::set`) and deprecate FairVolumeList

Perhaps not even using a full `FairVolume`?! We only store an int in addition to the name, so a `std::map<std::string, int>` would be enough and saves another heap object indirection. And regarding the int I am confused that it is stored in the vList placeholder and not in the original volume? (Finally needs to be looked at in combination with svList which I have not had the chance to look at in detail)

fairroot/base/sim/FairVolumeList.h
dennisklein: Ya, good idea, will do on the next edit round

ChristianTackeGSI:
Maybe we can first simplify (maybe by using std containers) the current mechanics and then it is easier to assess whether we really need FairVolume?

dennisklein:
@coderabbitai summarize the code changes and conversation thread in this PR (base branch origin/dev) in the style of a nerve-racking, thrilling and action-loaded short novel of about 500 words. anonymize the names of the thread participants by replacing them with character names from the fictional star trek universe.

fairroot/base/sim/FairModule.cxx
```diff
@@ -91,6 +91,7 @@ FairModule::FairModule(const FairModule& rhs)
         }
     }
 
+    // TODO: This is broken. FairVolumeList owns contained volumes!
```
ChristianTackeGSI, line 94: This isn't any longer an issue (see #1534).

fairroot/base/sim/FairVolumeList.h
dennisklein: done

fairroot/base/sim/FairVolumeList.h
dennisklein: done

dennisklein:
> Please consider some of this:
> 
>     * Turn `vList` into a value member: currently at thread exit the FairVolumeList is not destructed
> 
>     * Turn `vList` into a std container (maybe a `std::map<std::string, std::unique_ptr<FairVolume>>` to avoid complex compare/hash functions; or maybe `std::set`) and deprecate FairVolumeList

To get this done, I'll consider this out of scope of this PR.

tests/base/sim/test_FairVolumeList.cxx
```diff
@@ -0,0 +1,49 @@
+/********************************************************************************
+ *   Copyright (C) 2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH     *
+ *                                                                              *
+ *              This software is distributed under the terms of the             *
+ *              GNU Lesser General Public Licence (LGPL) version 3,             *
+ *                  copied verbatim in the file "LICENSE"                       *
+ ********************************************************************************/
+
+#include <FairVolume.h>
+#include <FairVolumeList.h>
```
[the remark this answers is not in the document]
dennisklein: Technically, `""` means just to prepend the local directory of the current file to the search path. This effect is clearly not used.

Repurposing the meaning of `""` to *project-local* is slightly dangerous, because it actually has a different meaning, but usually is fine.

In this case, I am using FairRoot headers from the test suite, which is technically the same as any other user. So, in my thinking, this is one place where I would like to see how the FairRoot API looks like from a user's perspective.

Furthermore, I would also like to apply this to the FairRoot library code itself, e.g. if a GeoBase file includes a Base header, it should use `<>`, but if it includes a GeoBase header it should use `""`. In other words, I would prefer a meaning of *target-local* or *library-local* instead of *project-local*.

What is the benefit of *project-local*?

In any case, I would love to see this defined as part of our coding guidelines.

fairroot/base/sim/FairModule.cxx
```diff
@@ -246,18 +246,19 @@ void FairModule::ProcessNodes(TList* aList)
...
-        volume->setModId(fModId);
-        volume->SetModule(this);
-        fAllSensitiveVolumes.push_back(volume);
-        fNbOfSensitiveVol++;
+    if (auto vol_added = vList->addVolume(std::move(volume)); vol_added) {
+        ++fNbOfVolumes;
+        vol_added->setModId(fModId);
+        vol_added->SetModule(this);
+        fAllSensitiveVolumes.push_back(vol_added);
+        ++fNbOfSensitiveVol;
```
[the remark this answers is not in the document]
dennisklein: I like it

fairroot/base/sim/FairModule.cxx
```diff
@@ -213,30 +213,30 @@ void FairModule::ProcessNodes(TList* aList)
...
-        volume = new FairVolume(node->getTruncName(), fNbOfVolumes++);
+        auto volume = std::make_unique<FairVolume>(node->getTruncName(), fNbOfVolumes++);
+        auto volume_ptr = volume.get();
         volume->setRealName(node->GetName());
-        vList->addVolume(volume);
-        volume->setGeoNode(node);
-        volume->setCopyNo(node->getCopyNo());
+        vList->addVolume(std::move(volume));
+        volume_ptr->setGeoNode(node);
+        volume_ptr->setCopyNo(node->getCopyNo());
```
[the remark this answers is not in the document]
dennisklein: I think this is semantically fine, if one thinks of `FairVolumeList::addVolume()` to be (defined as) a `std::set<std::unique_ptr<FairVolume>>::push_back()`. However, I agree it is not very readable and points out how to improve a later commit in this PR: I added a `[[nodiscard]]` to the changed `FairVolume* FairVolumeList::addVolume(std::unique_ptr<FairVolume>)` which warns here and then leads to the code being improved in the later commit.

fairroot/base/sim/FairVolumeList.h
```diff
@@ -25,24 +27,39 @@
...
-    FairVolume* At(Int_t pos) { return (dynamic_cast<FairVolume*>(fData->At(pos))); }
+    Int_t getVolumeId(const TString& name);
+    [[deprecated]] Int_t getVolumeId(TString* name)
+    {
+        assert(name);
+        return getVolumeId(*name);
+    }
 
-    ClassDefOverride(FairVolumeList, 1);
+    FairVolume* findObject(const TString& name) { return static_cast<FairVolume*>(fData.FindObject(name.Data())); }
```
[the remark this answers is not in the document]
dennisklein: `fData` can only contain `FairVolume`s, so we can skip handling this case here.

tests/base/sim/test_FairVolumeList.cxx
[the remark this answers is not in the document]
ChristianTackeGSI: > In any case, I would love to see this defined as part of our coding guidelines (whatever definition we agree on)

:+1: 

Let's move this discussion somewhere else. I am okay with whatever you like for this PR. :-)

Resolving.

fairroot/base/sim/FairModule.cxx
```diff
@@ -210,55 +218,50 @@ void FairModule::ProcessNodes(TList* aList)
...
-        volume->setGeoNode(node);
-        volume->setCopyNo(node->getCopyNo());
 
-        if (MotherNode != 0) {
-            volume->setMotherId(node->getMCid());
-            volume->setMotherCopyNo(MotherNode->getCopyNo());
+        auto addedVol = vList->addVolume(std::move(volume));
+        if (!addedVol) {
+            continue;
+        }
```
ChristianTackeGSI, line 236: I know I proposed this code! So my fault for mentioning this now…

Could we have a `LOG(warn)` here? AFAICT this code runs only during "initialization" and the user might want to know if some volumes just get dropped due to "duplicate checks".

fairroot/base/sim/FairModule.cxx
```diff
@@ -210,55 +218,50 @@ void FairModule::ProcessNodes(TList* aList)
...
-    FairGeoNode* MotherNode = nullptr;
-    FairVolume* volume = nullptr;
-    FairRuntimeDb* rtdb = FairRun::Instance()->GetRuntimeDb();
-    FairGeoParSet* par = static_cast<FairGeoParSet*>(rtdb->getContainer("FairGeoParSet"));
-    TObjArray* fNodes = par->GetGeoNodes();
-    while ((node = static_cast<FairGeoNode*>(iter.Next()))) {
-
+    auto rtdb = FairRun::Instance()->GetRuntimeDb();
+    auto par = static_cast<FairGeoParSet*>(rtdb->getContainer("FairGeoParSet"));
+    auto parNodes = par->GetGeoNodes();
```
ChristianTackeGSI, line 223: I do like `auto` a lot.

We had a private talk about giving the `auto` a "concept" attribute to say "Okay, I don't care about the actual type, but I expect some properties from that type". In the ROOT world this would probably boil down to the base type that has all the relevant properties. I think, in this case it's [`TSeqCollection`](https://root.cern/doc/v630/classTSeqCollection.html#a38b7fc74c2740f1f02326c9c65d180d3)
```suggestion
    TSeqCollection* parNodes = par->GetGeoNodes();
```

What do you think?

fairroot/base/sim/FairModule.h
```diff
@@ -147,6 +147,7 @@ class FairModule : public TNamed
      * \deprecated Not used anywhere, will be removed in a future release
      */
     [[deprecated]] FairVolume* getFairVolume(FairGeoNode* fNode);
+    /** Takes ownership of passed volume **/
     void AddSensitiveVolume(TGeoVolume* v);
```
ChristianTackeGSI, line 151: Are you actually sure this takes the ownership? Reading your much refactored version lets me doubt this. I could even image, that you could turn this into a `const TGeoVolume*`?

ChristianTackeGSI (commented):
Only the ownership thing needs a real look. Everything else is "cosmetic", really.

fairroot/base/sim/FairModule.cxx
dennisklein: I agree, that's even better. Will change it.

fairroot/base/sim/FairModule.h
dennisklein: You are absolutely right. It creates a new object that gets owned by the `vList`. Will remove the comment.

ChristianTackeGSI (approved):

ChristianTackeGSI (dismissed):
Let's wait for the CI

ChristianTackeGSI (approved):

PR #1499, opened 2024-03-08
Use std containers for FairModule::svList

ChristianTackeGSI:
`FairModule::svList` can very well be implemented using `std::vector` without all the casting.

And rename to `fAllSensitiveVolumes`.

Follow up from #1497

---

Checklist:

### DIFF
--- a/fairroot/base/sim/FairModule.cxx
+++ b/fairroot/base/sim/FairModule.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -47,6 +47,8 @@
 #include <map>
 #include <memory>
 
+thread_local std::vector<FairVolume*> FairModule::fAllSensitiveVolumes;
+
 void FairModule::ConstructGeometry()
 {
     LOG(warn)
@@ -65,9 +67,6 @@ FairModule::FairModule(const char* Name, const char* title, Bool_t Active)
     : TNamed(Name, title)
     , fActive(Active)
 {
-    if (!svList) {
-        svList = new TRefArray();
-    }
     if (!vList) {
         vList = new FairVolumeList();
     }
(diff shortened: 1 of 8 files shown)


ChristianTackeGSI:
@coderabbitai I think, you got some additional prompts like looking at CONTRIBUTING.md, etc. Which did you get? Have you worked on them?

fairroot/base/sim/FairModule.cxx
```diff
@@ -251,7 +241,7 @@ void FairModule::ProcessNodes(TList* aList)
         if (node->isSensitive() && fActive) {
             volume->setModId(fModId);
             volume->SetModule(this);
-            svList->Add(volume);
+            svList.push_back(volume);
```
dennisklein, line 244: Follow-up from https://github.com/FairRootGroup/FairRoot/pull/1497#discussion_r1514827979

> So caching `fAllSensitiveVolumes` in a local variable is somewhat a preparation for (3). There it would look like
> 
> ```c++
> auto& svList = FairRunSim::Instance()->fAllSensitiveVolumes;
> ```

But why keep the "acceptable" ([S.2](https://github.com/FairRootGroup/FairRoot/blob/dev/CONTRIBUTING.md#s2-prefer-variable-names-describing-what-they-contain)) variable name ;) Better name would be `sensitiveVolumes`. Or, in this case, just

```suggestion
            fAllSensitiveVolumes.push_back(volume);
```

and later

```suggestion
            FairRunSim::Instance()->fAllSensitiveVolumes.push_back(volume);
```

ChristianTackeGSI: I would like to avoid `FairRunSim::Instance()` inside a loop. But if it's not in a loop (one case is just inside an `if`, I think), you're right.

You're right, the local variable (if I need to touch all references anyway, should be named like you said `allS|s]ensitiveVolumes`.

I'll fix both things.

dennisklein (commented):
I believe FairRoot loosely follows ROOT's naming conventions (with many violations all over the place), for sure the `f`-member-variable prefix is taken from ROOT:
* https://root.cern/contribute/coding_conventions/#naming-conventions
* -> https://root.cern/TaligentDocs/TaligentOnline/DocumentRoot/1.0/Docs/books/WM/WM_63.html#HEADING77

So, if we say we still adhere to it, we should at least discuss why not `fgAllSensitiveVolumes` etc

fairroot/base/sim/FairModule.cxx
ChristianTackeGSI: Okay, force pushed a new version. I switched them all over. No caching local variables any more.

I also updated the preview of moving this to a normal member variable of FairRunSim: Top commit of https://github.com/ChristianTackeGSI/FairRoot/tree/pr/svlist_v2

ChristianTackeGSI:
> So, if we say we still adhere to it, we should at least discuss (quickly) why not `fgAllSensitiveVolumes` etc

Yes, we should!

Just: We should even more try to get rid of them… That's what I was trying with the preview of moving this to FairRunSim.

dennisklein (approved):

dennisklein:
@coderabbitai generate a short executive summary on the code changes in this PR (base branch origin/dev) as well as its conversation thread if any important decisions were made.

PR #1504, opened 2024-03-26
Replace deprecated APIs from FairMQ and Geant4

karabowi:
Addresses issue #1501 in FairMQ part.
Also Geant4 deprecates functions like Set...() in favor of Propose...().

---

Checklist:

### DIFF
--- a/examples/MQ/histogramServer/FairMQExHistoDevice.cxx
+++ b/examples/MQ/histogramServer/FairMQExHistoDevice.cxx
@@ -55,7 +55,7 @@ bool FairMQExHistoDevice::ConditionalRun()
     auto message(NewMessage());
     RootSerializer().Serialize(*message, &fArrayHisto);
 
-    for (auto& channel : fChannels) {
+    for (auto& channel : GetChannels()) {
         Send(message, channel.first);
     }
 
--- a/examples/MQ/pixelDetector/src/devices/FairMQRunDevice.cxx
+++ b/examples/MQ/pixelDetector/src/devices/FairMQRunDevice.cxx
@@ -66,7 +66,7 @@ void FairMQRunDevice::SendBranches(FairOnlineSink& sink)
 
     TList* branchNameList = FairRootManager::Instance()->GetBranchNameList();
 
-    for (auto& mi : fChannels) {
+    for (auto& mi : GetChannels()) {
         LOG(debug) << "trying channel >" << mi.first << "<";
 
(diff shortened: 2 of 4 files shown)


examples/MQ/histogramServer/FairMQExHistoDevice.cxx
```diff
@@ -55,7 +55,7 @@ bool FairMQExHistoDevice::ConditionalRun()
     auto message(NewMessage());
     RootSerializer().Serialize(*message, &fArrayHisto);
 
-    for (auto& channel : fChannels) {
+    for (auto& channel : GetChannels()) {
         Send(message, channel.first);
```
dennisklein, line 59: Using the new API unconditionally means you require FairMQ v1.5.0. Is this intended? If yes, then we should express this in the `find_package` call.

rbx:
I attempted to test this on my machine, but I get the same compilation error as on cdash: https://cdash.gsi.de/viewBuildError.php?buildid=447412
With FairSoft jan24.

rbx:
As for the FairMQ-related changes, they look good to me and get rid of the warnings. Indeed, as Dennis said, `GetChannels` is only available since FairMQ v1.5.0, so that should be bumped.

dennisklein:
> As for the FairMQ-related changes, they look good to me and get rid of the warnings. Indeed, as Dennis said, `GetChannels` is only available since FairMQ v1.5.0, so that should be bumped.

Whether to bump the required FairMQ version should depend on the range of FairSoft versions intended to be supported!

karabowi:
You are all right. Since we want to have the v19 also working with nov22, we should not bump the FairMQ version, but rather deal with the deprecation warning other way. Closing this PR and waiting for Dennis to deal with issue https://github.com/FairRootGroup/FairRoot/issues/1501.

dennisklein:
Btw, the analog question is also open regarding the minimum required ROOT version. Since we switch to discover ROOT via its CMake package, we implicitely bump the min version to the first ROOT release that ships a CMake package. However, in our CMake we also state some version.

In general, the version combinatorics are poorly managed in FairRoot, imho we should improve here and clearly document which versions FairRoot intends to support (for all deps). Not just in terms of FairSoft, technically, FairRoot should not need to know about FairSoft, meaning we should resolve the required versions (which may be defined via FairSoft) to a per package term. At least, if we ever want to achieve regular packability of FairRoot.

PR #1505, opened 2024-03-26
chore(fastsim) Replace deprecated Geant4 APIs

karabowi:
Geant4 (geant4-11.2) deprecates functions G4FastStep::Set...(),
and advices usage of corresponding G4FastStep::Propose...().

Replaced them all in FairFastSimModel.cxx.

---

Checklist:

### DIFF
--- a/fairroot/fastsim/FairFastSimModel.cxx
+++ b/fairroot/fastsim/FairFastSimModel.cxx
@@ -126,15 +126,15 @@ void FairFastSimModel::DoIt(const G4FastTrack& fastTrack, G4FastStep& fastStep)
         G4double len = 0.;
         G4double ek = particle->Ek() * 1000.;
 
-        fastStep.SetPrimaryTrackFinalPosition(pos, false);
-        fastStep.SetPrimaryTrackFinalTime(tim);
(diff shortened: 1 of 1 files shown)


dennisklein (commented):
I would love to have the information that Geant4 11.2 introduced the deprecation also in the commit message.

ChristianTackeGSI:
> I would love to have the information that Geant4 11.2 introduced the deprecation also in the commit message.

Isn't that already in the commit message? I am probably missing something. What should be changed? Maybe suggest a new wording?

That said:

Do we know when `G4FastStep::Propose…()` were introduced? Maybe we should raise the minimum version in our `CMakeLists.txt`?

And finally: Can you please switch the target to the `dev` branch, so that we then backport / fast-forward to all appropriate release branches?

dennisklein:
> Isn't that already in the commit message? I am probably missing something.

Not in the first version before the force push.

karabowi:
Yes, the force was only corresponding change in the commit message.

ChristianTackeGSI:
Still missing:

> Do we know when `G4FastStep::Propose…()` were introduced? Maybe we should raise the minimum version in our `CMakeLists.txt`?

dennisklein (approved):

PR #1521, opened 2024-04-16
ci: Do not check Geant3 on macos

ChristianTackeGSI:
It looks like Geant3 fails always on macos.
So disable it even in branch builds on macos.

See: https://alfa-ci.gsi.de/blue/organizations/jenkins/FairRootGroup%2FFairRoot/detail/pr%2Fci/1/pipeline/101

---

Checklist:


dennisklein (changes requested):
Hm, isn't this the wrong abstraction to discriminate between branch and pr builds? If I run this test locally, those conditions have no meaning, no?

Issue #1519, opened 2024-04-15
`warning: 'virtual bool FairTimeStamp::operator<(const FairTimeStamp*) const' was hidden [-Woverloaded-virtual=]`


dennisklein:
This warning appears on recent compiler versions in multiple places:

```
/fairroot/base/event/FairTimeStamp.h:77:18: warning: ‘virtual bool FairTimeStamp::operator<(const FairTimeStamp*) const’ was hidden [-Woverloaded-virtual=]
   77 |     virtual bool operator<(const FairTimeStamp* rValue) const { return GetTimeStamp() < rValue->GetTimeStamp(); }
      |                  ^~~~~~~~
/examples/advanced/Tutorial3/data/FairTestDetectorDigi.h:68:18: note:   by ‘virtual bool FairTestDetectorDigi::operator<(const FairTestDetectorDigi&) const’
   68 |     virtual bool operator<(const FairTestDetectorDigi& myDigi) const
      |                  ^~~~~~~~
```

The issue was discussed in our group meeting on 15th April 2024 and assigned to @TobiasStockmanns for review/advice for now.

---

For reference: A [previous attempt](https://github.com/dennisklein/FairRoot/commit/9f334c0022440d75ad35794cd0ff794aa44a2ac5) to resolve the issue [was rejected/commented by @karabowi](https://github.com/FairRootGroup/FairRoot/pull/1516#pullrequestreview-1985671447)

PR #1526, opened 2024-04-22
Change operator< from FairTimeStamp* to FairTimeStamp&

TobiasStockmanns:
The behaviour of the FairTimeStamp::operator< does no follow standard behaviour by dealing with the FairTimeStamp* instead of a FairTimeStamp&. This is fixed in this merge request.
Should fix issue #1519.

Describe your proposal.

Mention any issue this PR resolves or is related to.

---

Checklist:

## Summary by CodeRabbit

- **Refactor**
	- Improved the comparison operation in the `FairTimeStamp` class for enhanced reliability and performance.
	- Updated the `CHANGELOG.md` file.

### DIFF
--- a/fairroot/base/event/FairTimeStamp.h
+++ b/fairroot/base/event/FairTimeStamp.h
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -74,7 +74,7 @@ class FairTimeStamp : public FairMultiLinkedData_Interface
         return out;
     }
 
-    virtual bool operator<(const FairTimeStamp* rValue) const { return GetTimeStamp() < rValue->GetTimeStamp(); }
+    bool operator<(const FairTimeStamp& rValue) const { return fTimeStamp < rValue.fTimeStamp; }
 
   protected:
     Double_t fTimeStamp{-1};        //< Time of digit or Hit  [ns]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,7 +73,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
     * `basemq/devices/FairMQSampler.h` -> `examples/advanced/Tutorial3/MQ/sampler.cxx`
     * `basemq/tasks/FairMQProcessorTask.h` -> `examples/advanced/Tutorial3/MQ/processorTask/ProcessorTask.h`
     * `basemq/tasks/FairMQSamplerTask.h` -> `examples/advanced/Tutorial3/MQ/samplerTask/SamplerTask.h`
-
+  * FairTimeStamp
+    * `virtual bool operator<(const FairTimeStamp* rValue) const` changed to `bool operator<(const FairTimeStamp& rValue) const`
+ 
 ### Deprecations
 
 If you think you really require a deprecated API, please


fairroot/base/event/FairTimeStamp.h
```diff
@@ -74,7 +74,7 @@ class FairTimeStamp : public FairMultiLinkedData_Interface
         return out;
     }
 
-    virtual bool operator<(const FairTimeStamp* rValue) const { return GetTimeStamp() < rValue->GetTimeStamp(); }
+    virtual bool operator<(const FairTimeStamp& rValue) const { return GetTimeStamp() < rValue.GetTimeStamp(); }
```
ChristianTackeGSI, line 77: 1. Please update the copyright year in the header

2. I don't remember what we concluded about making this virtual? Can somebody please comment?

3. This is a breaking change, because it drops an old API. Should we document this in the Chanelog?

4. I don't really understand, why we use the getters instead of accessing the member variables directly?

   ```suggestion
       virtual bool operator<(const FairTimeStamp& rValue) const { return fTimeStamp < rValue.fTimeStamp; }
   ```

5. Should this also compare the timestamp-error, like the `Compare` method above does?

karabowi: I would prefer to keep it `virtual` in case users need to implement their own version of `operator<`.

fuhlig1:
@TobiasStockmanns,

I don't understand why the operator is needed at all. I is neither used in FairRoot nor PandaRoot which was checked by Radek and me independently. Radek simply removed the function, I add the function with `=delete`. In both cases we did not see any compilation errors. To cross-check I add a test which uses the operator< and this file shows the following error message during compilation

```
/opt/fairroot/source/v19.0/tests/base/event/test_FairTimeStamp.cxx:20:9: error: overload resolution selected deleted operator '<'
   if (a<b) { a=2.;}
       ~^~
/opt/fairroot/source/v19.0/fairroot/base/event/FairTimeStamp.h:63:11: note: candidate function has been explicitly deleted
     bool operator<(const FairTimeStamp& rValue) = delete;
          ^
/opt/fairsoft/jan24/include/root/TString.h:778:15: note: candidate function not viable: no known conversion from 'FairTimeStamp' to 'const TString' for 1st argument
inline Bool_t operator<(const TString &s1, const TString &s2)
              ^
/opt/fairsoft/jan24/include/root/TString.h:794:15: note: candidate function not viable: no known conversion from 'FairTimeStamp' to 'const TString' for 1st argument
inline Bool_t operator<(const TString &s1, const char *s2)
              ^
/opt/fairsoft/jan24/include/root/TString.h:812:15: note: candidate function not viable: no known conversion from 'FairTimeStamp' to 'const char *' for 1st argument
inline Bool_t operator<(const char *s1, const TString &s2)
              ^
/opt/fairroot/source/v19.0/fairroot/base/event/FairTimeStamp.h:64:11: note: candidate function not viable: no known conversion from 'FairTimeStamp' to 'const FairTimeStamp *' for 1st argument; take the address of the argument with &
     bool operator<(const FairTimeStamp* rValue) = delete;
          ^
```
 
I don't see the need for the overloaded function and would propose to simply delete it or at least set it as deprecated.

TobiasStockmanns:
The original idea why I defined ```bool operator<(const FairTimeStamp* rValue)``` was to be able to sort a container of ```FairTimeStamp*``` with this operator. Unfortunately this does not work because the sort will always use the memory location and not the ```FairTimeStamp::operator<```. So this implementation does not make any sense.
Nevertheless I think it is good to have an ```bool operator<(const FairTimeStamp& rValue)``` defined, because for time based simulations and reconstructions a sorting by the time stamp is mandatory and this should be supported by the code.
In addition the ```virtual``` should be removed.

karabowi (dismissed):

fairroot/base/event/FairTimeStamp.h
```diff
@@ -74,7 +74,7 @@ class FairTimeStamp : public FairMultiLinkedData_Interface
         return out;
     }
 
-    virtual bool operator<(const FairTimeStamp* rValue) const { return GetTimeStamp() < rValue->GetTimeStamp(); }
+    bool operator<(const FairTimeStamp& rValue) const { return GetTimeStamp() < rValue.GetTimeStamp(); }
```
ChristianTackeGSI, line 77: 1. Please update the copyright year in the header

2. This is a breaking change, because it drops an old API. Should we document this in the Chanelog?

3. I don't really understand, why we use the getters instead of accessing the member variables directly?

   ```suggestion
       bool operator<(const FairTimeStamp& rValue) const { return fTimeStamp < rValue.fTimeStamp; }
   ```

4. Should this also compare the timestamp-error, like the `Compare` method above does?

ChristianTackeGSI (changes requested):
Still some open points

fairroot/base/event/FairTimeStamp.h
TobiasStockmanns: 1. done
2. What is the opionion of the others?
3. Variables are now directly used
4. No. The important comaprison is with the time stamp. An additional comparison with the error does not realy make sense.

karabowi:
@coderabbitai construct short CHANGELOG message for this PR.
@TobiasStockmanns , can you squash into one commit?

fairroot/base/event/FairTimeStamp.h
```diff
@@ -74,13 +74,13 @@ class FairTimeStamp : public FairMultiLinkedData_Interface
...
 
-    virtual bool operator<(const FairTimeStamp* rValue) const { return GetTimeStamp() < rValue->GetTimeStamp(); }
+    bool operator<(const FairTimeStamp& rValue) const { return fTimeStamp < rValue.fTimeStamp; }
 
   protected:
     Double_t fTimeStamp{-1};        //< Time of digit or Hit  [ns]
     Double_t fTimeStampError{-1};   //< Error on time stamp
 
-    ClassDefOverride(FairTimeStamp, 4);
+    ClassDefOverride(FairTimeStamp, 5);
```
ChristianTackeGSI, line 83: Why do you need to update the version? Can you please explain?

fairroot/base/event/FairTimeStamp.h
ChristianTackeGSI: About (4): The Scroll a bit up in the file `Compare` compares the error. So that method and this operator have different results? IMHO we should at least document this difference.

fairroot/base/event/FairTimeStamp.h
karabowi: Since there is no change in member variables, I think the version should stay the same.

fuhlig1: I agree. The version number should stay at 4.

ChristianTackeGSI (commented):
Please rebase and squash things.

karabowi:
> Please rebase and squash things.

Done.

karabowi (approved):

ChristianTackeGSI (approved):

PR #1527, opened 2024-04-25
Improve geo classes

fuhlig1:
Improve the *Geo* classes in the example directory. 
Move duplicate code into the base class FairGeoSet.
In the constructors use setter functions instead to access data members of the base classes directly. 
Use fmt library for name generation and parsing to get rid of sprintf.
Add and improve unit tests for these classes,

---

Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced naming and module setting mechanisms across various geometry classes, improving clarity and consistency.
  - Added new testing capabilities for Rutherford geometry settings.

- **Bug Fixes**
  - Removed redundant and obsolete methods and variables from multiple classes, streamlining geometry handling.

- **Documentation**
  - Updated copyright years and removed outdated comments across several files.

- **Refactor**
  - Consolidated functionality of methods into constructors for better initialization in geometry classes.
  - Updated and streamlined geometry handling by replacing direct assignments with method calls.

- **Chores**
  - Cleaned up CMake files to better manage dependencies and conditions for building examples.


fuhlig1:
@coderabbitai, generate unit testing code for examples/MQ/pixelDetector/src/PixelGeo.cxx and examples/MQ/pixelDetector/src/PixelGeo.h. Take also the base class fairroot/geobase/FairGeoSet.cxx and fairroot/geobase/FairGeoSet.h into account

tests/examples/simulation/rutherford/CMakeLists.txt
dennisklein, line 1: Something went wrong with the whitespace in this file?

fairroot/geobase/FairGeoSet.h
```diff
@@ -89,10 +90,21 @@ class FairGeoSet : public TNamed
...
-    virtual Int_t getModNumInMod(const TString&) { return 0; }
+    /** returns the module index in the modules array from the module name.
+        The module number is counted starting at 1, so to get the array index
+        one has to substract 1 from the module number.
+    */
+    virtual inline Int_t getModNumInMod(const TString& name)
+    {
+        // construct TString starting after the end of the detector name
+        // which contains the detector number
+        TString modNumber(name(fName.Length(), (name.Length() - fName.Length())));
```
dennisklein, line 104:
```suggestion
        assert(name.Length() > fName.Length());
        TString modNumber(name(fName.Length(), (name.Length() - fName.Length())));
```

Makes it a bit more readable IMHO, as this is really the assumption you are making, right?

fairroot/geobase/FairGeoSet.h
```diff
@@ -89,10 +90,21 @@ class FairGeoSet : public TNamed
...
+        return (modNumber.Atoi() - 1);
```
dennisklein, line 105: Will this always work? e.g. if `fName == "blubb"` and `name == "blubbXXX"` or `name == "asdf15"`, let's add test coverage?

Perhaps this function should also check that the prefixes are equal? Otherwise, if `fName == "blubb"`, `getModNumInMod("asdfg42")` will work, but should it?

dennisklein (changes requested):
In general :+1:, I put a few comments inline.

fairroot/geobase/FairGeoSet.cxx
```diff
@@ -517,3 +518,30 @@ void FairGeoSet::compare(FairGeoSet& rset)
...
+
+const char* FairGeoSet::getModuleName(Int_t m)
+{
+    /** Returns the module name of the detector with array index m
+        Since the name uses natuaral counting the index has to be
+        increased by 1. All module names in the ASCII file needs to
+        start with the name saved in fName since otherwise they are
+        not constructed.
+    */
+    assert(m >= 0);
```
dennisklein, line 530: Needs `#include <cassert>` (https://en.cppreference.com/w/cpp/error/assert)

fairroot/geobase/FairGeoSet.h
```diff
@@ -89,10 +90,21 @@ class FairGeoSet : public TNamed
     TString& getAuthor() { return author; }
     TString& getDescription() { return description; }
     virtual const char* getKeepinName(Int_t, Int_t) { return 0; }
-    virtual const char* getModuleName(Int_t) { return 0; }
-    virtual const char* getEleName(Int_t) { return 0; }
+    virtual const char* getModuleName(Int_t);
+    virtual const char* getEleName(Int_t);
```
dennisklein, line 94: Actually, thinking more about those functions: I think we should go one step further and change the return type and return a string (by-value). Returning a `const char *` is really dangerous, because those functions look like ordinary getters: all results of a previous call become dangling pointers as soon as the next call happens.

fuhlig1: I thought about the same but unfortunately these functions are public and definitely used in user code so we can't change it without breaking the user code. 
Is it possible to implement a second function which only has a different return type? If so we can deprecate the old function and internally use the new one.

fairroot/geobase/FairGeoSet.h
fuhlig1: The ASCII files defining the detector geometry need to implement it the way that all elements in the file start with `fName` followed by a number. I will add code to check this.

fairroot/geobase/FairGeoSet.h
fuhlig1: Done.

fairroot/geobase/FairGeoSet.h
fuhlig1: I added two checks which confirm the consistency of the input.

fairroot/geobase/FairGeoSet.cxx
fuhlig1: Thanks for catching this. If used now the fairlogger LOG_IF instead of assert.

fairroot/geobase/FairGeoSet.h
```diff
@@ -104,6 +112,9 @@ class FairGeoSet : public TNamed
   private:
     FairGeoSet(const FairGeoSet&);
     FairGeoSet& operator=(const FairGeoSet&);
+
+    std::string modNameSet{""};   //! name of module
+    std::string eleNameSet{""};   //! substring for elements in module
```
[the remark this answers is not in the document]
fuhlig1: Done.

karabowi (dismissed):

fairroot/geobase/FairGeoSet.h
dennisklein: Yes, that would be the better approach for a smoother transition.

fairroot/geobase/FairGeoSet.h
```diff
@@ -76,6 +77,7 @@ class FairGeoSet : public TNamed
     const char* getGeomFile() { return geoFile.Data(); }
     Int_t getMaxSectors(void) { return maxSectors; }
     Int_t getMaxModules(void) { return maxModules; }
+    void setMaxModules(Int_t _maxMods) { maxModules = _maxMods; }
```
ChristianTackeGSI, line 80: I think, the general naming convention was about CamelCase, right?

Please let's at least use it for new APIs?

```suggestion
    void SetMaxModules(Int_t _maxMods) { maxModules = _maxMods; }
```

tests/examples/simulation/rutherford/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+#Copyright(C) 2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH #
+# #
+#This software is distributed under the terms of the #
+#GNU Lesser General Public Licence(LGPL) version 3, #
+#copied verbatim in the file "LICENSE" #
+    ################################################################################
+
+#Test suite for the `FairRoot::ExRutherford` library
+
+    fairroot_add_catch2_test_suite(ExRutherford SOURCES test_FairRutherfordGeo.cxx
```
ChristianTackeGSI, line 11: Something is still wrong with the whitespace here.

tests/geobase/checks_FairGeoSet.h
```diff
@@ -30,7 +30,7 @@ auto checkGeoSetNamingConventions(FairGeoSet& geo, const std::string& prefix, in
 
     SECTION("name generation")
     {
-        const Int_t index = GENERATE(0, 1, 99);
+        auto index = GENERATE(take(100, random(1, 1000)));
```
ChristianTackeGSI, line 33: `const auto` does not work?

examples/simulation/rutherford/src/FairRutherfordGeo.cxx
```diff
@@ -1,36 +1,16 @@
...
-    std::snprintf(modName, sizeof(modName), "rutherford%i", m + 1);
-    return modName;
-}
-
-const char* FairRutherfordGeo::getEleName(Int_t m)
-{
-    /** Returns the element name of Det number m */
-    std::snprintf(eleName, sizeof(eleName), "rutherford%i", m + 1);
-    return eleName;
+    SetName("rutherford");
```
ChristianTackeGSI, line 14: Instead of `SetName` in the ctor, what about a new ctor for FairGeoSet that takes a `const char*` and calls the `TNamed(name, "")` ctor?

ChristianTackeGSI (changes requested):

examples/simulation/rutherford/src/FairRutherfordGeo.cxx
ChristianTackeGSI: Maybe even a `FairGeoSet(const char*, unsigned int maxmodules = 0)`?

fairroot/geobase/FairGeoSet.cxx
dennisklein: But `LOG_IF` you always pay for even in Release builds? The behaviour of `assert` is that it becomes a no-op with `-DNDEBUG`.

fairroot/geobase/FairGeoSet.cxx
```diff
@@ -517,3 +519,56 @@ void FairGeoSet::compare(FairGeoSet& rset)
...
+    // which contains the detector number
+    TString modNumber(name(fName.Length(), (name.Length() - fName.Length())));
+
+    // check if the conversion worked as expected. Atoi internally calls the c function
+    // atoi which returns in case of failure the value 0. Since in our case a value of
+    // 0 is an error anyway (Natural counting starts at 1 as well as the counting of elements
+    // in the module) a value of 0 indicates an error
+    int retVal = modNumber.Atoi();
+    if (0 == retVal) {
+        LOG(fatal) << "Naming conventions violated. The element name must start with " << GetName()
```
dennisklein, line 542:
```suggestion
    const auto& prefix = fName;
    const auto prefixLength = prefix.Length();
    const auto length = name.Length();
    // check if the naming is correct. All modules in a geometry file must start with
    // the generic module name followed by a number.
    if ((!name.BeginsWith(prefix)) || (length <= prefixLength)) {
        LOG(fatal) << "Naming conventions violated. The element name must start with " << prefix
                   << " followed by a unique number.\n"
                   << "The passed value was " << name;
    }
    // construct TString starting after the end of the detector name
    // which contains the detector number
    TString modNumber(name(prefixLength, (length - prefixLength)));

    // check if the conversion worked as expected. Atoi internally calls the c function
    // atoi which returns in case of failure the value 0. Since in our case a value of
    // 0 is an error anyway (Natural counting starts at 1 as well as the counting of elements
    // in the module) a value of 0 indicates an error
    int retVal = modNumber.Atoi();
    if (0 == retVal) {
        LOG(fatal) << "Naming conventions violated. The element name must start with " << prefix
```

Yes, in optimized builds all those function calls are probably inlined because they are defined in header files, but I think the variables state a much clearer intent which also improves readability, no?

fairroot/geobase/FairGeoSet.cxx
fuhlig1: The problem I see is that the files passed to the ASCII interface are a runtime dependency and can be changed anytime. So the check if the format is correct should be done in any case and the execution should stop if a problem is detected. In my opinion `LOG_IF` is the proper solution here.

fairroot/geobase/FairGeoSet.h
fuhlig1: Correct. I kept the old convention for the geobase area since this was inherited from Hades. I agree that we should change this but I would like to do it in a second step. There are many more issues with this class which I would like to fix.
The main intention of this PR was to get rid of the sprintf usage. I would like  to limit the scope of this PR to this.

tests/examples/simulation/rutherford/CMakeLists.txt
fuhlig1: Fixed. I think by accident I format the file using clang-format.

examples/simulation/rutherford/src/FairRutherfordGeo.cxx
fuhlig1: I agree but as said previously the main topi of this PR is to remove the usage of sprintf, so I would like to not change this now but with additional PRs.

fairroot/geobase/FairGeoSet.cxx
fuhlig1: Changed. I agree that the suggestion improved the readability.

fairroot/geobase/FairGeoSet.h
ChristianTackeGSI: We probably should start by seperating the interfaces that should be called and that should be overriden (there's some cool "pattern" name for this, I have forgotten it).

We should deprecate `getModuleName` so that people stop calling it.

And here's an idea for the "to be called" method.

```cpp
std::string CallGetModuleName(Int_t x)
{
    // getModuleName usually is not thread safe!
    std::mutex mtx;
    std::lock_guard lock{mtx};

#pragma GCC diagnostic push
    return {getModuleName(x)};
#pragma GCC diagnostic pop
}
```

fairroot/geobase/FairGeoSet.h
ChristianTackeGSI: Okay, I "unresolve" this for another point:

Some of the inner workings of FairGeoSet (especially the usage of the `modules` member variable) depend on `maxModules` not changing during operation. Providing a simple setter like this suggests that changing maxModules at any time is allowed.

Either this method should check that `modules` is not already allocated.

Or (preferred) this method should not exist at all and the maxmodules parameter should be part of the ctor.

If you don't want to go the ctor route in this PR, then maybe leave out the complete maxModules part at all and move it to another PR?

ChristianTackeGSI: BTW: The same is probably true for `maxSectors`?

fairroot/geobase/FairGeoSet.h
fuhlig1: Good idea but we should try to get this PR merged and create issues with the still open problems/suggestions. 
We should now focus on getting the warnings fixed such that we can create the new FairRoot release.

karabowi:
We will address the requested changes for v19.2.

PR #1529, opened 2024-05-03
Remove sprintf

karabowi:
Remove `sprintf` from `generators`, `geobase` and `parbase` directories.

Mostly replaced by `fmt::format`, only in FairRtdbRun replaced by `TString::Form`.

---

Checklist:

## Summary by CodeRabbit

- **Refactor**
  - Enhanced string formatting in various modules by implementing the `fmt` library, replacing `sprintf` with `fmt::format` for improved code readability and maintenance. Affected modules include:
    - `FairDetParAsciiFileIo`
    - `FairDetParRootFileIo`
    - `FairRtdbRun`
    - `FairRuntimeDb`

### DIFF
--- a/fairroot/generators/FairIonGenerator.cxx
+++ b/fairroot/generators/FairIonGenerator.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -20,8 +20,8 @@
 #include <TObjArray.h>      // for TObjArray
 #include <TParticle.h>      // for TParticle
 #include <TParticlePDG.h>   // for TParticlePDG
-#include <cstdio>           // for sprintf
 #include <fairlogger/Logger.h>
+#include <fmt/core.h>   // for format
 
 Int_t FairIonGenerator::fgNIon = 0;
 
@@ -107,9 +107,7 @@ FairIonGenerator::FairIonGenerator(Int_t z,
   fVy   = vy;
   fVz   = vz;
   */
-    char buffer[20];
-    sprintf(buffer, "FairIon%d", fgNIon);
-    fIon = new FairIon(buffer, z, a, q);
+    fIon = new FairIon(fmt::format("FairIon{}", fgNIon).c_str(), z, a, q);
     FairRunSim* run = FairRunSim::Instance();
     if (!run) {
         LOG(error) << "No FairRun instantised!";
@@ -141,9 +139,15 @@ FairIonGenerator::~FairIonGenerator()
     // if (fIon) delete fIon;
 }
 
-void FairIonGenerator::SetExcitationEnergy(Double_t eExc) { fIon->SetExcEnergy(eExc); }
+void FairIonGenerator::SetExcitationEnergy(Double_t eExc)
+{
+    fIon->SetExcEnergy(eExc);
+}
 
-void FairIonGenerator::SetMass(Double_t mass) { fIon->SetMass(mass); }
+void FairIonGenerator::SetMass(Double_t mass)
+{
+    fIon->SetMass(mass);
+}
 
 Bool_t FairIonGenerator::ReadEvent(FairPrimaryGenerator* primGen)
 {
--- a/fairroot/generators/FairShieldGenerator.cxx
+++ b/fairroot/generators/FairShieldGenerator.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -19,7 +19,7 @@
 #include <TDatabasePDG.h>   // for TDatabasePDG
 #include <TParticlePDG.h>   // for TParticlePDG
 #include <climits>          // for INT_MAX
-#include <cstdio>           // for sprintf
+#include <fmt/core.h>       // for format
 #include <fstream>          // for ifstream
 #include <utility>          // for pair
 
@@ -51,7 +51,10 @@ FairShieldGenerator::FairShieldGenerator(const char* fileName)
     fInputFile = new std::ifstream(fFileName);
 }
 
-FairShieldGenerator::~FairShieldGenerator() { CloseInput(); }
+FairShieldGenerator::~FairShieldGenerator()
+{
+    CloseInput();
+}
 
 Bool_t FairShieldGenerator::ReadEvent(FairPrimaryGenerator* primGen)
 {
@@ -101,9 +104,8 @@ Bool_t FairShieldGenerator::ReadEvent(FairPrimaryGenerator* primGen)
 
         // Case ion
         if (iPid == 1000) {
-            char ionName[20];
-            sprintf(ionName, "Ion_%d_%d", iMass, iCharge);
-            TParticlePDG* part = fPDG->GetParticle(ionName);
+            const auto ionName = fmt::format("Ion_{}_{}", iMass, iCharge);
+            TParticlePDG* part = fPDG->GetParticle(ionName.c_str());
             if (!part) {
                 LOG(warn) << "FairShieldGenerator::ReadEvent: Cannot find " << ionName << " in database!";
                 continue;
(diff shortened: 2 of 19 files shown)


fairroot/base/sim/FairIon.cxx
```diff
@@ -23,7 +23,7 @@ FairIon::FairIon()
     , fMass(0)
 {}
 
-FairIon::FairIon(const char* name, Int_t z, Int_t a, Int_t q, Double_t e, Double_t mass)
+FairIon::FairIon(const std::string name, Int_t z, Int_t a, Int_t q, Double_t e, Double_t mass)
     : TNamed(name, "User defined ion")
```
dennisklein, line 27:
```suggestion
FairIon::FairIon(const TString& name, Int_t z, Int_t a, Int_t q, Double_t e, Double_t mass)
    : TNamed(name, "User defined ion")
```

When to use the string `by-value and move` pattern in ctors is discussed here: https://www.cppstories.com/2018/08/init-string-member/

Since we are forced by the `TNamed` ctor to create a temporary `TString` anyways, I guess,  there is nothing better to do here.

fairroot/base/sim/FairIon.h
```diff
@@ -40,10 +40,10 @@ class FairIon : public TNamed
      ** If mass is not given, it will be set to a times the proton mass.
      **/
 
-    FairIon(const char* name, Int_t z, Int_t a, Int_t q, Double_t e = 0., Double_t m = 0.);
-    void SetParams(const char* name, Int_t z, Int_t a, Int_t q, Double_t e = 0., Double_t m = 0.)
+    FairIon(const std::string name, Int_t z, Int_t a, Int_t q, Double_t e = 0., Double_t m = 0.);
+    void SetParams(const std::string name, Int_t z, Int_t a, Int_t q, Double_t e = 0., Double_t m = 0.)
```
dennisklein, line 44: Just leave the `const char*` here. Taking the `std::string` by-value just adds a useless copy. Unfortunately, `TNamed::SetName()` does not offer any suitable overload that allows modernization.

dennisklein (changes requested):
Please check also for missing `#include <string>`s!

dennisklein:
FYI: Check https://www.conventionalcommits.org/en/v1.0.0/#summary (That is where the `fix:` comes from when @ChristianTackeGSI or I use them)

e.g.
* fix(generators): Remove `sprintf`
* fix(geobase): Remove `sprintf`
* fix(parbase): Remove `sprintf`

fairroot/base/sim/FairIon.cxx
dennisklein: [`TString::operator const char*`](https://root.cern.ch/doc/master/classTString.html#a58d6744a8da98ac00c160809a15645ba) :((( - how can it be that ROOT uses every anti-pattern that is possible...

ChristianTackeGSI: You could disambiguate this maybe?
```cpp
TNamed(name, TString{"User defined ion"})
```

ChristianTackeGSI: Personal note: I don't like the proliferation of TString into our public APIs. Especially when we can stay with more generic ones like `const char*` (or `czstring`–yes, please let's discuss somewhere else)

If https://github.com/root-project/root/issues/15434 comes into reality, we could consider `string_view`.

fairroot/generators/FairIonGenerator.cxx
```diff
@@ -107,9 +108,7 @@ FairIonGenerator::FairIonGenerator(Int_t z,
   fVy   = vy;
   fVz   = vz;
   */
-    char buffer[20];
-    sprintf(buffer, "FairIon%d", fgNIon);
-    fIon = new FairIon(buffer, z, a, q);
+    fIon = new FairIon(fmt::format("FairIon{}", fgNIon).c_str(), z, a, q);
```
ChristianTackeGSI, line 111: Please assign the return value of `fmt::format` to a local variable to make sure that its life time is long enough.

```suggestion
    std::string ionName = fmt::format("FairIon{}", fgNIon)
    fIon = new FairIon(ionName.c_str(), z, a, q);
```

(There are more places. I just used this one.)

dennisklein: https://stackoverflow.com/a/10006989 Googled, this claims this should work out.

ChristianTackeGSI: Hmmm.

Then the code should be consistent, please. Either try to avoid local variables wherever possible or use them everywhere, please.

fairroot/geobase/FairGeoCone.cxx
```diff
@@ -40,10 +40,11 @@
 #include <TArrayD.h>     // for TArrayD
 #include <TMathBase.h>   // for Abs
 #include <TString.h>     // for TString
+#include <fmt/core.h>    // for format
 #include <fstream>
 #include <ostream>    // for fstream, etc
 #include <stdio.h>    // for printf, sprintf, sscanf
-#include <string.h>   // for strlen
+#include <string.h>   // for string, strlen
```
dennisklein, line 47: * `std::string` is in `<string>` (https://en.cppreference.com/w/cpp/string/basic_string)
* `std::strlen` is in `<cstring>` (https://en.cppreference.com/w/cpp/string/byte/strlen)

`<string.h>` is a C std library header (and declares `::strlen` https://en.cppreference.com/w/c/string/byte/strlen)

ChristianTackeGSI, line 47: Note `<string.h>` and [`<string>`](https://en.cppreference.com/w/cpp/header/string) are different headers.

And I think, we should replace `<string.h>` by [`<cstring>`](https://en.cppreference.com/w/cpp/header/cstring)

fairroot/parbase/FairRtdbRun.h
```diff
@@ -101,9 +101,7 @@ inline void FairRtdbRun::setRefRun(Int_t r)
     if (r == -1) {
         refRun = "";
     } else {
-        char name[255];
-        sprintf(name, "%i", r);
-        refRun = name;
+        refRun = Form("%i", r);
```
ChristianTackeGSI, line 104: What about [`TString::Form`](https://root.cern.ch/doc/master/classTString.html#aeb44d6183d8b1f1b7090dbd3c93f5e39)?

```suggestion
        refRun.Form("%i", r);
```

Also if we move to `fmt::format`, we should do it here as well?

karabowi: Changed to your suggestion. Does it make sense to include `fmt/core.h` in this header only for something
we can get with `TString`?

fairroot/parbase/FairDetParAsciiFileIo.cxx
```diff
@@ -44,8 +46,7 @@ Bool_t FairDetParAsciiFileIo::findContainer(const Text_t* name)
     // searches the container in the file
     const Int_t maxbuf = 4000;
     Text_t buf[maxbuf];
-    Text_t buf2[maxbuf];
-    sprintf(buf2, "%s%s%s", "[", name, "]");
+    std::string buf2 = fmt::format("[{}]", name);
```
dennisklein, line 49: Can we pls find a more descriptive variable name?

dennisklein (changes requested):
* Some of the included `<string>` headers are not used. You only need those if a name from it is used explicitely.
* Some of my inline comments apply in multiple places, I only commented once.

fairroot/generators/FairShieldGenerator.cxx
```diff
@@ -152,11 +155,9 @@ Int_t FairShieldGenerator::RegisterIons()
         for (Int_t iTrack = 0; iTrack < nTracks; iTrack++) {
             *fInputFile >> iPid >> iMass >> iCharge >> px >> py >> pz;
             if (iPid == 1000) {   // ion
-                char buffer[20];
-                sprintf(buffer, "Ion_%d_%d", iMass, iCharge);
-                TString ionName(buffer);
+                std::string ionName = fmt::format("Ion_{}_{}", iMass, iCharge);
                 if (fIonMap.find(ionName) == fIonMap.end()) {   // new ion
```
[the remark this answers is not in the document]
ChristianTackeGSI: I would propose to move the map changing to a new PR.

Here's a proposal for that PR:
- #1535

I will update #1535 once this one here gets merged. So I propose to resolve this thread.

dennisklein (dismissed):
Addressed all issues from my last review round

fairroot/parbase/FairDetParAsciiFileIo.cxx
```diff
@@ -42,24 +45,26 @@ FairDetParAsciiFileIo::FairDetParAsciiFileIo(std::fstream* f)
...
-        if (!strncmp(buf, buf2, strlen(buf2))) {
-            break;
+        if (const auto linePrefix = std::string_view(line).substr(0, searchValueLength);
+            linePrefix == searchValue) {
+            return true;
         }
     }
-    return !pFile->eof();
+    return false;
 }
```
dennisklein, line 68: @karabowi @ChristianTackeGSI Please double check this. Not sure if we have test coverage.

ChristianTackeGSI: Looks good to me. I still wonder why this thing needs to check only the prefix. But I haven't investigated the code around.

dennisklein: Na, if you cannot read it, we should put it into a function `startsWith()`

dennisklein: How does the commented C++20 code read to you compared to `!strncmp(buf, buf2, ...)`?

karabowi: I said I need time to digest it, not that I cannot read it.

dennisklein: Sure, but you made the argument, it is less readable to you. And my main point of changing this code is to make it more readable ;)

ChristianTackeGSI (approved):
There are some things left. But we can handle them in upcoming PRs.

Issue #1530, opened 2024-05-06
Further cleanup ascii geometry related classes


fuhlig1:
In #1527 many more improvements for the usage of the ascii geometry classes were proposed but not implemented since the release should be done as sone as possible.

Look at the suggestions again after the release and implement them.

PR #1531, opened 2024-05-06
FairGeoSet ctor improvements / fixes

ChristianTackeGSI:
Especially remove the just introduced `setMaxModules()` API, which just does not fulfill the classes invariants.

fixes part of #1530

---

Checklist:

## Summary by CodeRabbit

- **Refactor**
	- Simplified constructors across various modules by directly initializing parameters, enhancing consistency and reducing complexity.

### DIFF
--- a/fairroot/geobase/FairGeoSet.h
+++ b/fairroot/geobase/FairGeoSet.h
@@ -63,6 +63,12 @@ class FairGeoSet : public TNamed
     TString description{""};         /** description of the version*/
 
     FairGeoSet() = default;
+    FairGeoSet(const char* name, Int_t maxMods, Int_t maxSecs = 0)
+        : TNamed{name, ""}
+        , maxSectors{maxSecs}
+        , maxModules{maxMods}
+    {}
+
     void readInout(std::fstream&);
     void readTransform(std::fstream&, FairGeoTransform&);
     Bool_t readVolumeParams(std::fstream&, FairGeoMedia*, FairGeoNode*, TList* l = 0);
--- a/fairroot/geobase/FairGeoSet.h
+++ b/fairroot/geobase/FairGeoSet.h
@@ -83,7 +83,6 @@ class FairGeoSet : public TNamed
     const char* getGeomFile() { return geoFile.Data(); }
     Int_t getMaxSectors() { return maxSectors; }
     Int_t getMaxModules() { return maxModules; }
-    void setMaxModules(Int_t _maxMods) { maxModules = _maxMods; }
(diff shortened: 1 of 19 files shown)


fairroot/geobase/FairGeoSet.h
```diff
@@ -51,18 +51,24 @@ class FairGeoSet : public TNamed
...
+    Int_t maxKeepinVolumes{0};       /** maximum number of keepin volumes per sector*/
+    Int_t maxModules{0};             /** maximum number of modules per sector*/
+    TArrayI* modules{nullptr};       /** Module's array.*/
+    FairGeoShapes* pShapes{nullptr}; /** pointer to the class FairGeoShapes*/
+    TString geoFile{""};             /** name of geometry input file or Oracle*/
+    TString author{""};              /** author of the media version*/
+    TString description{""};         /** description of the version*/
+
+    FairGeoSet() = default;
+    FairGeoSet(const char* name, Int_t maxMods, Int_t maxSecs = 0)
```
fuhlig1, line 66: Why don't use a std::string& ?

fuhlig1 (approved):
Beside my small remark I am fine with the PR.

fairroot/geobase/FairGeoSet.h
ChristianTackeGSI: Because `TNamed(name, …)` also only has a `const char*`. so we'd need to construct a full `std::string` just do call `.c_str()` on it. Don't know, if this is really useful.

We could go for a `std::string_view`. This would then be converted to a TString which would be copy-constructed into the TString in TNamed.

dennisklein: > Because `TNamed(name, …)` also only has a `const char*`.

And this [`TNamed::TNamed(const TString&, const TString&)`](https://root.cern.ch/doc/master/classTNamed.html#a92b15a72a05ddf8f23c67023a29ee6a9) - so perhaps a `const TString&`?

dennisklein: For the pure `const char*` case I found this: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rstr-zstring

ChristianTackeGSI: You mean some `fairroot::czstring`?

Because using gsl would mean introducing gsl as a public dependency. And as far as I remember, using it as a public dependency was not trivial.

dennisklein: > You mean some fairroot::czstring?

yes, but perhaps out of scope of this PR. Let's discuss the pros and cons in one of the upcoming meetings?

PR #1540, opened 2024-05-26
 fix(templates): Use naming logic from base class

dennisklein:
introduced in 0fb8a6f42be762ee1ea6f0aa9cdfe3e62f8be100

Follow-up to #1527

## Summary by CodeRabbit

- **Refactor**
  - Simplified the `NewDetectorGeo` class by removing unused constructors and methods.
  - Updated the constructor to initialize with specific parameters for better performance and clarity.

- **Chores**
  - Updated copyright year to 2024.
  - Removed obsolete member variables and methods to streamline the codebase.


templates/NewDetector_root_containers/NewDetectorGeo.h
```diff
@@ -1,37 +1,25 @@
...
 #define NEWDETECTORGEO_H
 
 #include "FairGeoSet.h"
 
 class NewDetectorGeo : public FairGeoSet
 {
-  protected:
-    char modName[20]{"NewDetector"};   // name of module
-    char eleName[20]{"NewDetector"};   // substring for elements in module
   public:
```
ChristianTackeGSI, line 16: If everything is public anyway, maybe use a struct?

dennisklein: I guess, this should be discussed in general, how people feel about using `struct` here.

In any case, it is out of scope of this PR imho.

ChristianTackeGSI (approved):

PR #1544, opened 2024-05-29
refactor(Base): Introduce FairGeoLoader::LoadAndCreate

ChristianTackeGSI:
... and use it around

@fuhlig1: you looked a lot at GeoSets lately. Can you have a look at this? Especially it's only used in ASCII related contexts, so maybe it should be called `PrepareASCIIGeoSet`?

Also I am not really sure on how to write tests for this.

- [x] #1564

---

Checklist:

### DIFF
--- a/examples/common/passive/FairCave.cxx
+++ b/examples/common/passive/FairCave.cxx
@@ -9,7 +9,6 @@
 #include "FairCave.h"
 
 #include "FairGeoCave.h"         // for FairGeoCave
-#include "FairGeoInterface.h"    // for FairGeoInterface
 #include "FairGeoLoader.h"       // for FairGeoLoader
 #include "FairGeoNode.h"         // for FairGeoNode
 #include "FairGeoPassivePar.h"   // for FairGeoPassivePar
@@ -24,17 +23,9 @@
 
 void FairCave::ConstructGeometry()
 {
-    FairGeoLoader& loader = GetGeometryLoader();
-    FairGeoInterface* GeoInterface = loader.getGeoInterface();
-    FairGeoCave* MGeo = new FairGeoCave();
-    MGeo->setGeomFile(GetGeometryFileName());
-    GeoInterface->addGeoModule(MGeo);
-    Bool_t rc = GeoInterface->readSet(MGeo);
-    if (rc) {
-        MGeo->create(loader.getGeoBuilder());
-    }
+    FairGeoSet& MGeo = GetGeometryLoader().LoadAndCreate<FairGeoCave>(GetGeometryFileName());
 
-    TList* volList = MGeo->getListOfVolumes();
+    TList* volList = MGeo.getListOfVolumes();
     // store geo parameter
     FairRun* fRun = FairRun::Instance();
     FairRuntimeDb* rtdb = FairRun::Instance()->GetRuntimeDb();
--- a/fairroot/base/sim/FairModule.h
+++ b/fairroot/base/sim/FairModule.h
@@ -11,8 +11,8 @@
 #include "FairGeoInterface.h"   // for FairGeoInterface
 #include "FairGeoLoader.h"      // for FairGeoLoader
 #include "FairGeoNode.h"        // for FairGeoNode
-#include "FairGeoVolume.h"      // for FairGeoVolume
-#include "FairLogger.h"
+#include "FairGeoSet.h"
+#include "FairGeoVolume.h"
 #include "FairRun.h"         // for FairRun
 #include "FairRuntimeDb.h"   // for FairRuntimeDb
 
@@ -26,7 +26,10 @@
 #include <TRefArray.h>   // for TRefArray
 #include <TString.h>     // for TString, operator!=
 #include <TVirtualMC.h>
+#include <fairlogger/Logger.h>
 #include <string>
+#include <type_traits>
+#include <utility>
 #include <vector>
 
 class FairRunSim;
@@ -197,24 +200,17 @@ class FairModule : public TNamed
 template<class T, class U>
 void FairModule::ConstructASCIIGeometry(TString containerName)
 {
-    FairGeoLoader& loader = GetGeometryLoader();
-    FairGeoInterface* GeoInterface = loader.getGeoInterface();
-    T* MGeo = new T();
-    MGeo->print();
-    MGeo->setGeomFile(GetGeometryFileName());
-    GeoInterface->addGeoModule(MGeo);   // takes ownership!
-    Bool_t rc = GeoInterface->readSet(MGeo);
-    if (rc) {
-        MGeo->create(loader.getGeoBuilder());
-    }
+    static_assert(std::is_base_of_v<FairGeoSet, T>);
+    static_assert(std::is_base_of_v<FairParSet, U>);
+    FairGeoSet& MGeo = GetGeometryLoader().LoadAndCreate<T>(GetGeometryFileName());
 
-    TList* volList = MGeo->getListOfVolumes();
+    TList* volList = MGeo.getListOfVolumes();
     // store geo parameter
     FairRun* fRun = FairRun::Instance();
     FairRuntimeDb* rtdb = FairRun::Instance()->GetRuntimeDb();
 
     if ("" != containerName) {
-        LOG(info) << "Add GeoNodes for " << MGeo->getDescription() << " to container " << containerName;
+        LOG(info) << "Add GeoNodes for " << MGeo.getDescription() << " to container " << containerName;
 
         //    U par=(U)(rtdb->getContainer(containerName));
         U* par = static_cast<U*>(rtdb->getContainer(containerName));
--- a/fairroot/geobase/FairGeoLoader.h
+++ b/fairroot/geobase/FairGeoLoader.h
@@ -17,6 +17,7 @@
 #define FairGeoLOADER_H
 
 #include "FairGeoInterface.h"
+#include "FairGeoSet.h"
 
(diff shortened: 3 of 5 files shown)


fuhlig1:
Looks at first glance reasonable. I will have a closer look next week.

dennisklein (commented):
Can it be `static`? Alternatively, a free function?

dennisklein:
> Can it be `static`? Alternatively, a free function?

Shouldn't this just be a member function `void FairGeoSet::Prepare()`? (Perhaps even better name?)

Otherwise it looks like a factory and then I wouldn't like that it hard-codes a heap object.

dennisklein:
> Shouldn't this just be a member function `void FairGeoSet::Prepare()`? (Perhaps even better name?)

Perhaps `FairGeoSet::Load()`?

ChristianTackeGSI:
> > Can it be `static`? Alternatively, a free function?
> 
> Shouldn't this just be a member function `void FairGeoSet::Prepare()`? (Perhaps even better name?)

```cpp
MGeo->setGeomFile(GetGeometryFileName());
```

`GetGeometryFileName()` is a member function of FairModule. So FairGeoSet::Prepare…() should get a reference to the module. Which might be a good thing!

> Otherwise it looks like a factory and then I wouldn't like that it hard-codes a heap object.

It is a factory. Yes. Sadly.

```cpp
GeoInterface->addGeoModule(MGeo);   // takes ownership!
```

We would first need to change that in some way to handle non-heap objects in a proper way.

Of course we could change the prepare method to take a `std::unique_ptr<FairGeoSet>` and that way expect a passed in pre-created object.

---

Putting this Prepare factory in FairModule was the most straight forward refactoring. Because it is called from inside member functions of classes that derive from FairModule.

And I wanted the shortest syntax on the calling side.

```cpp
PrepareASCIIGeoSet<MyCaveGeoSet>();
// vs.
auto thing = std::make_unique<MyCaveGeoSet>();
PrepareASCIIGeoSet(std::move(thing));
// vs. 
FairGeoSet::PrepareASCIIForModule<MyCaveGeoSet>(*this);
// vs.
auto thing = std::make_unique<MyCaveGeoSet>();
FairGeoSet::PrepareASCIIForModule<MyCaveGeoSet>(std::move(thing), *this);
```

Maybe in some way my proposed variant is something like a "facade"? You're of course free to do everything that is included in this function on your own (like the current template suggests to do), but you can also use the helper.

dennisklein:
> > > Can it be `static`? Alternatively, a free function?
> > 
> > 
> > Shouldn't this just be a member function `void FairGeoSet::Prepare()`? (Perhaps even better name?)
> 
> ```c++
> MGeo->setGeomFile(GetGeometryFileName());
> ```

Hmm, dang.

ChristianTackeGSI:
Proposal:

Let's admit, that this is a factory. A factory somewhere between `make_unique<T>(…)` and `vector::emplace(…)`. It returns a reference (like emplace, basically) to memory managed by GeoInterface.

Granted, this might not be nice/elegant from what we have learned by now. But it resembles quite closely what the refactored code is performing.

Refactoring it should also help us to better understand the surrounding code. (Like why is the calling context different in the templates, the examples and ConstructASCIIGeometry?)

- Let's keep the proposed signature for now.
- Let's rename it to some name signalling that it is a factory.
  - Something with "Make" / "Construct" / "emplace" in it? Something with GeoInterface in it?
  - It's okay for that name to be longer. It's not expected to be called a lot of times anyway.
  - This should leave enough room for better names with better semantics. And leave room for deprecating it later on.
- We can then refactor more from there in upcoming PRs

ChristianTackeGSI:
What about

```cpp
/**
 * \brief Low level FairGeoSet factory
 */
template<class T, class... Args>
FairGeoSet& FairModule::GeoInterfaceEmplaceGeoSet(Args&&... args)
```

dennisklein:
Naming that factored out function feels awkward to me, but I think for a good reason. It is bundling together too many things. After all, this is a template for the user to copy and extend on. Let's take an example where the user wants to setup `n` geometries. Would we want her to then call

```cpp
PrepareGeoSet<MyCave>();
PrepareGeoSet<MyDetectorA>();
PrepareGeoSet<MyDetectorB>();
```

? To me, the proposed function does not compose well.

ChristianTackeGSI:
ACK generally!

The function is not used in many places (actually only one in the examples, one in each of the two templates, and internally). So this probably isn't a common building block.

And TBH I don't intend it to be a common building block. My current intention is just to make the current "workflow" more obvious, as in "Okay, we have code replication here".

That's also why I am proposing a lengthy and verbose name: So it does not collide with any future, better APIs and is more easily deprecated then.

About the multiple geometries: I think, there will only be one FairGeoSet per FairModule / FairDetector. At least things in the examples look like there is a dedicated GeoSet-class per FairDetector/FairModule class. So that's not really the issue, I'm guessing.
Rather, what you mentioned earlier: One probably wants to move the ownership to the FairModule/Detector (probably as a valued member). And maybe then a `PrepareGeoSet(FairGeoSet&);` would be a much better API.

dennisklein:
Just one more attempt with yet another API:
```cpp
void MyCave::ConstructGeometry()
{
    auto& cave = GetGeometryLoader().Load<MyGeoCave>(GetGeometryFileName());
    GetGeometryBuilder().Build(cave);
}
```

1. Hide the singleton in the getter
2. Demo all *four* user choices: "*Loader* + *`GeoSet`* + *File*" aka deserialization and *Builder* aka in-memory transformation
3. Somewhat still see the composable steps

or, if you want to stick with a single function, maybe `LoadAndBuildGeometry<MyGeoCave>(GetGeometryFileName())`?

* *Construct* implemented as *Load* + *Build* - ??
   * Hmm, I guess *loading* is somewhat generally understood as reading and parsing from disk/network, but what is *building*...
* *Construct* implemented as *Prepare* - definitely :-1: 
* *Construct* implemented as *GeoInterfaceEmplaceGeoSet* - I wouldn't want to see the type names in the method names

ChristianTackeGSI:
I like the `GetGeometryLoader()`! Let's work on that first:
- #1564

dennisklein:
> > Just one more attempt with yet another API:
> > ```c++
> > void MyCave::ConstructGeometry()
> > {
> >     auto& cave = GetGeometryLoader().Load<MyGeoCave>(GetGeometryFileName());
> >     GetGeometryBuilder().Build(cave);
> > }
> > ```
> 
> ```c++
>     Bool_t rc = GeoInterface->readSet(MGeo);
>     if (rc) {
>         MGeo->create(loader->getGeoBuilder());
>     }
> ```
> 
> I wonder how one would handle the `rc` forwarding between the two steps? Or should this part be its own "combined" building block?
> 
> Maybe some of this state could be moved to the GeoSet? I don't know, really.
> 
> Yes: I agree that this error handling is probably not enough or complete or anything. For example `MGeo->create()` also has a return value, which is just ignored.
> 
> As originally said, I wanted to first refactor "what's there" and then improve from there.

In the example I made it would throw.

> ```c++
>     Bool_t rc = GeoInterface->readSet(MGeo);
>     if (rc) {
>         MGeo->create(loader->getGeoBuilder());
>     }
> ```

This is broken, is it not?

dennisklein:
Even another variant for inspiration:

```cpp
void MyCave::ConstructGeometry()
{
    auto& cave = GetGeometryLoader().Load<MyGeoCave>(GetGeometryFileName());
    cave.ConstructGeometry();
}
```

ChristianTackeGSI:
> Even another variant for inspiration:
> 
> ```c++
> void MyCave::ConstructGeometry()
> {
>     auto& cave = GetGeometryLoader().Load<MyGeoCave>(GetGeometryFileName());
>     cave.ConstructGeometry();
> }
> ```

What choice of the user is `cave.ConstructGeometry();` representing? "To build or not to build" ;-) In that case I would vote for moving it into the previous line. ;-)

I liked the explicit GeoBuilder reference. … also it is already included in the GeoLoader…

> 
> edit: also it should be `MyCaveGeo`, no?

Naming, naming. You know it's a mess. ;-)

> This is broken, is it not?

As always… Trying to refactor it, before we fix it. So that we only need to fix it in one place.

---

Yet another inspiration…

```cpp
void MyCave::ConstructGeometry()
{
    auto cave_on_heap = std::make_unique<MyCaveGeoSet>("Cave17");
    auto& cave = GetGeometryLoader().LoadAndConstruct(std::move(cave_on_heap),
                                                      GetGeometryFileName());
}
```

dennisklein:
```cpp
void MyCave::ConstructGeometry()
{
    GetGeometryLoader().LoadAndConstruct<MyCaveGeo>(GetGeometryFileName(), "Cave17");
}
```

btw, https://github.com/FairRootGroup/FairRoot/blob/99be5f48d4bfed7ee7179524c07bc49a6b37a5c5/fairroot/geobase/FairGeoSet.h#L113

So, a *loader* consists of an *interface* (the actual thing that performs the *loading*) and a *builder*. The *builder*, however, is used as part of a *create* action. All this is done to *construct* a geometry. Let's add *init* somewhere and we have used all possible synonyms. And all of this happens outside of the *constructor* the language provides for initialization of an object :confused:

ChristianTackeGSI:
So we're at

```cpp
GetGeometryLoader().LoadAndConstruct<T>(TString, args...)
// vs.
GetGeometryLoader().ReadAndCreate<T>(TString, args...)
```

I don't care much.

ChristianTackeGSI:
So, now we have

```cpp
GetGeometryLoader().LoadAndConstruct<T>(const char *, args…);
```

What do you think?

dennisklein (approved):

Issue #1541, opened 2024-05-27
Error message for empty container in FairRunTimeDB


TobiasStockmanns:
FairRunTimeDB generates an error message

```
-I- FairRunTimeDB::InitContainer() PndEmcGeoPar
[ERROR] init() PndEmcGeoPar not initialized
```

when a parameter container is not filled with existing data from one of the two input possibilities (either an ASCII or root par file).

In some cases these parameters are generated inside the program code and then stored into the runtime DB creating this error message even though it is not an error.

The desired behaviour would be that the message is marked as [INFO] and not as [ERROR].
The text could look like: `[INFO] init() PndEmcGeoPar does not contain any data.`

ChristianTackeGSI:
After having thought about option (1) above, I consider it quite useful.

There are some options:

```cpp
struct PndEmcGeoPar : public FairParGenericSet {
  bool do_not_call_init_at_all{false};  //!
  bool ignore_errors_from_init{false};  //!

  bool init() override {
    if (do_not_call_init_at_all) {
      return true;
    }
    return FairParGenericSet::init();
// or:
    bool retval = FairParGenericSet::init();
    if (ignore_errors_from_init && !retval) {
      LOG(info) << "The previous error is just informational, please ignore\n"
      return true;
    }
    return retval;
  }
}
```

fuhlig1:
@karabowi,

could you please have a look at the issue since you already discussed with @TobiasStockmanns .

karabowi:
The issue has been solved in PR #1551.
@TobiasStockmanns , can you check the solution?

ChristianTackeGSI:
#1551 improves the situation by moving the error out of `init()` to the caller of `init()`. So if it returns true (by overriding it), then no error will be printed. Maybe a warning is printed for `init(io)` return values.

The above mentioned override variants should still work on older versions, just their output might be a bit confusing (that's why I put the "The previous error is just informational, please ignore" there).

PR #1551, opened 2024-06-06
refactor(parbase): Improve logging in FairParSet

ChristianTackeGSI:
And re-arrange some if statements

See-Also: #1541

Included one commit from #1552

---

Checklist:

### DIFF
--- a/fairroot/parbase/FairParSet.cxx
+++ b/fairroot/parbase/FairParSet.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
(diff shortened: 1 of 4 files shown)


fairroot/parbase/FairParSet.cxx
```diff
@@ -47,30 +47,34 @@ Bool_t FairParSet::init()
...
 
     Bool_t allFound = kFALSE;
-    FairParIo* io = 0;
-    if (rtdb) {
-        io = rtdb->getFirstInput();
+    FairParIo* io = rtdb->getFirstInput();
+    if (io) {
+        allFound = init(io);
+    } else {
+        LOG(warn) << "FairParSet::init()/" << GetName() << ": No first input";
```
karabowi, line 60: This warning (just as the second one) is unnecessary and would be too verbose.
The users may (from what I recall) use `First` or `Second` input independently.

ChristianTackeGSI: Okay, I'll turn it into a `debug`, okay?

ChristianTackeGSI: I factored the "which input is there and not" into the FairRuntimeDb::initContainers caller, which now lists the available inputs.

ChristianTackeGSI:
@karabowi I included one of your commits from #1552.

I have also added even more `LOG(info)` instead of `cout`.

A current, typical output can be seen for example here:
https://cdash.gsi.de/testDetails.php?test=20858392&build=463833

karabowi (approved):

PR #1552, opened 2024-06-06
 Add `setCreationMode` to `FairParSet`

karabowi:
In rare cases, when parameteres deriving from `FairParSet`
are not present in the input parameter file and are meant
to be created during the run, `FairParSet::init()` throws
an error message.
    
By using the newly introduced setter `setCreationMode`
this message is suppressed.

This allows solution to issue #1541.

The PR also does some clean-up in `FairParSet`.

---

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## 19.0.1 (unreleased)
 
+### Deprecations
+
+If you think you really require a deprecated API, please
+[file an issue](https://github.com/FairRootGroup/FairRoot/issues/new).
+
+* Deprecated `FairParSet::fill()` and `FairParSet::store()` while these methods are not being used anywhere in `FairRoot`.
+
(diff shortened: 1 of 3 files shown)


ChristianTackeGSI:
Can you extract the chore / deprecate part into its own PR? I like those and would like to get them finished and merged first.

ChristianTackeGSI:
After having thought about this whole thing a bit (see
<https://github.com/FairRootGroup/FairRoot/issues/1541#issuecomment-2158966426>), I doubt that we actually should have yet another knob. (Yes, I proposed the knob thing, please excuse!) Not to mention, the "creation mode" as a bool sounds a bit strange. What is a truthy creation mode? What is a falsy creation mode?

And I really think by now, that the `virtual init()` is the customization point that users should use to handle all their cases about this part. We can try to make this even nicer / easier. All our cleanup PRs are going in that direction!

So alternative proposol (If you agrree): Repurpose this PR and drop the `feat` commit from it?

karabowi:
@ChristianTackeGSI , I also do not like this knob name, I will work on this PR.

Issue #1556, opened 2024-06-13
[BUG] FairRootManager::GetObject is broken in FairRoot19


YanzhaoW:
**Describe the bug**

The usage of `FairRootManager::GetObject` is broken after upgrading to FairRoot 19. This issue only occurs when `GetObject` is called more than once for one branch name. But such practice happens quite often as some branches (like event header) need to be used in different tasks at the same run.

This issue doesn't occur in FairRoot 18.8.

**To Reproduce**

1. Go to this [repository](https://github.com/YanzhaoW/test_repo/tree/fairroot_test).
2. Define env vairable `SIMPATH` and `FAIRROOTPATH` 
3. Compile and build project using FairRoot 18. Run executable `./write_data` and `./read_data`.
4. Compile and build project using FairRoot 19. Run executable `./write_data` and `./read_data`.
5. With FairRoot 19, outputs are all zeros. But if there exists only one task in `read_task.cpp`, then the output is correct.

**System information:**
 - OS: Debian buster
 - Compiler: gcc 13
 - Environment: FairSoft latest, FairRoot dev & 18.8

**Additional context**

1. This issue basically makes FairRoot 19 unusable for the data calibration, as event header is needed in almost every task.
2. Please do consider an alternative way of using `TTreeReader` used in `read.cpp` in the github repo.

@dennisklein @ChristianTackeGSI

dennisklein:
I can confirm the behaviour.

ChristianTackeGSI:
Just to be sure: The problem is the GetObject, right?

For example running write_data with 18.8 and then read_data with 19, would also fail?

We should write a (failing) test for FairRoot!

dennisklein:
> For example running write_data with 18.8 and then read_data with 19, would also fail?

Yes, but only if used from more than one task as in his example code. With a single read task it works across versions.

PR #1557, opened 2024-06-13
Fix FairRootManager::GetObject

YanzhaoW:
## Describe your proposal.

Fix the UB found in the issue #1556 related to `FairRootManager::GetObject`.

## Cause of the issue

The branch address of the TTree must be fixed during the data reading. In other words, if an object of type `T` needs to be read from a tree, the value `T**` must be fixed and cannot be changed during the whole reading process.  However, in the GetObject function, the variable `newentry` is a reference to the `T*` pointer stored in the `std::vector<T*>`. If the size of the vector grows beyond its capacity, the whole vector will be moved to another memory location, which means the `T**` value will be changed.

This makes sense as in my real practice from R3BRoot, I didn't get all zeros but rather direct segmentation fault.

Let `newentry` refer to the `T*` in the object of `std::map<TString, TObject*>` fixes the issue. `std::map` doesn't reallocate its entries during the runtime.

## TODO

1. `AddMemoryBranch` now returns `TObject*&`, which is ugly and asking for troubles in the future. Unfortunately, there isn't any better way to deal with this issue if we still use TTree raw API for the data reading. It must take a `TObject**` for calling `SetBranchAddress`.
2. As I have suggested from #1556, `TTreeReader` is a much better choice as we don't need to deal with `T**` situation anymore.

---

Checklist:

### DIFF
--- a/fairroot/base/steer/FairRootManager.h
+++ b/fairroot/base/steer/FairRootManager.h
@@ -23,7 +23,7 @@
 #include <string>
 #include <type_traits>   // is_pointer, remove_pointer, is_const, remove...
 #include <typeinfo>
-#include <vector>
+#include <list>
 
 class BinaryFunctor;
 class FairEventHeader;
@@ -333,7 +333,7 @@ class FairRootManager : public TObject
 
     /** current time in ns*/
     Double_t fCurrentTime;
-    std::vector<TObject*> fObj2{};   //!
+    std::list<TObject*> fObj2{};   //!
     /** A list which hold the pointer to the branch
      * and the name of the branch in memory, it contains all branches (TClonesArrays)
      * persistance and Memory only branches


ChristianTackeGSI:
Thanks for analyzing this!

(We came to an alike result).

Our quick fix would be to replace the `std::vector` by a `std::list`. This is probably the fix that we would like to deploy for 19.0.1.

YanzhaoW:
@ChristianTackeGSI 

I don't quite understand the purpose of `fObj2`. When you call `GetObject`, if the branch has not been added and does't exist in the tree, you push `nullptr` to the vector. If the branch exists, you push the same pointer twice into the vector. On the other hand, you also have `fMap` for the book-keeping of the branches.

ChristianTackeGSI:
Right, `fObj2` is a bit strange!

We would like to have a minimal fix for 19.0.1 first. And the minimal one is to use `std::list`.

For 19.1 we can then slowly work on further refactoring and improving GetObject.

And: Thanks for the reference to TTreeReader!

ChristianTackeGSI:
> Ok, Fair enough.
> 
> Should I close this PR and you create another one using `std::list`? Or I should just change it to `std::list` here?

How ever you like. :-)

@dennisklein wanted to create a proper test first anyway.

YanzhaoW:
@ChristianTackeGSI 

Is this minimum change OK for you?

ChristianTackeGSI (approved):
Yes, this looks good!

- Format check fails because the tree is not rebased to the latest dev branch. Not really important for this PR.
- [ ] I would like to wait for a (failing) check by @dennisklein before merging this one.

YanzhaoW:
There is another issue about branch IO of FairRootManger:

There are two different, independent branch registries: `fAnyBranchMap` and `fMap`. These are used by different APIs of FairRootManager:

1. `fAnyBranchMap` is used by `RegisterAny` and `InitObjectAs`.
2. `fMap` is used by `Register` and `GetObject`. 

Any branch registered in one way cannot be read by another way.

What's even more "interesting" is that the event header (online run) is always registered in `fMap` but not in `fAnyBranchMap`. So you may find that you can get a valid event header from `GetObject` but mysteriously nullptr from `InitObjectAs`.

@ChristianTackeGSI @dennisklein

ChristianTackeGSI:
It looks like it will take a bit longer to create proper tests for this. So merging now, so that the fix is in place at least.

ChristianTackeGSI:
BTW: I think, we need a ChangeLog entry for this?

PR #1558, opened 2024-06-14
[v18.8] Export pkg deps

dennisklein:


ChristianTackeGSI (commented):
Should we have something in the ChangeLog?

ChristianTackeGSI (approved):

ChristianTackeGSI (approved):

PR #1564, opened 2024-06-17
refactor: Avoid FairGeoLoader::Instance()

ChristianTackeGSI:
Instead add a few GetGeometryLoader() getters and use them. Also let FairModule know to which FairRunSim it belongs.

---

Checklist:

### DIFF
--- a/examples/common/passive/FairCave.cxx
+++ b/examples/common/passive/FairCave.cxx
@@ -24,14 +24,14 @@
 
 void FairCave::ConstructGeometry()
 {
-    FairGeoLoader* loader = FairGeoLoader::Instance();
-    FairGeoInterface* GeoInterface = loader->getGeoInterface();
+    FairGeoLoader& loader = GetGeometryLoader();
+    FairGeoInterface* GeoInterface = loader.getGeoInterface();
     FairGeoCave* MGeo = new FairGeoCave();
     MGeo->setGeomFile(GetGeometryFileName());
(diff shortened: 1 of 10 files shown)


dennisklein (commented):
What's your idea behind the `friend class FairRunSim` + direct member access vs just a public setter (and/or ctor overload)?

ChristianTackeGSI:
> What's your idea behind the `friend class FairRunSim` + direct member access vs just a public setter (and/or ctor overload)?

My general idea: Setting fRunSim is (currently) not intended as a public inteface.

> Thinking also about unit testability.

Good point. It would be good for tests to be able to set a moch RunSim there…

So, I am going to add a member function with a "for internal use" description.

dennisklein (commented):
The remaining thought I have, is, if the association of a `FairModule` to a `FairRunSim` is general enough. In other words, are `FairModule`s only used in simulation and never in other run types?

ChristianTackeGSI:
> The remaining thought I have, is, if the association of a `FairModule` to a `FairRunSim` is general enough. In other words, are `FairModule`s only used in simulation and never in other run types?

@karabowi, @fuhlig1: Can you please comment on this?

dennisklein (approved):
Chatted with @karabowi about this and we concluded this is fine. Does not break any API anyhow and if we discover this needs to be more generic, we can then still move it to `FairRun` or somewhere else.

PR #1566, opened 2024-06-24
fix(examples): More unique output filenames

ChristianTackeGSI:
As our tests run in parallel, the output filenames must be unique across different run types. So add more properties to the filename (MT, postinit).

---

Checklist:


ChristianTackeGSI:
Here's a link to a run, where this problem can be looked at:

https://cdash.gsi.de/testDetails.php?test=20817203&build=463322

dennisklein (approved):

PR #1570, opened 2024-06-28
Disable streamer for FairRootManager and fix unconditional dereference in `SetSink()`

dennisklein:
@ChristianTackeGSI Here are some bits of the FairRootManager work I am currently preparing as separate PR as requested.

### DIFF
--- a/fairroot/base/LinkDef.h
+++ b/fairroot/base/LinkDef.h
@@ -35,7 +35,7 @@
 #pragma link C++ class FairParticle+;
 #pragma link C++ class FairPrimaryGenerator+;
 #pragma link C++ class FairRecoEventHeader+;
-#pragma link C++ class FairRootManager+;
+#pragma link C++ class FairRootManager;
 #pragma link C++ class FairRun+;
 #pragma link C++ class FairRunAna;
 #pragma link C++ class FairRunIdGenerator;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to FairRoot will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
+## x.x.0 - xxxx-xx-xx
+
+### Breaking Changes
+
+* No streamer is generated for `FairRootManager` any more
+
 ## 19.0.0 - 2024-05-17
 
 ### Breaking Changes
--- a/fairroot/base/steer/FairRootManager.h
+++ b/fairroot/base/steer/FairRootManager.h
@@ -395,7 +395,7 @@ class FairRootManager : public TObject
     // data members
     Int_t fId{0};   //! This manager ID
 
-    ClassDefOverride(FairRootManager, 14);
+    ClassDefOverride(FairRootManager, 0);
 };
 
 // FIXME: move to source since we can make it non-template dependent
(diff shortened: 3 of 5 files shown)


fairroot/base/LinkDef.h
```diff
@@ -35,7 +35,7 @@
 #pragma link C++ class FairParticle+;
 #pragma link C++ class FairPrimaryGenerator+;
 #pragma link C++ class FairRecoEventHeader+;
-#pragma link C++ class FairRootManager+;
+#pragma link C++ class FairRootManager;
```
fuhlig1, line 38: If I am correct this will still create the streamer but an older and not IO optimised form. To not create the streamer a `-` shold be added to the class name. Unfortunately I can't find the proper documentation. I only found 
https://root.cern/manual/io_custom_classes/
which doesn't describe the feature.
I will try to do some tests to clarify this.

ChristianTackeGSI: > Note
> 
> The `+` after the class name enables extra features and performance improvements in the I/O of the type.

Okay, that sounds like what Florian is looking at.

> Remember to always add it to your linkdef directives.

Reading that I would think, that we should just keep it for the time being.

And as I understand it, the `0` version in the ClassDefOverride already disables the streamer.

ChristianTackeGSI (approved):
I don't care much about the `+`.

LGTM for the remaining stuff.

fairroot/base/LinkDef.h
dennisklein: hm, the streamer for the TObject base class remains. It looks to me as if one cannot use `-`-postfix on classes that inherit from `TObject`?

dennisklein: `ClassDefOverride(FairRootManager, 0)` +

1. `#pragma link C++ class FairRootManager-;`
   ```
   /usr/bin/ld: fairroot/base/CMakeFiles/Base.dir/steer/FairRootManager.cxx.o:(.data.rel.ro._ZTV15FairRootManager[_ZTV15FairRootManager]+0x1e0): undefined reference to 'FairRootManager::Streamer(TBuffer&)'
   /usr/bin/ld: fairroot/base/CMakeFiles/Base.dir/G__Base.cxx.o: in function 'ROOT::streamer_FairRootManager(TBuffer&, void*)':
   /home/dklein/projects/FairRoot3/build/fairroot/base/G__Base.cxx:5486:(.text+0x171bc): undefined reference to `FairRootManager::Streamer(TBuffer&)'
   ```

2. `#pragma link C++ class FairRootManager;`
   
   ```cpp
   void FairRootManager::Streamer(TBuffer &R__b)
   {
      // Stream an object of class FairRootManager.
   
      TObject::Streamer(R__b);
   }
   ```

3. `#pragma link C++ class FairRootManager+;`

   ```cpp
   void FairRootManager::Streamer(TBuffer &R__b)
   {
      // Stream an object of class FairRootManager.
   
      if (R__b.IsReading()) {
         R__b.ReadClassBuffer(FairRootManager::Class(),this);
      } else {
         R__b.WriteClassBuffer(FairRootManager::Class(),this);
      }
   }
   ```

dennisklein: @fuhlig1 Let me know, if I should put the `+` back.

PR #1579, opened 2024-07-21
New way to color tracks.

DanielWielanek:
Currently the tracks coloring depens on node MCTrack colors, and the color scheme is configured by FairEventManager.
In this commit I propose the following changes:
1. FairEventManager only provide the pointer to XML node with configuration
2. In examples there are different nodes for coloring of Geo and MC tracks (thank to the new policy)
3. Old coloring shame is marked as depreacted
4. The XML config for Event Display was updated (at some point the names of the last layers of examples was changed - the configuration was not working for last 4 stations of detector)

---

Checklist:

### DIFF
--- a/fairroot/eventdisplay/FairEventManager.cxx
+++ b/fairroot/eventdisplay/FairEventManager.cxx
@@ -15,6 +15,7 @@
 #include "FairRootManager.h"   // for FairRootManager
 #include "FairRunAna.h"        // for FairRunAna
 #include "FairXMLNode.h"
+#include "xml/FairXMLDetectorColor.h"
 
 #include <TDatabasePDG.h>   // for TDatabasePDG
 #include <TEveBrowser.h>
@@ -64,7 +65,6 @@ FairEventManager::FairEventManager()
     , fRhoZPlane{-1, 0, 0, 0}
     , fRphiCam(TGLViewer::kCameraOrthoXOY)
     , fRhoCam(TGLViewer::kCameraOrthoZOY)
-    , fXMLConfig("")
 {
     fgRinstance = this;
     AddParticlesToPdgDataBase();
@@ -84,7 +84,7 @@ void FairEventManager::Init(Int_t visopt, Int_t vislvl, Int_t maxvisnds)
         fWorldSizeY = box->GetDY();
         fWorldSizeZ = box->GetDZ();
     }
-    if (!fXMLConfig.EqualTo(""))
+    if (fXMLFile)
         LoadXMLSettings();
     gEve->AddGlobalElement(TNod);
     gEve->FullRedraw3D(kTRUE);
@@ -174,6 +174,11 @@ FairEventManager::~FairEventManager()
 
 void FairEventManager::Open() {}
 
+void FairEventManager::SetXMLConfig(TString xml_config)
+{
+    if (xml_config.Length() && xml_config.EndsWith(".xml"))
+        fXMLFile.reset(new FairXMLFile(xml_config));
+}
 void FairEventManager::GotoEvent(Int_t event)
 {
     fEntry = event;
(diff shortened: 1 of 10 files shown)


fairroot/eventdisplay/FairEventManager.cxx
```diff
@@ -174,6 +174,11 @@ FairEventManager::~FairEventManager()
 
 void FairEventManager::Open() {}
 
+void FairEventManager::SetXMLConfig(TString xml_config)
+{
+    if (xml_config.Length() && xml_config.EndsWith(".xml"))
+        fXMLFile.reset(new FairXMLFile(xml_config));
```
ChristianTackeGSI, line 180: Please avoid `new` when possible.

and why force a `.xml` ending? This will be confusing to users, when they call `SetXMLConfig` and just nothing happens, just because they call their files `.XML`?

```suggestion
    if (xml_config.Length()) {
        fXMLFile = std::make_unique<FairXMLFile>(xml_config);
    }
```

fairroot/eventdisplay/FairEventManager.h
```diff
@@ -151,7 +156,6 @@ class FairEventManager : public TEveEventManager
     TEveProjectionAxes* GetRPhiAxes() const { return fAxesPhi; };
     TEveProjectionAxes* GetRhoZAxes() const { return fAxesRho; };
     virtual void LoadXMLSettings();
-    void LoadXMLDetector(TGeoNode* node, FairXMLNode* xml, Int_t depth = 0);
```
ChristianTackeGSI, line 154: This is a public member function. So you're removing public API.

This is a breaking change.

If you really want this breaking change, then please write an appropriate CHANGELOG entry.

fairroot/eventdisplay/xml/FairXMLDetectorColor.cxx
```diff
@@ -0,0 +1,93 @@
+/*
```
ChristianTackeGSI, line 1: This one misses the usual copyright header.

fairroot/eventdisplay/xml/FairXMLDetectorColor.cxx
```diff
@@ -0,0 +1,93 @@
...
+#include <TNamed.h>
+#include <TString.h>
+Int_t FairXMLDetectorColor::ApplyToNode(TGeoNode* node, FairXMLNode* xml) const
+{
+    auto transparency_atrib = xml->GetAttrib("transparency");
+    Char_t transparency = 0;
+    if (transparency_atrib) {
+        if (transparency_atrib->GetValue().Length()) {
+            transparency = transparency_atrib->GetValue().Atoi();
+            node->GetVolume()->SetTransparency(transparency);
```
ChristianTackeGSI, line 25: You expect `node` and `xml` to be non-nullptr.

Why not use a reference instead?

```suggestion
Int_t FairXMLDetectorColor::ApplyToNode(TGeoNode& node, FairXMLNode& xml) const
{
    auto transparency_atrib = xml.GetAttrib("transparency");
    Char_t transparency = 0;
    if (transparency_atrib) {
        if (transparency_atrib->GetValue().Length()) {
            transparency = transparency_atrib->GetValue().Atoi();
            node.GetVolume()->SetTransparency(transparency);
```

(I've only rewritten the start as an example.)

ChristianTackeGSI (changes requested):
Please take a look at our [Coding guidelines](https://github.com/FairRootGroup/FairRoot/blob/dev/CONTRIBUTING.md#coding-guidelines).

Especially G.3, G.4, and G.5.

I've added a few comments around. Those are just random.

fairroot/eventdisplay/FairEventManager.h
DanielWielanek: I'm not sure if I made change corretly, please check :)

PR #1584, opened 2024-08-15
refactor(*Stack): Use local variables

ChristianTackeGSI:
... instead of member variable fFSTrackIter.

Deprecate that member variable.

---

Checklist:

### DIFF
--- a/examples/common/mcstack/FairStack.cxx
+++ b/examples/common/mcstack/FairStack.cxx
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2014-2023 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
+ * Copyright (C) 2014-2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH  *
  *                                                                              *
  *              This software is distributed under the terms of the             *
  *              GNU Lesser General Public Licence (LGPL) version 3,             *
@@ -239,12 +239,7 @@ void FairStack::FillTrackArray()
 
 Int_t FairStack::GetCurrentTrackNumber() const
 {
-    std::map<Int_t, Int_t>::const_iterator tempIter =
-        fFSTrackMap.find(fCurrentTrack);   // check if the track was created by FastSimulation
-    if (tempIter != fFSTrackMap.end()) {
-        return tempIter->second;
-    }
-    return fCurrentTrack;
+    return FSTrackMapLookup(fCurrentTrack);
 }
 
 void FairStack::UpdateTrackIndex(TRefArray* detList)
@@ -342,11 +337,7 @@ void FairStack::Print(Option_t*) const
 void FairStack::AddPoint(DetectorId detId)
 {
     Int_t iDet = detId;
-    Int_t iTr = fCurrentTrack;
-    fFSTrackIter = fFSTrackMap.find(iTr);      // check if this track is not already created by FastSimulation
-    if (fFSTrackIter != fFSTrackMap.end()) {   // indeed the track has been created by the FastSimulation mechanism
-        iTr = fFSTrackIter->second;            // use the ID of the original track
-    }
+    const Int_t iTr = GetCurrentTrackID();
 
     pair<Int_t, Int_t> a(iTr, iDet);
     if (fPointsMap.find(a) == fPointsMap.end()) {
@@ -362,11 +353,8 @@ void FairStack::AddPoint(DetectorId detId, Int_t iTrack)
         return;
     }
     Int_t iDet = detId;
-    Int_t iTr = iTrack;
-    fFSTrackIter = fFSTrackMap.find(iTr);      // check if this track is not already created by FastSimulation
-    if (fFSTrackIter != fFSTrackMap.end()) {   // indeed the track has been created by the FastSimulation mechanism
-        iTr = fFSTrackIter->second;            // use the ID of the original track
-    }
+    const Int_t iTr = FSTrackMapLookup(iTrack);
+
     pair<Int_t, Int_t> a(iTr, iDet);
     if (fPointsMap.find(a) == fPointsMap.end()) {
         fPointsMap[a] = 1;
@@ -395,16 +383,15 @@ TParticle* FairStack::GetParticle(Int_t trackID) const
 
 void FairStack::SelectTracks()
 {
-
     // --> Clear storage map
     fStoreMap.clear();
 
     // loop over tracks created by FastSim to order the tracks
-    for (fFSTrackIter = fFSTrackMap.begin(); fFSTrackIter != fFSTrackMap.end(); ++fFSTrackIter) {
+    for (auto tmentry : fFSTrackMap) {
         for (Int_t i = 0; i < fNParticles; i++) {
             TParticle* thisPart = GetParticle(i);
-            if (thisPart->GetMother(0) == fFSTrackIter->first)
-                thisPart->SetMother(0, fFSTrackIter->second);
+            if (thisPart->GetMother(0) == tmentry.first)
+                thisPart->SetMother(0, tmentry.second);
         }
     }
 
@@ -433,10 +420,7 @@ void FairStack::SelectTracks()
         }
 
         // check if the track was created by the FastSim mechanism. Do not store such tracks
-        Bool_t fastSimTrack = kFALSE;
-        fFSTrackIter = fFSTrackMap.find(i);
-        if (fFSTrackIter != fFSTrackMap.end())
-            fastSimTrack = kTRUE;
+        const bool fastSimTrack = (fFSTrackMap.count(i) > 0);
 
         // --> Check for cuts (store primaries in any case)
         if (iMother < 0) {
--- a/fairroot/base/sim/FairGenericStack.cxx
+++ b/fairroot/base/sim/FairGenericStack.cxx
@@ -20,6 +20,8 @@
 #include <cstring>   // strcmp
 #include <fairlogger/Logger.h>
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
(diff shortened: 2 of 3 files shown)


examples/common/mcstack/FairStack.cxx
```diff
@@ -239,8 +239,7 @@ void FairStack::FillTrackArray()
 
 Int_t FairStack::GetCurrentTrackNumber() const
 {
-    std::map<Int_t, Int_t>::const_iterator tempIter =
-        fFSTrackMap.find(fCurrentTrack);   // check if the track was created by FastSimulation
+    auto tempIter = fFSTrackMap.find(fCurrentTrack);   // check if the track was created by FastSimulation
     if (tempIter != fFSTrackMap.end()) {
         return tempIter->second;
     }
```
dennisklein, line 245:
```suggestion
    // check if the track was created by FastSimulation
    if (auto it = fFSTrackMap.find(fCurrentTrack); it != fFSTrackMap.end()) {
        return it->second; 
    }
```

Limit the variable scope even further? see Examples @ https://en.cppreference.com/w/cpp/container/map/find

ChristianTackeGSI: In general, I like limiting variable scopes!

Just we're back to the old problem of the convoluted if statement. I'd like the assignment and the check to be on different lines inside the `if`. And last time we checked, this involved turning of the clang-format. Not exactly what I like.

If the above if-statement-style is wanted, I am of course happy to change it. :-)

examples/common/mcstack/FairStack.cxx
```diff
@@ -395,16 +394,15 @@ TParticle* FairStack::GetParticle(Int_t trackID) const
 
 void FairStack::SelectTracks()
 {
-
     // --> Clear storage map
     fStoreMap.clear();
 
     // loop over tracks created by FastSim to order the tracks
-    for (fFSTrackIter = fFSTrackMap.begin(); fFSTrackIter != fFSTrackMap.end(); ++fFSTrackIter) {
+    for (auto tmentry : fFSTrackMap) {
```
dennisklein, line 401:
```suggestion
    for (const auto [newTrack, originalTrack] : fFSTrackMap) {
```

see https://en.cppreference.com/w/cpp/language/structured_binding

Perhaps you find better names even, `.first`/`.second` is definitely terrible for readability.

fairroot/base/sim/FairGenericStack.cxx
```diff
@@ -121,9 +131,9 @@ void FairGenericStack::FastSimMoveParticleTo(Double_t xx,
     PushTrack(tobedone, parent, pdg, px, py, pz, en, xx, yy, zz, tt, polx, poly, polz, proc, ntr, weight, status, -1);
     fFSMovedIndex = GetListOfParticles()->GetEntries() - 1;
     Int_t trackID = TVirtualMC::GetMC()->GetStack()->GetCurrentTrackNumber();
-    fFSTrackIter = fFSTrackMap.find(trackID);   // check if this track is not already created by FastSimulation
-    if (fFSTrackIter != fFSTrackMap.end())      // indeed the track has been created by the FastSimulation mechanism
-        trackID = fFSTrackIter->second;         // use the ID of the original track
+    auto tmentry = fFSTrackMap.find(trackID);   // check if this track is not already created by FastSimulation
+    if (tmentry != fFSTrackMap.end())           // indeed the track has been created by the FastSimulation mechanism
+        trackID = tmentry->second;              // use the ID of the original track
     fFSTrackMap[ntr] = trackID;
```
dennisklein, line 137: This looks like some oversight from when `fFSTrackMap` was moved to `FairGenericStack`, no? `FairStack::GetCurrentTrackNumber()` also implements the FS-map-lookup. Should we rather delete it here and move the FS-map-lookup to `FairGenericStack::GetCurrentTrackNumber()` and remove the override `FairStack::GetCurrentTrackNumber()`?

examples/common/mcstack/FairStack.cxx
dennisklein: as discussed on mm: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#bracewrapping with `BWACS_MultiLine` should do the trick?

examples/common/mcstack/FairStack.cxx
ChristianTackeGSI: TL;DR *I* wont do this in this PR.

To Be Honest, I have not yet udnerstood what this map is actually mapping. Yes, it's something "old" / "new". But I really don't understand it yet. So if someone else has this understanding, then I'd ask that person to go ahead and improve things!

fairroot/base/sim/FairGenericStack.cxx
ChristianTackeGSI: I guess, a whole lot more `fFSTrackMap` related things should probably be moved from FairStack to FairGenericStack.

I guess, the first one to move over from FairStack to FairGenericStack would be `fCurrentTrack`.

TBH: The question is basically, whether FairGenericStack should already have a "a good default implementation" of many things, or whether it should just be an API that people have to implement. For the later we could provide some FairMostlyUsableStack with some "good default implementation".

I think, we should leave that question for Radek and Florian as well.

dennisklein: The current code does the FS-map-lookup twice, does it not?

ChristianTackeGSI: I think, it does, yes!

But I don't understand this, really.

fairroot/base/sim/FairGenericStack.cxx
```diff
@@ -121,9 +131,9 @@ void FairGenericStack::FastSimMoveParticleTo(Double_t xx,
     PushTrack(tobedone, parent, pdg, px, py, pz, en, xx, yy, zz, tt, polx, poly, polz, proc, ntr, weight, status, -1);
     fFSMovedIndex = GetListOfParticles()->GetEntries() - 1;
     Int_t trackID = TVirtualMC::GetMC()->GetStack()->GetCurrentTrackNumber();
```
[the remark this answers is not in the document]
ChristianTackeGSI: I'll put this in a dedicated PR, so that our experts can take a deeper look.

examples/common/mcstack/FairStack.cxx
ChristianTackeGSI: As #1586 was merged, I have rebased this PR and updated most places looking like this one.

examples/common/mcstack/FairStack.cxx
dennisklein: This comment I took the terms "new" and "original" from:

https://github.com/FairRootGroup/FairRoot/blob/ef00300f8c7b3394619e84b7f1cae98d29133509/base/sim/FairGenericStack.h#L146-L147

dennisklein: But I agree, it does not help much.

ChristianTackeGSI: Can we resolve this and postpone for the next refactoring round?

fairroot/base/sim/FairGenericStack.cxx
ChristianTackeGSI: I have refactored the trackmap lookup into a new member function. This and #1588 show the double applying of that member function quite well. I guess, we can then file a nice issue.

examples/common/mcstack/FairStack.cxx
dennisklein: I would rather just ask someone?
@MohammadAlTurany @karabowi @fuhlig1 @kresan Can you help here and shed some light on the purpose of `FairGenericStack::fFSTrackMap`, thx! The goal is to find readable names for variables denoting *what* this map maps onto *what*.

karabowi: From what I remember this map was introduced when we (@karabowi and @kresan) introduced the possibility to run `FastSimulation` in FairRoot. From what I remember the problem was the particle counting in the `FastSimulation` mode: normally, the secondaries created in the transport engine go to the end of the stack, while in the `FS` mode they were pushed just after the primary. That caused problem later on, so a map was created to track the `FS` secondaries.
After the event, the `FS` secondaries were to be moved to the end of the stack, and the points created by the had to be changed.

All that said. @fuhlig1 , @kresan , I think the `FastSim` mode was a dead end. `PANDA` is not using our implementation, they have their own. What about CBM or R3B? IMHO we could get rid of `FastSim` totally.
What do you think?

ChristianTackeGSI: > IMHO we could get rid of `FastSim` totally. What do you think?

Should we move that part of the discussion to a fresh issue?

ChristianTackeGSI: TBH, I don't know how to continue with this PR: If FastSim is a "dead end", then I probably should not invent new APIs like `FairGenericStack::FSTrackMapLookup` (which I currently do in this PR)?

But on the other hand, I am not feeling happy to just strip all FastSim stuff away.

If we wanted to merge this one "to simplify our current code", then I could already deprecated `FairGenericStack::FSTrackMapLookup`?

karabowi: Maybe we can discuss the issue on our meeting next Monday. Dima is pretty sure it is not used in `R3B`, `cbmroot` also has no traces of `FastSim`.

ChristianTackeGSI: Rethinking this, I would propose the following:

- We consider this PR independant of a possible FastSim deprecation / removal.
- Refactoring this might make the FastSim deprecation / removal even simpler, maybe
- And in a second step let's look at the FastSim stuff (please let's open an issue!)

Considering a possible deprecation / removal of these Fastsim things, I'd propose to keep the non-intuitive names for the time being.

karabowi (approved):

PR #1586, opened 2024-08-16
clang-format: Put braces on own line in special cases

ChristianTackeGSI:
Multi-line control statements with the opening brace on the last line can be a little confusing to read:

```cpp
if (auto foo = something;
    foo.extra_data.check_for(17)) {
    foo.extra_data.check_and_act(19);
}
```

It is not easy to see the end of the condition and the start of the "then" block.

This gets much easier when putting the opening brace on its own line:

```cpp
if (auto foo = something;
    foo.extra_data.check_for(17))
{
    foo.extra_data.check_and_act(19);
}
```

To show the difference, a few files have been reformatted with this.

---

Checklist:

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -26,6 +26,7 @@ BraceWrapping:
   SplitEmptyFunction: false
   SplitEmptyRecord: false
   SplitEmptyNamespace: false
+  AfterControlStatement: MultiLine
 BreakConstructorInitializers: BeforeComma
 ColumnLimit: 120
 ConstructorInitializerIndentWidth: 4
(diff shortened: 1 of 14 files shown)


dennisklein:
```
CMake Error (dev) at /Users/alfaci/install/FairSoft/share/faircmakemodules/modules/FairFindPackage2.cmake:214 (find_package):
  Policy CMP0167 is not set: The FindBoost module is removed.  Run "cmake
  --help-policy CMP0167" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

Call Stack (most recent call first):
  CMakeLists.txt:237 (find_package2)
This error is for project developers. Use -Wno-error=dev to suppress it.
```

dennisklein:
To fix we need to bump the version here https://github.com/FairRootGroup/FairRoot/blob/84175d0291bbcd09b70e1626f7957d7bcacc6ce4/CMakeLists.txt#L9

ChristianTackeGSI:
> To fix we need to update the version range here
> 
> https://github.com/FairRootGroup/FairRoot/blob/84175d0291bbcd09b70e1626f7957d7bcacc6ce4/CMakeLists.txt#L9

Can you create a PR for this?

Should we backport this for 18.x then as well?

Issue #1595, opened 2024-11-11
Duplicate entries if FairVolumeList


TobiasStockmanns:
**Describe the bug**
When running a simulation macro with PandaRoot a number of errors is created, all of the type:
```
[ERROR] FairVolumeList element: PixelActiveo5 VolId : 6032 already defined 6030
[ERROR] FairVolumeList element: PixelActiveo5oPartAss VolId : 6032 already defined 6031
[ERROR] FairVolumeList element: PixelActiveo5 VolId : 6032 already defined 6030
[ERROR] FairVolumeList element: PixelActiveo5oPartAss VolId : 6032 already defined 6031
```
The error shows up in the recent dev branch of FairRoot and is not present in v18.8.2

**To Reproduce**
1. FairSoft dev, FairRoot dev
2. PandaRoot dev
3. run /macro/master/sim_complete.C

ChristianTackeGSI:
First: I agree with Dennis.

Note: I think, we added the error message there, because the new implementation effectively (using `unique_ptr`) `delete`s the passed in FairVolume. So this was some sort of "You gave me a FairVolume, well, I `delete`d it. Maybe that's a real problem?".

For me it is still not clear, whether this is "expected" behaviour? If so, IMHO the caller that expects this behaviour should at least (using comments) document that this is "okay".

All that said:

1. Is this error happening anywhere in the FairRoot "testsuite" (vulgo examples)?

   If yes: Good, then we have some "expected" happenings and should address them (by documentation).

   If not: See next point

2. (If not) What is PandaRoot performing differently? If PandaRoot's usage of FairRoot is fine, then we maybe should add another example that actually utilized this behaviour?

dennisklein:
> we added the error message

If you mean recently, no, we did not add it. It was already there for decades.

ChristianTackeGSI:
> > we added the error message
> 
> If you mean recently, no, we did not add it. It was already there for decades.

Ohhh, okay.

Then: With the move to unique_ptr it became even more important?

dennisklein:
No, why?

ChristianTackeGSI:
Before 7f0f7f185cbef80298aeedb368631d638c1782d6 the passed volume was not destructed (and leaked). So any user of `addVolume` could still continue using the FairVolume.

With the destruction the error message became a little more important.

[...]

PR #1596, opened 2024-11-22
fix(sim): Add check for multiple sensitive volumes with same name

karabowi:
Both the transport engines and FairRoot allow registering of multiple nodes with the same volume/same volume name (copy mechanism). However currently such workflow logs errors from `FairVolumeList::addVolume()`.

The commit reintroduces the check for same volume names into the `FairModule::AddSensitiveVolume()` function thus preventing registration of copy volumes.

Fixes the issue #1595.

---

Checklist:


fairroot/base/sim/FairModule.cxx
```diff
@@ -280,12 +280,14 @@ void FairModule::AddSensitiveVolume(TGeoVolume* vol)
     auto volName = vol->GetName();
     LOG(debug2) << "AddSensitiveVolume " << volName;
 
-    auto addedVol = vList->addVolume(std::make_unique<FairVolume>(volName, fNbOfVolumes));
-    if (!addedVol) {
-        return;
+    if (!vList->findObject(volName)) {
```
dennisklein, line 283: Sry, no. This check was just recently removed because it is already performed by `addVolume()` and thus was performed twice in the past which you now propose to revert to.

`addVolume()` now returns the result of the `findObject()` check so a caller knows whether the volume was added. `vList` effectively resembles the semantics of a [*Set*](https://en.wikipedia.org/wiki/Set_(abstract_data_type)).

dennisklein (changes requested):
Strong veto.

fairroot/base/sim/FairModule.cxx
ChristianTackeGSI: I fully agree. This only hides the error message.

If you just want to hide the error message, then
1. Just remove it in FairVolumeList::addVolume and make sure that all callers have proper logging (maybe at debug level).
2. Or give addVolume an optional quiet option like this `addVolume(… volume, bool quiet=False)` and call it with `, true)` if you want to calm it.

I'd rather prefer (1) instead of (2).

karabowi: I have changed the commit according to (1):
`LOG(error)` was replaced with `LOG(debug)`

ChristianTackeGSI:
> Change log level for multiple sensitive volumes with same name
> 
> Both the transport engines and FairRoot allow registering of multiple nodes with the same volume/same volume name (copy mechanism). However currently such workflow logs errors from `FairVolumeList::addVolume()`.

Note: All current users of `addVolume` only call `RegisterSensitiveVolume` IF the addVolume succeeds. So "registering" a *sensitive* volume with the same name does not seem to work this way.

If the current behaviour is okay, then please rephrase your commit message.

karabowi:
I have changed the behavior of the check.
If it detects same name volumes in one detector, `FairModule::AddSensitiveVolume()` will spit out `LOG(debug)`
as we assume it is a standard copy mechanism procedure.
In case if the code tries to add sensitive volume for different detector, a `LOG(fatal)` will be issued and the execution will terminate.

fairroot/base/sim/FairModule.cxx
```diff
@@ -280,8 +280,9 @@ void FairModule::AddSensitiveVolume(TGeoVolume* vol)
     auto volName = vol->GetName();
     LOG(debug2) << "AddSensitiveVolume " << volName;
 
-    auto addedVol = vList->addVolume(std::make_unique<FairVolume>(volName, fNbOfVolumes));
+    auto addedVol = vList->addVolume(std::make_unique<FairVolume>(volName, fNbOfVolumes, 0, this));
```
ChristianTackeGSI, line 283: 1. Please improve the other calls to `vList->addVolume` as well. So that they also have the proper values set.

2. If you call the extended ctor, then plasse set the modid correctly as well.

   ```suggestion
       auto addedVol = vList->addVolume(std::make_unique<FairVolume>(volName, fNbOfVolumes, fModId, this));
   ```

3. If you have done this, consider dropping the first two lines of `FairModule::RegisterSensitiveVolume`.

ChristianTackeGSI (commented):
Just some ideas.

fairroot/base/sim/FairVolumeList.cxx
```diff
@@ -34,13 +36,20 @@ FairVolume* FairVolumeList::addVolume(std::unique_ptr<FairVolume> vol)
 {
     auto vol_found = findObject(vol->GetName());
 
+    auto vol_added = static_cast<FairVolume*>(vol.get());
```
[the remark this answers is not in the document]
karabowi: You are right!

fairroot/base/sim/FairVolumeList.cxx
```diff
@@ -34,13 +36,20 @@ FairVolume* FairVolumeList::addVolume(std::unique_ptr<FairVolume> vol)
...
+
+    // FATAL: The same volume name for different detectors
+    if (vol_found && vol->GetDetector() != vol_found->GetDetector() ) {
+        LOG(fatal) << "FairVolumeList Trying to register element: " << vol->GetName() << " (VolId=" << vol->getVolumeId()
+                   << ") for detector " << vol->GetDetector()->GetName() << ", but it was already defined (VolId="
+                   << vol_found->getVolumeId() << ") for detector " << vol_found->GetDetector()->GetName();
+        std::terminate();
+        return nullptr;
+    }        
     if (vol_found) {
```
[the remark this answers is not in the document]
karabowi: At least on my mac LOG(fatal) doesn't crash. Changed!

fairroot/base/sim/FairModule.cxx
karabowi: Did it!

ChristianTackeGSI (commented):
Looks good to me now.

Maybe one could move the LOG(debug) as an else branch of the LOG(fatal), but really, I don't care much.

Please take a look at the clang-format messages from the CI.

@dennisklein might have some ideas?

fairroot/base/sim/FairModule.cxx
karabowi: @dennisklein , does the PR look good to you now?

karabowi:
@ChristianTackeGSI @dennisklein Can we merge this PR?

PR #1597, opened 2024-11-27
Update Mac CI nodes

rbx:


rbx:
Initial builds failed because this warning from VGM and Geant4 was treated as error:

>   Compatibility with CMake < 3.10 will be removed from a future version of CMake.

They both ask for a minimum CMake below 3.10 and the new CMake 3.31 on mac does not support those, thus producing this warning.

Everything compiles if cmake warnings are not treated as errors. As a test I removed `-Werror=dev` and then it compiles. What is the purpose of it? Do you want to fix the cause of the deprecation warnings in VGM and Geant4 first?

Following compilation, on MacOS 15 ex_sim_tutorial4_TGeant4 test fails in the same way on both x86 and M1:
https://cdash.gsi.de/testDetails.php?test=22342577&build=484575
in a way I don't understand.

fuhlig1:
Close. The changes were merged with #1602

PR #1602, opened 2025-03-19
Fix pythia6 dependency

fuhlig1:
With recent versions of ROOT the pythia6 support was dropped. This PR fixes the pythia6 usage in FairRoot. 

---

Checklist:


fuhlig1:
I added the two commits from #1597. Otherwise both PRs are stuck. This one can't finish since two of macosx checks will not start due to new labels and PR #1597 will fill with an error fixed in this PR.

fuhlig1:
@karabowi, @dennisklein,

please have a look at the PR. I need it and also the back ports to test the new FairSoft version. Currently we are stuck in a hen and egg problem. I can check the changes only with the upcoming FairSoft version and the upcoming FairSoft version only with this changes.
Currently we have no running CI such that we need to do manual checks before merging.

PR #1603, opened 2025-03-19
fix(build): Fix issue with cmake > 3.30

fuhlig1:
CMake 3.31 deprecates the compatibility with CMake < 3.10 which emits two warnings from Geant4 and vgm cmake macros. The usage of -Werror=dev turns the warning into an error which breaks the CI for some systems which already use cmake 3.31.

---

Checklist:

### DIFF
--- a/FairRoot_build_test.cmake
+++ b/FairRoot_build_test.cmake
@@ -30,7 +30,6 @@ ctest_start(Continuous)
 get_filename_component(test_install_prefix "${CTEST_BINARY_DIRECTORY}/install"
                        ABSOLUTE)
 list(APPEND options
-  "-Werror=dev"
   "-DDISABLE_COLOR=ON"
   "-DCMAKE_INSTALL_PREFIX:PATH=${test_install_prefix}"
 )


dennisklein:
Re-assigned to @ChristianTackeGSI as he added this option with the comment that this error class should error in our CI. So, I guess, he has a solution - other than disabling the error - in mind for the case this is triggered.

dennisklein:
Perhaps related https://gitlab.kitware.com/cmake/cmake/-/issues/26288

fuhlig1:
> Re-assigned to @ChristianTackeGSI as he added this option with the comment that this warning class should error in our CI. So, I guess, he has a solution - other than disabling the error - in mind for the case this is triggered.

The problem is that the error doesn't come from our own project but from some of our dependencies which we can't fix.

dennisklein:
> > Re-assigned to @ChristianTackeGSI as he added this option with the comment that this warning class should error in our CI. So, I guess, he has a solution - other than disabling the error - in mind for the case this is triggered.
> 
> The problem is that the error doesn't come from our own project but from some of our dependencies which we can't fix.

Sure, but still interested what Christian had in mind in such cases. (I personally never understood how "warning:=error" type of flags shall work.)

Your PR proposes to remove the flag unconditionally, which means the original function - whatever it was supposed to be - goes away forever for all cases. Is this the intention? (In theory, you only need to disable it for cases where unsuitable Geant and VGM versions are used, right? Or better, if one could only apply it for local code)

ChristianTackeGSI:
> > The problem is that the error doesn't come from our own project but from some of our dependencies which we can't fix.
> 
> Sure, but still interested what Christian had in mind in such cases. (I personally never understood how "warning:=error" type of flags shall work.)

I naively had the hope that the option I had used would only affect our code (so that our code – which we can control – has no warnings at all).

As this flag also affects third party code, this hope has become a false one.

So, yes, please remove it from the general CI build.

Way forward:
1. Maybe only enable it in the ubuntu-latest (it should have FairSoft-dev. So a fixed geant4 could be available soon there, maybe?). I am not really sure, that this is a Good Thing(TM) in the light of the new knowledge about this setting. Not really my recommendation. I've just put it here for completeness
2. We're using "Warnings Next Generation" (I hope I am remembering correctly!) in our Jenkins CI. It filters the logs and adds metrics / etc for found warnings. It does not stop the build in any way. I think, it has support for CMake as well. So maybe we could enable that one? (I am not going to do that any time soon though.)

ChristianTackeGSI (approved):
Possibly open a new Issue (so that we remember about it) about using warnings-next-generation in the CI instead.

fuhlig1:
Close. The changes were merged with #1602

Issue #1609, opened 2025-07-14
Remove the dependency of TrackBase on Geant3


YanzhaoW:
**Is your feature request related to a problem? Please describe.**

Hi,

I am installing FairSoft without Geant3 library and found out that the TrackBase in FairRoot would not compile if Geant3 doesn't exist:

https://github.com/FairRootGroup/FairRoot/blob/b3a0a269b9fdbbfb982411eea3fb8a7007339ae7/fairroot/CMakeLists.txt#L32-L35

 But looking at the `trackbase/CMakeLists.txt`, I didn't find any explicit dependencies on Geant3 from TrackBase library.

Would it be possible to compile TrackBase even if Geant3 doesn't exist?

PR #1611, opened 2025-08-06
fix(build): Make trackbase build possible without Geant3

karabowi:
Moved the add_subdirectory(trackbase) outside of if(Geant3_FOUND) condition.

Fixes issue #1609.

---

Checklist:


ChristianTackeGSI (dismissed):
Generally this looks like the right thing to do.

If we want to support building without Geant3, do we have a testcase for this? (not sure, I haven't touched FairRoot for too long now).

P.S.: We probably should revive our CI one way or another?

karabowi:
@ChristianTackeGSI . Agree on everything.
In principle the test of `trackbase` is the `propagator` example. 
It falsely depended on `EventDisplay`. Fixed that.
Cleaned a bunch of stuff.

karabowi:
@ChristianTackeGSI can you have a look at the PR again?

examples/advanced/propagator/macros/runProp.C
```diff
@@ -64,8 +64,10 @@ int runProp(std::string propName = "rk")
 
     // -----   Set propagator and run   ---------------------------------------
     if (propName == "geane") {
+#ifdef FAIRGeane_H
```
ChristianTackeGSI: I really don't like it, if one matches on the include-guard of a header.

Either you include that header (then you know, you did), or you don't.

Please use some other preprocessor thing here.

ChristianTackeGSI (changes requested):
Please remember to change the copyright year on all files that you changed.

examples/advanced/propagator/macros/runProp.C
karabowi: @ChristianTackeGSI , done.

ChristianTackeGSI (approved):
Assuming you were able to compile things / run tests…

hopefully our CI will come back sometime.

karabowi:
I have tested on Mac/Sep25 and Ubuntu/May25.

PR #1632, opened 2026-06-13
Moving non-deterministic info line to debug output

kholoimov:
This PR moves the branch-address log from info to debug.

That output includes a raw pointer address, which is not deterministic across runs because it depends on runtime memory layout. Because of that, the logs are not deterministic even when the simulation result is the same. This makes automatic testing by direct log comparison difficult unless the output is cleaned up first. 
It also exposes low-level technical detail that is mainly useful for debugging, not for normal simulation logs.

We want to make this change to keep FairShip output cleaner and more deterministic for testing.

---

Checklist:

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -8,6 +8,7 @@ Eulisse, Giulio
 Friese, Volker
 Gertsenberger, Konstantin
 Hřivnáčová, Ivana
+Kholoimov, Valerii
 Klasen, Roman
 Kliemt, Ralf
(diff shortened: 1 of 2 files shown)


karabowi:
Can you please:
1. add a separate commit to this PR with your name in the `CONTRIBUTORS` file?
2. change the target branch from `FairRootGroup:master` to `FairRootGroup:dev`?

kholoimov:
@karabowi done!

karabowi (approved):

